PR #124, opened 2022-10-26
Fix/support-create-from-element

lawvs:
- Support create editor from custom element without store config
- Remove module mark
  - Due to our not regular enough esm config
img_v2_d00b01cf-158f-488e-a4d7-4c01076fbdcg

> ## Mandatory file extensions[#](https://nodejs.org/api/esm.html#mandatory-file-extensions)
> A file extension must be provided when using the import keyword to resolve relative or absolute specifiers. Directory indexes (e.g. './startup/index.js') must also be fully specified.
> This behavior matches how import behaves in browser environments, assuming a typically configured server.

### DIFF
--- a/packages/blocks/package.json
+++ b/packages/blocks/package.json
@@ -1,7 +1,6 @@
 {
   "name": "@blocksuite/blocks",
   "version": "0.2.19",
-  "type": "module",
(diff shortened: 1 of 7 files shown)


doodlewind:
Test playground breaks, this requires a hotfix.

PR #131, opened 2022-10-26
Bugfix/list set type error

DiamondThree:


doodlewind:
Some test cases are failed, should be marked as draft for now.

PR #141, opened 2022-10-27
fix: disable select list symbol

DiamondThree:


DiamondThree:
image
after fix
image

PR #142, opened 2022-10-27
fix: link experience

lawvs:
- add url valid
- optimize hotkey disable logic
- fix icon color
- fix link input not autofocus


packages/blocks/src/__internal__/utils/hotkey.ts
```diff
@@ -27,6 +27,19 @@ class HotkeyManager {
   enableHotkey() {
     this._setScope('page');
   }
+
+  /**
+   * Create a context to shielding against global hotkey
+   */
+  async withHotkeyShield(fn: () => void | Promise<unknown>) {
```
doodlewind, line 34: Seems that `switchContext` can be more expressive about the intention in comment. The concept of `Shield` looks rare in related API design.

doodlewind (approved):

packages/blocks/src/__internal__/utils/hotkey.ts
lawvs: Now updated to `withDisabledHotkey`

PR #145, opened 2022-10-28
build: migrate to self-hosted multi core runner

darkskygit:


lawvs:
GitHub Action is free for open-source project, blocksuite has been open-sourced, and we may not need self-hosted runners.

> GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners.

darkskygit:
> GitHub Action is free for open-source project, blocksuite has been open-sourced, and we may not need self-hosted runners.
> 
> > GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners.

Using a self-hosted server can put the complie time cache on the server, and github seems to count the compile time cache as storage size

lawvs:
In this repo, the cache only took 40MB, which is insignificant.

Screen Shot 2022-11-02 at 7 47 46 PM

darkskygit:
> In this repo, the cache only took 40MB, which is insignificant.
> 
> Screen Shot 2022-11-02 at 7 47 46 PM

oh, the cache size of this repository is smaller than I expected, it seems that a private runner is not needed at present

doodlewind:
I think the bottle neck for now is the time running tests, which can be significantly improved by having CI machine with more cores. @darkskygit can we apply for more quota?

image

darkskygit:
> I think the bottle neck for now is the time running tests, which can be significantly improved by having CI machine with more cores. @darkskygit can we apply for more quota?

larger action runner provide by github not free for public repo
our self-hosted machine can provide up to 16 physical core in single vm

doodlewind:
Sounds good, how about giving it a try with the `workers` config here updated? I'm open to migration, since we always have default GitHub actions as the baseline.

https://github.com/toeverything/blocksuite/blob/master/playwright.config.ts#L18

doodlewind:
Looks like we run into flaky test. The failed test case here may be resulted from different platform-specific font rendering strategies. What's the self-host platform version for now?

darkskygit:
image

image

The current self-hosted runner is 4 cores, which is 23% faster than github's runner

I monitored the runner cpu usage, current test can use all the cores, if we add more the number of cores, the speed will still be greatly improved

But the current runner will cause a test error. After some editing, press the up and down buttons and check whether the index is correct. I guess the dpi or ppi of the gui virtual monitor used by the runner may be different from the normal environment, so the The index obtained by the test is deviated

We may need to improve this test, my nas at home using lxc to perform this test will also have different results

darkskygit:
> Looks like we run into flaky test. The failed test case here may be resulted from different platform-specific font rendering strategies. What's the self-host platform version for now?

self hosted runner: 4c8t
- ubuntu 22 not work (14)
my nas's lxc container: 4c4t
- ubuntu 20 work
- arch linux 2022.09.03 not work (12)
- alpine 3.7 not work (14)




2022.09.03
--

doodlewind:
I think @SaikaSakura can take a look at the CI log. If we can just make that test case stable across platform, the migration speedup can be landed smoothly!

tzhangchi:
@SaikaSakura how it is going

tzhangchi:
@lawvs do you have some information to share

PR #147, opened 2022-10-31
fix: keyup cursor to title in first paragraph

DiamondThree:


packages/blocks/src/__internal__/utils/selection.ts
```diff
@@ -139,10 +139,19 @@ function focusRichTextByModel(
   position: SelectionPosition,
   model: BaseBlockModel
 ) {
-  const element = getBlockElementByModel(model);
-  const editableContainer = element?.querySelector('[contenteditable]');
-  if (editableContainer) {
-    focusRichText(position, editableContainer);
+   // FIXME: Then it will turn the input into the div
+  if (model.flavour === 'group') {
```
doodlewind, line 143: I think this looks too ad-hoc and violate the semantic of this function. Can we explicitly detect when is this case in `handleKeyUp`?

doodlewind (changes requested):

doodlewind:
Test case is also needed. We need to stop the unhealthy maintenance approach of ignoring test cases because of the pressure from fast iteration. It should be fine to put title-related test case into `paragraph.spec.ts`.

packages/blocks/src/__internal__/utils/selection.ts
DiamondThree: ok, that's a good idea, I will fix it.

doodlewind (approved):

PR #148, opened 2022-10-31
feature: add type jump code and link

SaikaSakura:


packages/blocks/src/page-block/utils/cursor.ts
```diff
@@ -0,0 +1,24 @@
+import {
+  caretRangeFromPoint,
+  resetNativeSelection,
+  SelectionEvent,
+} from '../../__internal__';
+
+export function repairerContextMenuRange(e: SelectionEvent) {
```
doodlewind, line 7: `repairer` -> `repair`

doodlewind (approved):

packages/blocks/src/page-block/utils/cursor.ts
SaikaSakura: Got , next commit I will correct it .

PR #154, opened 2022-11-01
Test: jsx snapshot parse text format

lawvs:
- parse text with format
- update snapshot
- throw an error at console error

### DIFF
--- a/tests/list.spec.ts
+++ b/tests/list.spec.ts
@@ -157,35 +157,17 @@ test('nested list blocks', async ({ page }) => {
   >
     <list
       prop:checked={false}
-      prop:text={
-        <>
-          <text
-            insert="123"
-          />
-        </>
-      }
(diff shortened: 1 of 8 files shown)


tests/list.spec.ts
```diff
@@ -157,17 +157,35 @@ test('nested list blocks', async ({ page }) => {
...
+            insert="123"
+          />
+        </>
+      }
       prop:type="bulleted"
     >
       <list
         prop:checked={false}
-        prop:text="456"
+        prop:text={
```
doodlewind, line 171: Can we make it compatible with current `prop:text="456"` style? For now all the test cases have not involved in rich text.

lawvs: Ok, now it will print string if the text no any format

doodlewind:
Great, thanks for the enhancement!

PR #156, opened 2022-11-01
feat: optimize cursor focus behavior on line start

SaikaSakura:


packages/blocks/src/__internal__/utils/selection.ts
```diff
@@ -68,22 +68,22 @@ function fixCurrentRangeToText(
   return range;
 }
 
-export function focusRichText(
+async function awaitUntilNextTick() {
```
doodlewind, line 71: I think this is not "next tick" but rather a so called "macro task". Why not reuse the `sleep` util here? https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/__internal__/utils/std.ts#L51

packages/blocks/src/__internal__/utils/selection.ts
```diff
@@ -93,16 +93,31 @@ export function focusRichText(
     const { x, y } = position;
     let newTop = y;
     let newLeft = x;
-    if (oldBottom <= y) {
-      newTop = bottom - lineHeight / 2;
+    if (bottom <= y) {
+      let finalBottom = bottom;
+      if (bottom < 100 && scrollContainer) {
```
doodlewind, line 98: Is the `100` here a magic number? If so we can make it a `const` since looks like it repeats for multi times.

packages/blocks/src/__internal__/utils/query.ts
```diff
@@ -94,7 +94,7 @@ export function getPreviousBlock(container: Element, blockId: string) {
     if (previousBlock?.model) {
       if (previousBlock.model.children.length) {
         let firstChildren =
```
doodlewind, line 96: `firstChildren` -> `firstChild`

packages/blocks/src/__internal__/utils/selection.ts
doodlewind: Also, just saying that we don't have to reply on each review comment, some comments that be easily resolved can be directly marked.

doodlewind (commented):
Also we'd better merge this after test cases are added.

PR #161, opened 2022-11-02
fix: cursor state after deleting on list or empty page

DiamondThree:


packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
```diff
@@ -159,11 +159,14 @@ export function handleLineStartBackspace(store: Store, model: ExtendedModel) {
       if (!parent || parent?.flavour === 'group') {
         const container = getContainerByModel(model);
         const previousSibling = getPreviousBlock(container, model.id);
-        if (previousSibling) {
+        if (previousSibling && previousSibling.flavour === 'paragraph') {
           store.captureSync();
+          let preTextLength = model.text?.length || 0;
           previousSibling.text?.join(model.text as Text);
           store.deleteBlock(model);
           asyncFocusRichText(store, previousSibling.id);
```
doodlewind, line 167: Can we switch current `asyncFocusRichText` implementation to your `quill.setSelection` approach, or just remove this line? We shouldn't set cursor twice with 2 different APIs.

doodlewind (changes requested):
This PR looks a bit careless, the case is not too tricky to add test case alongside, and again hope you *reinstall* VSCode to avoid the basic lint issue :(

packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
doodlewind: See comment in #162

doodlewind (approved):

tests/paragraph.spec.ts
```diff
@@ -310,3 +310,23 @@ test('The key "Up" respond when the cursor located in first paragraph  ', async
   await page.keyboard.press('ArrowUp', { delay: 50 });
   await assertPageTitleFocus(page);
 });
+
+
+test('After deleting a row of the text, the cursor will jump to the end of the previous row of the list', async ({ page }) => {
```
lawvs, line 315: You should use 'should' instead of 'will' in unit tests to indicate it is expected behavior.

doodlewind: Good catch!

PR #162, opened 2022-11-02
fix: edit link broken

lawvs:
- edit link text will cancel link

It is a workaround, it broke format skip by paste

### DIFF
--- a/packages/blocks/src/__internal__/rich-text/rich-text.ts
+++ b/packages/blocks/src/__internal__/rich-text/rich-text.ts
@@ -90,42 +90,22 @@ export class RichText extends LitElement {
           const nextParentElement = nextLeaf[0]?.domNode?.parentElement;
           const nextEmbedElement = nextParentElement?.closest(selector);
           const insertedString = delta.ops[1]?.insert.toString();
-          if (nextEmbedElement && nextEmbedElement !== currentEmbedElement) {
-            model.text?.delete(retain, delta.ops[1]?.insert.toString().length);
-            // @ts-ignore
-            if (!this.host.isCompositionStart) {
-              model.text?.insert(delta.ops[1]?.insert.toString() || '', retain);
-            } else {
-              // FIXME we must add a noon width space to fix cursor
-              model.text?.insert(' ', retain);
-              this.quill.setSelection(retain + 1, 0, 'api');
-            }
-          }
           if (
-            !nextEmbedElement &&
-            insertedString &&
-            source === Quill.sources.USER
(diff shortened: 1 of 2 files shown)


packages/blocks/src/__internal__/rich-text/rich-text.ts
```diff
@@ -87,35 +87,30 @@ export class RichText extends LitElement {
...
-                retain,
-                delta.ops[1]?.insert.toString() || ''
-              );
+              model.text?.insert(delta.ops[1]?.insert.toString() || '', retain);
             } else {
               // FIXME we must add a noon width space to fix cursor
-              this.quill.insertEmbed(retain, 'text', ' ');
+              model.text?.insert(' ', retain);
+
               this.quill.setSelection(retain + 1, 0, 'api');
```
lawvs, line 98: Is `quill.setSelection` recommended?

doodlewind: I think it's good to have a better reusable `focusRichText`, its API could be like `focusRichText(id: string, index?: number, async = false)`, which could be implemented by using `quill.setSelection` for now. Updating selection state doesn't mutate model, so using quill API is fine here. And this is also required in #161 .

cc @SaikaSakura @DiamondThree

packages/store/src/text-adapter.ts
```diff
@@ -181,6 +185,15 @@ export class Text {
       this._yText.meta = { delete: true };
     });
   }
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  cover(index: number, length: number, content: string, attributes?: Object) {
```
lawvs, line 189: Maybe `replace` is better to understand?

lawvs (commented):
Can you add some comments to explain what the `text-change` do?

And is needed to off the listener at `disconnectedCallback`?

packages/store/src/text-adapter.ts
```diff
@@ -181,6 +185,15 @@ export class Text {
...
+    this._transact(() => {
+      this._yText.delete(index, length);
+      this._yText.insert(index, content, attributes);
+      // @ts-ignore
```
doodlewind, line 193: Don't use `@ts-ignore` here, check out what we are doing here https://github.com/toeverything/blocksuite/blob/master/packages/store/src/text-adapter.ts#L92-L107

packages/store/src/text-adapter.ts
doodlewind: +1 for `replace`

SaikaSakura (approved):

PR #163, opened 2022-11-02
fix: After deleting all the text, two characters appear by hitting on…

### LINKED ISSUE #157: After range delete, redundant character shows after cursor position
**Description:**
At first, I pressed cmd + a to select all the text and pressed BackSpace to delete all of them. Then the next input would be triggered twice. And the extra char seems can not be deleted.

**Recording:**
select-all

Then I tried commenting the code in next pic, and it worked...

image

thorseraq:
Fix #157


doodlewind:
Quite a decent fix, thanks!

PR #165, opened 2022-11-03
Fix: selection and delete in edgeless mode

SaikaSakura:


doodlewind:
Thanks for looking into edgeless mode! Just reach me out for any further question.

PR #168, opened 2022-11-03
Add IndexDBProvider

colelawrence:
Please don't squash. Please use "rebase and merge" if you like the commit messages.

We refactor a couple more places with TypeScript types.
We have also added a wrapper around `yjs`'s `IndexeddbPersistence` as `IndexedDBProvider` for store.

### DIFF
--- a/packages/playground/src/main.ts
+++ b/packages/playground/src/main.ts
@@ -1,16 +1,34 @@
 import '@blocksuite/blocks';
 import '@blocksuite/editor';
 import { createEditor } from '@blocksuite/editor';
-import { DebugProvider } from '@blocksuite/store';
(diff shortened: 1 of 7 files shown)


packages/playground/src/main.ts
```diff
@@ -10,7 +10,7 @@ const room = params.get('room') ?? '';
 window.onload = () => {
   const editor = createEditor({
     room,
-    providers: [DebugProvider],
+    providers: [IndexDBProvider, DebugProvider],
   });
```
colelawrence, line 14: Change: Separate out another playground entry page with just IndexDB to ensure that CI tests work the same and we can test individual configurations separately.

PR #173, opened 2022-11-04
Create postcss.config.js

### LINKED ISSUE #171: Can not run dev environment
Potentially an issue isolated to Apple M-series chips.

Error on M1 (Pro).

Screen Shot 2022-11-04 at 09 42 11

ShortCipher5:
This would fix #171.

### DIFF
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -1,8 +1,1 @@
-module.exports = {
-    plugins: {
-      tailwindcss: {
-        content: ["./src/**/*.{html,js}"],
-      },
-      autoprefixer: {},
-    },
-  }
+module.exports = {}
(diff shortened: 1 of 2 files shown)


ShortCipher5:
Added a configuration to disable ESLint rules for *.config.js

postcss.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
doodlewind, line 1: I think we're not using it for now, so there may be another root cause for #171

ShortCipher5:
As it seems that tailwind is not used I removed the entry from the config.

A blank postcss.config.js does not seem to work, but an empty module export does:
`module.exports = {}`

ShortCipher5:
> As it seems that tailwind is not used I removed the entry from the config.
> 
> A blank postcss.config.js does not seem to work, but an empty module export does: `module.exports = {}`

Actually, as long as the file is not empty it can work - so even an empty comment `//` is ok.

Issue #150, opened 2022-10-31
Missing debug icons in Safari


doodlewind:
<img width="637" alt="image" src="https://user-images.githubusercontent.com/7312949/199019143-30f2d9b2-7b81-431d-89fb-df88f2379624.png">

In `debug-menu`, some icons adjusted by div wrapper were missing.

PR #180, opened 2022-11-08
fix: missing debug icon in safari

LuciNyan:
Closes: #150


lawvs:
Thank you for the work! These icons are copied from the fontawesome, I think it's better to keep it as it is.

Maybe we can add a style to the `DebugMenu`.

```css
/* https://github.com/toeverything/blocksuite/blob/7a98db05d2d6cea6457437e37d1c678e6dbecfdd/packages/blocks/src/__internal__/debug-menu.ts#L258-L259 */

.debug-menu > button > * {
  flex: 1;
}
```

LuciNyan:
@lawvs Thanks for your suggestion! fixed.

lawvs (approved):

PR #183, opened 2022-11-08
feat: fix selection and link problems

SaikaSakura:


packages/blocks/src/__internal__/utils/selection.ts
```diff
@@ -439,91 +439,93 @@ function expandRangesByCharacter(
...
-          i === currentNodeIndex ? selection.anchorOffset : undefined
-        );
-        if (text) {
-          const reverseText = Array.from(text).reverse().join('');
-          const index = reverseText.search(checkReg);
-          if (index !== -1) {
-            startNode = node;
-            startOffset = reverseText.length - index;
-            break;
+  if (leafNodes.length) {
```
lawvs, line 442: IMO, using early return can reduce the mental burden when reviewing code.

```suggestion
  if (!leafNodes.length) {
    return
  }
```

doodlewind: +1

PR #186, opened 2022-11-09
fix: when data loaded from indexeddb, there is a error on create new block

alt1o:


packages/playground/src/main.ts
```diff
@@ -5,35 +5,60 @@ import {
...
         );
     }
   });
 
+  /**
+   * Specified using "uuidv4" when providers have indexeddb.
+   * Because when persistent data applied to ydoc, we need generator different id for block.
+   * Otherwise, the block id will conflict.
+   */
+  const idGenerator = modes.includes('indexeddb')
```
lawvs, line 48:
```suggestion
  const idGenerator = providers.includes(IndexedDBProvider)
```

lawvs (approved):

PR #187, opened 2022-11-09
Feat: format quick bar

lawvs:
- [x] UI
- [x] position
- [ ] tests

---

## API Design

```ts
type showFormatQuickBar = async ({
  anchorEl: HTMLElement;   // Positioning points
  position: 'left-top' | 'right-bottom' | '...'
  container?: HTMLElement; // mount point, default is document.body
  abortController?: AbortController; // Can abort the format bar show from outside
}): Promise<void> // The promise will resolve after the format bar hidden
```


doodlewind (approved):
Looks good for now, and as a possible follow up, I think it's worth considering how to expose the QuickBar API, which can be part of the plan making BlockSuite customizable.

PR #191, opened 2022-11-09
feat: unbind block from store

QiShaoXuan:
Unbind block from store, block only depend on Space

### DIFF
--- a/packages/blocks/src/__internal__/debug-menu.ts
+++ b/packages/blocks/src/__internal__/debug-menu.ts
@@ -6,7 +6,7 @@ import {
   convertToList,
   createEvent,
 } from '../__internal__';
-import type { BaseBlockModel, Store } from '@blocksuite/store';
+import type { BaseBlockModel, Space } from '@blocksuite/store';
 import type { GroupBlockModel } from '../group-block';
 
 // Font Awesome Pro 6.2.0 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license (Commercial License) Copyright 2022 Fonticons, Inc.
@@ -123,7 +123,7 @@ const icons = {
 @customElement('debug-menu')
 export class DebugMenu extends LitElement {
   @property()
-  store!: Store;
+  space!: Space;
 
   @property()
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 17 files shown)


tests/basic.spec.ts
```diff
@@ -83,6 +99,15 @@ test('A open and edit, then joins B', async ({ browser, page: pageA }) => {
 
   const pageB = await browser.newPage();
   await enterPlaygroundRoom(pageB, room);
+  await pageB.evaluate(() => {
+    const space = window.store
+      .createSpace('page-test')
+      .register(window.blockSchema);
+    const editor = document.createElement('editor-container');
+    // @ts-ignore
```
doodlewind, line 107: Any idea to avoid `@ts-ignore` here?

packages/blocks/src/__internal__/debug-menu.ts
```diff
@@ -141,18 +141,18 @@ export class DebugMenu extends LitElement {
...
 
   private _onToggleConnection() {
-    if (this.connected === true) {
-      this.store.providers.forEach(provide => provide.disconnect());
-      this.connected = false;
-    } else {
-      this.store.providers.forEach(provide => provide.connect());
-      this.connected = true;
-    }
+    // if (this.connected === true) {
```
doodlewind, line 149: I think this should be enabled

packages/editor/src/components/editor-container/editor-container.ts
```diff
@@ -97,23 +90,23 @@ export class EditorContainer extends LitElement {
...
+        .space=${this.space}
         .model=${this.model}
       ></default-page-block>
     `;
 
     const edgelessContainer = html`
       <edgeless-page-block
         .mouseRoot=${this as HTMLElement}
-        .store=${this.store}
+        .space=${this.space}
```
doodlewind, line 109: I think passing down `space` here is fine, just saying that in the future, it may require other spaces to be used together with this one (for some possible cross page operations). But since we'll have properties like `space.store` and wrap spaces into pages, it shouldn't be a big deal.

packages/blocks/src/__internal__/debug-menu.ts
QiShaoXuan: It will be reopen soon, debug-menu will be moved to the editor file

Issue #189, opened 2022-11-09
Sub command of `pnpm dev` fails due to invalid environment variable setup on Windows


JimmFly:
When I try to use "pnpm dev" command in windows system, it throws the following error
<img width="377" alt="image" src="https://user-images.githubusercontent.com/102217452/200793300-f50f60f8-5286-4b66-a796-77fe4dde611f.png">

JimmFly:
I'll handle this myself

lawvs:
In general, if you want to compatible with Windows, you can add [cross-env](https://github.com/kentcdodds/cross-env) to fix it.

doodlewind:
@JimmFly welcome to give it a try! 😀

JimmFly:
I'll try my best

> 



> @JimmFly welcome to give it a try! 😀

PR #193, opened 2022-11-10
fix/environment-variable-setup (issue#189)

JimmFly:
add cross-env to fix the bug about [issue#189](https://github.com/toeverything/blocksuite/issues/189)


lawvs (approved):
Thanks!

PR #203, opened 2022-11-10
fix:selection test fails on windows

JimmFly:
In the original code, the clicked area on windows is the gap in the middle of the list, and the list is not clicked.
Now,I update click area of test to pass the test.
image

### DIFF
--- a/tests/selection.spec.ts
+++ b/tests/selection.spec.ts
@@ -38,7 +38,7 @@ test('click on blank area', async ({ page }) => {
   const above123 = await page.evaluate(() => {
     const paragraph = document.querySelector('[data-block-id="2"] p');
     const bbox = paragraph?.getBoundingClientRect() as DOMRect;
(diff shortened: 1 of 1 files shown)


JimmFly:
image

tests/selection.spec.ts
```diff
@@ -38,23 +38,23 @@ test('click on blank area', async ({ page }) => {
   const above123 = await page.evaluate(() => {
     const paragraph = document.querySelector('[data-block-id="2"] p');
     const bbox = paragraph?.getBoundingClientRect() as DOMRect;
-    return { x: bbox.left, y: bbox.top - 5 };
+    return { x: bbox.left, y: bbox.top + 5 };
```
lawvs, line 41: The mouse operates is relative to the top-left corner of the viewport in the playwright. So the minus calculations caused the operator can not click into the block on Windows.

See https://playwright.dev/docs/api/class-mouse

lawvs (approved):

PR #204, opened 2022-11-10
feat: remove debug-menu from @blocksuite/blocks to @blocksuite/editor

QiShaoXuan:
Remove debug-menu from @blocksuite/blocks to @blocksuite/editor
1. @blocksuite/blocks should just purely contain by blocks 
2. Debug-menu depend on Store model, is not exist in block any more

### DIFF
--- a/packages/blocks/src/index.ts
+++ b/packages/blocks/src/index.ts
@@ -13,6 +13,8 @@ export * from './list-block';
 export * from './group-block';
 export * as SelectionUtils from './__internal__/utils/selection';
 export * from './__internal__/utils/types';
+export * from './__internal__/utils/common-operations';
+export * from './__internal__/utils/std';
 
 const env =
   typeof globalThis !== 'undefined'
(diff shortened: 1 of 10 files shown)


packages/editor/src/components/debug-menu/debug-menu.ts
```diff
@@ -3,12 +3,12 @@ import { customElement, property, state } from 'lit/decorators.js';
```
lawvs, line 3: Why not just use `components/debug-menu.ts`

packages/blocks/src/index.ts
```diff
@@ -13,6 +13,8 @@ export * from './list-block';
 export * from './group-block';
 export * as SelectionUtils from './__internal__/utils/selection';
 export * from './__internal__/utils/types';
+export * from './__internal__/utils/common-operations';
+export * from './__internal__/utils/std';
```
lawvs, line 17: Is it necessary?

lawvs (changes requested):

packages/editor/src/components/debug-menu/debug-menu.ts
QiShaoXuan: I think  as the complexity of the features goes up, debug-menu may will be split by some atom compontents, so put it in a file is fine.

doodlewind (approved):

PR #206, opened 2022-11-10
Support loading playground with `init=everything` config

colelawrence:
Please review commit by commit.

<a href="https://www.loom.com/share/c99dd7c878e344d0b244245f39b63249">Cole's screen capture of 2022-11 Pull Request #206 · toeverythingblocksuite - 10 November 2022 2022-11-10 at 08 25 05@2x
<b>2022-11 Pull Request #206 · toeverything/blocksuite - 10 November 2022</b>
</a>

This essentially cleans up some of the ways we were scaffolding out documents for tests and it adds an "everything" option configurable via `?init=everything`.

Cole's screen capture of BlockSuite Playground 2022-11-10 at 08 28 09@2x

### DIFF
--- a/packages/playground/src/addEverythingToPage.ts
+++ b/packages/playground/src/addEverythingToPage.ts
@@ -1,0 +1,94 @@
+import type { PageBlockModel } from '@blocksuite/blocks';
+import { init } from './init-helper';
+
+export const addEverythingToPage = (page: PageBlockModel) =>
+  init.fromModel(page).self(page => {
+    page.addChild('affine:group', {}).self(group => {
+      group.addChild('affine:paragraph', { type: 'h2' }).withText('Headings');
+
+      const sampleText = addEverythingToPage
+        .toString()
+        .replace(/[^\w]+/g, ' ')
+        .slice(0, 240)
+        .trim();
+
+      group
+        .addChild('affine:paragraph', { type: 'h1' })
+        .withText('Heading 1 with trailing text');
+      group.addChild('affine:paragraph', { type: 'text' }).withText(sampleText);
+      group
+        .addChild('affine:paragraph', { type: 'h2' })
+        .withText('Heading 2 with trailing text');
+      group.addChild('affine:paragraph', { type: 'text' }).withText(sampleText);
+      group
+        .addChild('affine:paragraph', { type: 'h3' })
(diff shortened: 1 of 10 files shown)


lawvs:
We have a plan to implement a function for init page from JSX   

It's like a reverse implementation about https://github.com/toeverything/blocksuite/pull/154

```tsx
await initPageFromJSX(`
<page>
  <group
    prop:xywh="[0,0,720,32]"
  >
    <paragraph
      prop:text={
        <>
          <text
            insert="bold"
            bold={true}
          />
        </>
      }
      prop:type="text"
    />
  </group>
</page>`)

// Or may have a simple version
await initContentFromJSX(`<paragraph prop:text="this is a text" prop:type="text" />`)
```

What do you think?

colelawrence:
Just a note, I found the following resource describing a way to specify separate locally defined JSX pragma in TypeScript. If I understand this correctly, this could be a great approach for us to use so we aren't requiring that the definitions go into the global type space under JSX. https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#locally-scoped-jsx-namespaces

doodlewind:
> I'm worried about not having IDE help in hand writing this XML format as string.

Serialized JSX code *can* be generated from YDoc, and the YDoc itself can be generated by editing in our example page or just paste in content from clipboard. So I believe it's better to stick to JSX instead of introducing all these new concepts and fixtures in this PR. Note that this is also just a secondary nice-to-have feature being kept in mind.

> we have to explain that if we start expanding the usage of this JSX syntax in other places

I think this is not a major concern, since in the future OpenBlock stage, we need a comprehensive documentation about our file format - while I'm not saying that standard should be based on JSX.

So I'm also not feeling strongly about this PR, it sounds better to only support `?init=everything` until we need corresponding test cases, and it can fine to just keep the playground entry as simple as possible.

colelawrence:
@doodlewind, sounds great! I'm on the same page there. I think it will be very easy to add a JSX in place of the builder I threw together here. So, let's block this on #218, which should be a relatively fun feature to contribute.

doodlewind:
I'm closing this in favor of JSX and the [pattern](https://github.com/toeverything/blocksuite/tree/master/packages/playground/examples) that uses multi examples as the entry of module test cases.

PR #211, opened 2022-11-11
WIP: Blob uploader interfaces

colelawrence:
Work in progress, but it shows a bit of the direction I'm pointed in for a "BlobStorage" interface and how that interacts through the "Store" & "Space".

### DIFF
--- a/packages/store/src/providers.ts
+++ b/packages/store/src/providers.ts
@@ -3,15 +3,23 @@ import { WebrtcProvider } from 'y-webrtc';
 // In the future, consider making y-indexdb a separate package sync provider.
 import { IndexeddbPersistence } from 'y-indexeddb';
 import type { Awareness } from 'y-protocols/awareness';
+import type { BlobStorage } from './blob-storage/blob-storage-types';
+import { IndexedDBBlobStorage } from './blob-storage/indexeddb-blob-storage';
 
 /**
  * Different examples of providers could include webrtc sync,
- * database sync like SQLite / LevelDB, or even web IndexDB.
+ * database sync like SQLite / LevelDB, or even web IndexedDB.
  *
  * Usually a class will also implement {@link SyncProviderConstructor}.
  */
 export interface SyncProvider {
   awareness?: Awareness;
+  /**
+   * Each {@link SyncProvider} is essentially used as a mirror in practice, so for now,
+   * the easiest approach is to simply enable every sync provider to provide their own
+   * separate blob storage option.
+   */
+  blobStorage?: BlobStorage;
   connect: () => void;
   disconnect: () => void;
   clearData: () => Promise<void>;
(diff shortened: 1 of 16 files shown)


packages/store/src/Ability.type.ts
```diff
@@ -0,0 +1,18 @@
+export type Ability<T> =
+  | ({
+      enabled: true;
+    } & T)
+  | {
+      enabled: false;
+      userMessage: string;
```
colelawrence, line 7: A message to show in UI when this ability is not enabled.

For example, `"Webcam permissions have not been granted"` or `"Store was not configured with any providers that support uploads."`

packages/store/src/BlobStorage.types.ts
```diff
@@ -0,0 +1,35 @@
...
+  // // Hmm: How will we handle exporting large files / lots of files?
+  // // We shouldn't just glob together everything into a big JSON, right?
+  // // Should we require exporting by individual resource ids?
+  // // Should exporting be a part of a higher interface on a per storage basis?
+  // // e.g. export for a Torrent or IPFS stored resources might actually export Content Hashes
+  // // and import takes those to re-associate them...
+  // // But, that feels like a specific for the specific storage.
+  // /** Export to something... need goals defined */
+  // export(): Promise<Record<string, unknown>>;
+  // import(): Promise<void>;
```
colelawrence, line 34: I'm not entirely sure what the context is of the `export` + `import`. Is this supposed to look similar to anything that already exists?

packages/store/src/IndexedDBBlobStorage.ts
```diff
@@ -0,0 +1,73 @@
...
+    res: null! as (value: T) => unknown,
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    rej: null! as (err: unknown) => unknown,
+    promise: new Promise<T>((res, rej) => {
+      control.res = res;
+      control.rej = rej;
+    }),
+  };
+  return control;
+}
```
colelawrence, line 73: Helper for latent resolved promise

packages/playground/src/createDefaultUglyUploadUI.ts
```diff
@@ -0,0 +1,95 @@
...
+import { ChangeEvent } from 'react';
+
+export function createDefaultUglyUploadUI(options: {
+  mountContainer: HTMLElement;
+}): StoreOptionsShowUploadUI {
+  const $promptText$ = new Signal<string>();
+  const $filesSelected$ = new Signal<FileList | null>();
+  const $show$ = new Signal<boolean>();
+  const $cancel$ = new Signal<true>();
+  const { dom } = renderSpec(document, [
```
doodlewind, line 17: I don' really understand why do we need `renderSpec` pattern here, if we are just adding it for mocking example UI, this looks overkill.

packages/store/src/providers.ts
```diff
@@ -12,6 +14,12 @@ import type { Awareness } from 'y-protocols/awareness';
  */
 export interface SyncProvider {
   awareness?: Awareness;
+  /**
+   * Each {@link SyncProvider} is essentially used as a mirror in practice, so for now,
+   * the easiest approach is to simply enable every sync provider to provide their own
+   * separate blob storage option.
+   */
+  blobStorage?: BlobStorage;
```
doodlewind, line 22: I think `BlobStorage` can be separated from the SyncProvider, since I consider the SyncProvider should only be be used to sync YDoc (via different databased or network protocols).

doodlewind:
Also we are following the `kebab-case.ts` convention for file naming.

packages/store/src/BlobStorage.types.ts
doodlewind: `export` means to dump the storage content to user file system (for backup and data ownership), just like what the `export workspace` feature in https://livedemo.affine.pro/ is doing. And with `import` we load them back.

doodlewind:
I still consider the scope of this PR too basic for our expected blob persistence feature, so please don't merge this PR at the moment. Will sync about that with you later.

doodlewind:
I'm going to close this in favor of the [experimental blob modules](https://github.com/toeverything/blocksuite/tree/master/packages/store/src/blob), which currently work and are covered by test cases. @colelawrence thanks for the co-designing effort!

Issue #184, opened 2022-11-08
Add super call in all Lit lifecycle callbacks


doodlewind:
We didn't enforce the [best practice](https://lit.dev/docs/v1/components/lifecycle/#:~:text=All%20lifecycle%20methods%20need%20to%20call%20the%20super%20method.) to always add `super` call in lifecycles of Lit components yet. This may lead to subtle bugs and should be fixed.

Another points to decide:

1. Whether the `super` call should always happen at the start of the overridden method? Doing this seems to break some test cases
2. Whether should we add the `override` and `protected` prefix (and `: void` suffix) in all Lit lifecycle callbacks?
3. Whether should we use `private _foo` instead of `private foo`, in terms of variable names and method names?

LuciNyan:
There’s a lot of discussion about this [here](https://github.com/angular/angular.io/issues/1108), and overall it’s probably a holy war. XD

In my $0.02. I think `private foo` is better. Except in one case: 
```typescript
private _foo;
get foo() { return this._foo; }
set foo(value: any) { 
  console.log(`${this._foo} changed to ${value}`);
  this._foo = value;
}
```

doodlewind:
Thanks for your feedback! I think we can remove underscore to unify the naming style (PR welcome), though it's not very urgent yet.

PR #213, opened 2022-11-18
chore: add super call in all lit lifecycle callbacks

LuciNyan:
Closes: #184


doodlewind:
Thanks for the improvement!

Issue #194, opened 2022-11-10
The rectangle of the selection is in the wrong position


JimmFly:
The rectangle position of the selection is wrong when the page is not at the top
![select](https://user-images.githubusercontent.com/102217452/201026485-34a38361-adc2-4eba-81c7-1d15acb2d803.gif)

PR #214, opened 2022-11-22
fix(blocks): rectangle of the selection in wrong position

LuciNyan:
### Description
Fix the rectangle of the selection is in the wrong position.

There are two smol issues here 🤔: 
1. Is there a better-named alternative to `affine-default-viewport`? What do you think.
2. I am trying to understand your design's original intention, but I am confused. In your original design, if both `selectionRect` and `selectedRects` need to be positioned relative to the viewport, why does one use position absolute and the other use position fixed? Or do you want the selectionRect to scroll with the page?

Closes: #194


doodlewind:
> Is there a better-named alternative to `affine-default-viewport`?

This makes sense!

> if both selectionRect and selectedRects need to be positioned relative to the viewport, why does one use position absolute and the other use position fixed?

I think this is just being carelessly implemented...Thanks for tracking it down!

PR #215, opened 2022-11-23
fix(blocks): remove listeners

LuciNyan:
### Description
Did some opinionated optimizations.

### DIFF
--- a/packages/blocks/src/page-block/default/default-page-block.ts
+++ b/packages/blocks/src/page-block/default/default-page-block.ts
@@ -234,12 +234,8 @@ export class DefaultPageBlockComponent extends LitElement implements BlockHost {
   private _removeHotkeys() {
     removeCommonHotKey();
     hotkey.removeListener([
-      HOTKEYS.UNDO,
-      HOTKEYS.REDO,
       HOTKEYS.BACKSPACE,
       HOTKEYS.SELECT_ALL,
-      HOTKEYS.INLINE_CODE,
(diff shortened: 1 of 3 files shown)


packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -234,12 +234,8 @@ export class DefaultPageBlockComponent extends LitElement implements BlockHost {
   private _removeHotkeys() {
     removeCommonHotKey();
     hotkey.removeListener([
-      HOTKEYS.UNDO,
```
LuciNyan, line 237: It seems that some listeners have been removed in `removeCommonHotKey`, is it still needed here? 🤔

packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
```diff
@@ -118,7 +118,19 @@ export class EdgelessPageBlockComponent
   }
 
   private _removeHotkeys() {
-    hotkey.removeListener([HOTKEYS.BACKSPACE], this.flavour);
+    hotkey.removeListener([
```
LuciNyan, line 121: These listeners seem should be removed. I'm not quite sure if it's careless or intentional here.

packages/blocks/src/page-block/default/default-page-block.ts
doodlewind: Looks good to avoid duplication here!

PR #223, opened 2022-11-28
fix anchor node of native selection

thorseraq:


packages/blocks/src/page-block/utils/cursor.ts
```diff
@@ -65,7 +65,9 @@ export function getNativeSelectionMouseDragInfo(e: SelectionEvent) {
   const direction: DragDirection = getDragDirection(e);
   console.log(`direction: ${direction}`);
```
doodlewind, line 66: We'd better remove this kind of log code in our codebase :)

doodlewind (approved):

packages/blocks/src/page-block/utils/cursor.ts
lawvs: This PR is merged to the `feat/quick-format-bar`, I will clear the log before merging to master.

PR #227, opened 2022-11-28
feat: add divider support

JimmFly:
Added divider support for markdown syntax

https://user-images.githubusercontent.com/102217452/204328548-9f70810b-e29e-4420-ad0b-bcf53a8666f8.mov


doodlewind:
It would be best if you could adhere to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for commit messages :)

packages/playground/vite.config.ts.timestamp-1669732289014.mjs
```diff
@@ -0,0 +1,10 @@
+// vite.config.ts
```
doodlewind, line 1: What is this module used for? Should it get committed?

packages/editor/src/block-loader.ts
```diff
@@ -12,4 +13,5 @@ export const BlockSchema = {
   'affine:page': PageBlockModel,
   'affine:list': ListBlockModel,
   'affine:group': GroupBlockModel,
+  'affine:divider': DividerBlockModel,
```
doodlewind, line 16: Please also add schema definition here https://github.com/toeverything/blocksuite/blob/master/packages/store/src/__tests__/space.spec.ts#L18-L23

This is used for unit testing.

packages/blocks/src/divider-block/divider-model.ts
```diff
@@ -0,0 +1,32 @@
+import { Space, BaseBlockModel, IBaseBlockProps } from '@blocksuite/store';
+
+type DividerType = 'normal';
+//TODO: add another type like https://jjuliano.github.io/markdown-ui/docs/divider.html
+
+export interface DividerBlockProps extends IBaseBlockProps {
+  flavour: 'affine:divider';
+  type: DividerType;
```
doodlewind, line 8: We shouldn't add this `type` field if no sub-typing is required here.

packages/playground/vite.config.ts.timestamp-1669732289014.mjs
JimmFly: This was added automatically without my knowledge, I will delete it.

packages/editor/src/block-loader.ts
JimmFly: Ok, I'll gradually add some missing code.

JimmFly:
divider-block currently supports adding markdown syntax such as "---+space" or "***+space". There are also functions such as clicking to select, selecting and deleting, copying, and pasting. When the cursor passes over a divider, that divider is selected. By the way, it can also be deleted after drag selection.

https://user-images.githubusercontent.com/102217452/205338495-f3e3a880-fcc6-4e30-aac2-77c751d9483c.mov

packages/blocks/src/__internal__/utils/common-operations.ts
```diff
@@ -93,3 +93,35 @@ export function convertToParagraph(
   }
   return true;
 }
+
+export function convertToDivider(
+  page: Page,
+  model: ExtendedModel,
+  type: 'affine:divider' | 'normal',
```
doodlewind, line 100: What is the `normal` argument used for?

packages/blocks/src/divider-block/divider-model.ts
```diff
@@ -0,0 +1,23 @@
...
+  flavour: 'affine:divider';
+  type: 'normal';
+}
+
+export class DividerBlockModel
+  extends BaseBlockModel
+  implements DividerBlockProps
+{
+  flavour = 'affine:divider' as const;
+  type: 'normal';
```
doodlewind, line 13: The `normal` field should be removed if it doesn't distinguish something.

packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -193,6 +200,60 @@ function EmbedOptionContainer(embedOption: EmbedOption | null) {
...
+    if (!nextBlock) {
+      return;
+    } else if (
+      nextBlock.flavour === 'affine:list' ||
+      nextBlock.flavour === 'affine:paragraph'
+    ) {
+      focusNextBlock(model, 'start');
+      state.clear();
+      return;
+    } else if (nextBlock.flavour === 'affine:divider') {
```
doodlewind, line 245: Please use `matchFlavours(['affine:divider'])` in *ALL* `block.flavour === 'affine:foo'` cases. This can make us able to better track *where* do we depend on such typing.

packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -306,6 +381,68 @@ export class DefaultPageBlockComponent extends LitElement implements BlockHost {
...
+    hotkey.addListener(UP, e => {
+      switch (state.type) {
+        case 'none':
+          break;
+        case 'block':
+          state.type = 'divider';
+          break;
+        case 'divider':
+          this.signals.updateSelectedRects.emit([]);
+          handleUP(this.selection);
```
doodlewind, line 393: `UP` -> `Up`

packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -306,6 +381,68 @@ export class DefaultPageBlockComponent extends LitElement implements BlockHost {
...
+    hotkey.addListener(RIGHT, e => {
+      switch (state.type) {
+        case 'none':
+          break;
+        case 'block':
+          state.type = 'divider';
+          break;
+        case 'divider':
+          this.signals.updateSelectedRects.emit([]);
+          handleDOWN(this.selection);
```
doodlewind, line 439: `DOWN` -> `Down`

packages/editor/src/managers/clipboard/paste-manager.ts
```diff
@@ -263,7 +270,10 @@ export class PasteManager {
       };
       const id = this._editor.page.addBlock(blockProps, parent, index + i);
       const model = this._editor.page.getBlockById(id);
-      if (model?.type !== 'affine:embed') {
+      if (
```
doodlewind, line 273: use `matchFlavour(['affine:embed', 'affine:divider'])` instead

packages/store/src/utils/utils.ts
```diff
@@ -99,6 +100,9 @@ export function syncBlockProps(
   if (props.flavour === 'affine:embed' && !yBlock.has('prop:caption')) {
     yBlock.set('prop:caption', props.caption ?? '');
   }
+  if (props.flavour === 'affine:divider' && !yBlock.has('prop:type')) {
```
doodlewind, line 103: Remember to remove here if we are going the remove `divider.type` :)

doodlewind (changes requested):

doodlewind (approved):

doodlewind:
Thanks for your hard work!

PR #228, opened 2022-11-28
search feature

darkskygit:
this pr add basic indexing quality local search feature into @blocksuite/store

the indexing service will refresh index all prop:text and prop:title in a block when they add/update/remove, at this time indexing feature support ascii words and cjk tokenize, if user use non-chromium core, tokenize function will fallback to char based tokenize

image

in future we will implement a native version in rust with same apis, and must allow add more custom props from outside

### DIFF
--- a/packages/store/src/search.ts
+++ b/packages/store/src/search.ts
@@ -61,14 +61,13 @@ export class Indexer {
   }
 
   search(query: QueryContent) {
-    this._indexer.search(query as string);
+    return this._indexer.search(query as string);
   }
 
   private _handleSpaceIndexing(spaceId: string, space?: YMap<YBlock>) {
     if (space) {
       space.forEach((block, key) => {
         this._refreshIndex(spaceId, key, 'add', space.get(key));
-        console.log(block, key);
       });
 
       space.observeDeep(events => {
@@ -105,11 +104,6 @@ export class Indexer {
             block.get('prop:title') || block.get('prop:text')
           );
           if (content) {
-            console.log({
-              content,
-              reference: '',
-              tags: [space],
-            });
             this._indexer.add(id, {
               content,
               reference: '',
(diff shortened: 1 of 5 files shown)


packages/store/src/search.ts
```diff
@@ -0,0 +1,146 @@
+import { Document as DocumentIndexer, DocumentSearchOptions } from 'flexsearch';
+import { Doc, Map as YMap, Text as YText } from 'yjs';
+import type { YBlock } from './space';
+
+export type QueryContent = string | Partial<DocumentSearchOptions<boolean>>;
+
+function tokenizeZh(text: string) {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  const tokenizer = Intl?.v8BreakIterator;
```
doodlewind, line 10: Can we replace this non-standard API with standardized [Intl.Segmenter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter)?

packages/store/src/store.ts
```diff
@@ -51,6 +59,8 @@ export class Store {
       spaceId,
       new Space(spaceId, this.doc, this.awareness, this.idGenerator)
     );
+    this._indexer.onCreateSpace(spaceId);
```
doodlewind, line 62: What should we do after the space is updated afterward?

packages/store/src/search.ts
darkskygit: v8BreakIterator can support chromium after 2018, Intl.Segmenter add official support after 2020/11, maybe we can make v8BreakIterator as a fallback

packages/store/src/store.ts
darkskygit: seem we cannot observe root map change (doc update event / abstract type event)

packages/store/src/search.ts
doodlewind: Browser support after 2020 looks promising. I think it'll be fine to support browser versions within a year for the oversea market, which is what [photopea](https://www.photopea.com/) is doing.

PR #232, opened 2022-11-29
Fix: Boundary case of native dragging between block components

thorseraq:
`comparePoint()` is unstable when dragging across HTMLELement. During native dragging, wrong `isForward` derived by it will cause [failure of focuing TextNode ](https://github.com/toeverything/blocksuite/blob/3a07c6f540f5dd9c8bc1ae2bf5f05cca7dcbf196/packages/blocks/src/__internal__/utils/selection.ts#L74).  Use `getDragDirection()` to cover this case. 

# Before
https://user-images.githubusercontent.com/20554850/204439699-db0a36f2-3821-4651-a75e-b56785b7b3c2.mov
# After
https://user-images.githubusercontent.com/20554850/204439704-305efdf8-87bf-4467-90e0-8327efee514c.mov

### DIFF
--- a/packages/blocks/src/__internal__/utils/selection.ts
+++ b/packages/blocks/src/__internal__/utils/selection.ts
@@ -14,6 +14,8 @@ import {
 } from './query';
 import { Rect } from './rect';
 import type { SelectionEvent } from './gesture';
+import type { DragDirection } from '../../page-block/utils';
+import { getDragDirection } from '../../page-block/utils';
 
 const SCROLL_THRESHOLD = 100;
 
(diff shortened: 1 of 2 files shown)


doodlewind:
LGTM, I hope for more precise shots like this one!

doodlewind:
It's weird that CI doesn't run for this PR, anything wrong here?

thorseraq:
I checked the test.yaml file, CI won't be triggered if the pull request is not targeted at master branch. Is that meet our expectations?

doodlewind:
> I checked the test.yaml file, CI won't be triggered if the pull request is not targeted at master branch. Does that meet our expectations?

Got it, this looks fine :)

thorseraq:
OK!

PR #233, opened 2022-11-29
Feat/workspace

QiShaoXuan:
1. Add a attribute 'attribute' in Space model
2. Change a metohd's name in Store model

### DIFF
--- a/packages/store/src/space.ts
+++ b/packages/store/src/space.ts
@@ -57,6 +57,8 @@ export class Space {
     rootDeleted: new Signal<string>(),
     textUpdated: new Signal<Y.YTextEvent>(),
     updated: new Signal(),
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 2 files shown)


packages/store/src/space.ts
```diff
@@ -57,6 +57,8 @@ export class Space {
     rootDeleted: new Signal<string>(),
     textUpdated: new Signal<Y.YTextEvent>(),
     updated: new Signal(),
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    attribute: new Signal<Map<string, any>>(),
```
doodlewind, line 61: * `attribute` -> `attributes`
* Can we replace `any` with `unknown`?

QiShaoXuan: cool, I will repair it

doodlewind: I have done the rename stuff, just checkout the master branch :)

PR #236, opened 2022-11-29
fix (clean code and fix unit test): boundary case of native dragging between block components

thorseraq:


thorseraq:
Before creating a pr request to feature branch, run unit test locally can prevent this problem, I should have done this :)

PR #237, opened 2022-11-29
Feat/new image

DiamondThree:


packages/blocks/src/page-block/default/utils.ts
```diff
@@ -0,0 +1,70 @@
+import type { BaseBlockModel } from '@blocksuite/store';
+import { assertExists, getBlockById } from '../../__internal__/utils';
+
+export const pick = (blocks: BaseBlockModel[], x: number, y: number) => {
```
lawvs, line 4: You need to add more description to explain this function.

As a reviewer, I want to know what it picks, what the `x` and `y` work, and what it returns, so you need to explain it.

packages/blocks/src/page-block/default/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+    x > block.left + block.width + 50 ||
+    y < block.top ||
+    y > block.top + block.height
+  ) {
+    return false;
+  }
+  return true;
+}
+
+export function downloadImage(url: string | undefined) {
```
lawvs, line 38: the `undefined` should not exist in the type of input, because it always fails if the input is undefined.

So the `assertExists` can be safely removed.

packages/blocks/src/page-block/default/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+  link.setAttribute('target', '_blank');
+  document.body.appendChild(link);
+  link.download = 'test';
+  link.click();
+  document.body.removeChild(link);
+  link.remove();
+}
+
+export async function writeClipImg(imgURL: string | undefined) {
+  assertExists(imgURL);
```
lawvs, line 51:
```suggestion
export async function writeClipImg(imgURL: string) {
```

packages/blocks/src/page-block/default/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    await navigator.clipboard.write([
+      new ClipboardItem({
+        [blob.type]: blob,
+      }),
+    ]);
+  } catch (err: unknown) {
+    if (err instanceof Error) {
+      console.error(err.name, err.message);
+    }
```
lawvs, line 64: It seems the catch clause doesn't to be doing anything.

packages/blocks/src/page-block/default/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+}
+
+export function deleteBlock(model: BaseBlockModel) {
+  model.space.deleteBlock(model);
+}
```
lawvs, line 70: the function seems unnecessary

lawvs (changes requested):

packages/blocks/src/image-block/image-block.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  // disable shadow DOM to workaround quill
+  createRenderRoot() {
+    return this;
+  }
+
+  // This is the initial width before event resize is applied
+
+  private _selectImage() {
+    this._canEditor = true;
```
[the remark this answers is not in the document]
DiamondThree: this does's work  ,I will delete it

doodlewind:
We need `embed.spec.ts` for testing embed block features.

JimmFly:
I think the separator block I'm currently writing should be based on your embed block, we'll need further communication to confirm this. I hope you can reply me, thank you.

doodlewind:
> I think the [divider-block](https://github.com/toeverything/blocksuite/pull/227) I'm currently writing should be based on your embed block

I am afraid that's not the case. Embed blocks should include `figma`, `video`, `youtube`, `file`, .etc. But for DividerBlock, I don't see the need for reusing such [embed-level](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/embed) extensibility. In this case, extending such a complex component will only make your implementation heavier are harder to maintain.

JimmFly:
> > I think the [divider-block](https://github.com/toeverything/blocksuite/pull/227) I'm currently writing should be based on your embed block
> 
> I am afraid that's not the case. Embed blocks should include `figma`, `video`, `youtube`, `file`, .etc. But for DividerBlock, I don't see the need for reusing such [embed-level](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/embed) extensibility. In this case, extending such a complex component will only make your implementation heavier are harder to maintain.

Ok, I'll take your advice and act on it as soon as possible, thanks.

lawvs (approved):

PR #242, opened 2022-11-30
Test/format quick bar

lawvs:
- Add test for format quick bar
- Upgrade playwright to v1.28.1 due to https://github.com/microsoft/playwright/issues/16517 and it is fixed in https://github.com/microsoft/playwright/pull/17477

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   "license": "MPL-2.0",
   "devDependencies": {
     "@changesets/cli": "^2.25.0",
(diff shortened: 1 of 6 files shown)


doodlewind:
How to upgrade the global playwright version?

lawvs:
After I updated the lock file, you will encounter the error when running the test locally, just follow the guide to update the playwright.
Screenshot 2022-12-01 at 8 15 06 AM

PR #243, opened 2022-12-01
basic blob cloud sync

darkskygit:
this pr provide the blob cloud sync with jwst server

- [x] blob sync pipeline & retry mechanism
- [x] jwst http apis
- [x] test case

compare to origin blob storage with idb, this implement will deduplicate blobs based on the unique hash of the blob content and provides a lazy loading strategy

cloud sync test case can run with jwst server: `cargo run --bin keck`


packages/store/src/blob/__tests__/test-entry.ts
```diff
@@ -57,8 +60,10 @@ async function testBasic() {
   });
 
   testSerial('can delete image', async () => {
-    await storage.delete(id);
-    const url = await storage.get(id);
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
doodlewind, line 63: We can do this:

``` ts
const url = await storage.get(id);
assertExists(url);
```

packages/store/src/blob/__tests__/test-entry.ts
```diff
@@ -114,14 +119,64 @@ async function testRefreshAfter() {
...
 function clearIndexedDB() {
   return new Promise<void>(resolve => {
-    const request = indexedDB.deleteDatabase('keyval-store');
+    const request = indexedDB.deleteDatabase('test_blob');
     request.onsuccess = () => {
-      console.log('IndexedDB cleared');
-      resolve();
+      console.log('IndexedDB test_blob cleared');
+
+      const request = indexedDB.deleteDatabase('test_padding');
```
doodlewind, line 126: `pending`

packages/store/src/blob/providers.ts
```diff
@@ -0,0 +1,226 @@
...
+      await this._padding.set(task.id, {
+        type: task.type,
+        retry: task.retry + 1,
+      });
+      this._pipeline.push({ ...task, retry: task.retry + 1 });
+      await sleep(Math.min(10, task.retry) * 100);
+    }
+  }
+
+  private async taskRunner() {
```
doodlewind, line 168: Please keep the `private async _taskRunner` naming convention :)

packages/store/src/blob/types.ts
```diff
@@ -0,0 +1,24 @@
...
+    blobAdded: Signal<BlobId>;
+  };
+  get(id: BlobId): Promise<BlobURL | null>;
+  set(blob: Blob): Promise<BlobId>;
+  delete(id: BlobId): Promise<void>;
+  clear(): Promise<void>;
+}
+
+export type IdbInstance<T = ArrayBufferLike> = {
+  get: (key: BlobId) => Promise<T | undefined>;
```
doodlewind, line 18: * `IdbInstance` -> `IDBInstance` 
* Can we use `<T | null>` instead of `<T | undefined>`?

packages/store/src/blob/providers.ts
```diff
@@ -0,0 +1,226 @@
...
+              .post(api, { body: task.blob, retry: 3 })
+              .json<BlobStatus>();
+            await this.addTaskRetry(task, status);
+          }
+        } catch (e) {
+          console.warn('Error while syncing blob', e);
+          await this.addTaskRetry(task);
+        }
+      }
+      await sleep(500);
```
doodlewind, line 192: How about making here a `RETRY_TIMEOUT` const variable?

packages/store/src/blob/providers.ts
```diff
@@ -0,0 +1,226 @@
...
+
+export class BlobCloudSync {
+  private readonly _abortController = new AbortController();
+  private readonly _fetcher: typeof ky;
+  private readonly _database: IdbInstance;
+  private readonly _padding: IdbInstance<{
+    retry: number;
+    type: 'add' | 'delete';
+  }>;
+  private readonly _paddingPipeline: SyncTask[] = [];
```
doodlewind, line 117: `pending`

packages/store/src/blob/types.ts
darkskygit: idb-keyval define type with Promise<T | undefined>, additional runtime check is required if we wish to convert correctly

PR #244, opened 2022-12-01
add image test

DiamondThree:
Add image drop test


packages/playground/src/main.ts
```diff
@@ -91,5 +91,19 @@ window.onload = () => {
 
     document.body.appendChild(editor);
     document.body.appendChild(debugMenu);
+    const pageId = page.addBlock({ flavour: 'affine:page', title: 'hello' });
```
doodlewind: This snippet should only be used for local testing and not be committed.

doodlewind (approved):

PR #245, opened 2022-12-01
fix: paragraph convert

lawvs:
- Fix can not convert list to heading
- Add test


lawvs:
@doodlewind What is the different between `updateTextType` and `batchUpdateTextType` in `container-operations`?

I tried using `batchUpdateTextType` but it did not work.

doodlewind:
> What is the difference between updateTextType and batchUpdateTextType in `container-operations`?

I think this may be just some tech debt. Removing one of them would be fine.

PR #247, opened 2022-12-02
feat: edgeless drag component

himself65:
https://user-images.githubusercontent.com/14026360/205458450-a6db0e49-7aaf-450e-985a-58b129fcb76e.mov

- [x] Test case
- [x] Persistence after switch mode

### DIFF
--- a/packages/blocks/src/page-block/default/default-page-block.ts
+++ b/packages/blocks/src/page-block/default/default-page-block.ts
@@ -32,7 +32,6 @@ import {
   handleBlockSelectionBatchDelete,
   handleSelectAll,
   removeCommonHotKey,
-  tryUpdateGroupSize,
   updateTextType,
 } from '../utils';
 import style from './style.css';
@@ -480,11 +479,6 @@ export class DefaultPageBlockComponent extends LitElement implements BlockHost {
     this.signals.updateEmbedOption.on(embedOption => {
       this.embedOption = embedOption;
       this.requestUpdate();
-    });
-    tryUpdateGroupSize(this.page, 1);
-    this.addEventListener('keydown', e => {
-      if (e.ctrlKey || e.metaKey || e.shiftKey) return;
-      tryUpdateGroupSize(this.page, 1);
     });
 
     // TMP: clear selected rects on scroll
--- a/packages/blocks/src/page-block/default/default-page-block.ts
+++ b/packages/blocks/src/page-block/default/default-page-block.ts
@@ -32,6 +32,7 @@ import {
   handleBlockSelectionBatchDelete,
   handleSelectAll,
   removeCommonHotKey,
+  tryUpdateGroupSize,
   updateTextType,
 } from '../utils';
 import style from './style.css';
@@ -479,6 +480,11 @@ export class DefaultPageBlockComponent extends LitElement implements BlockHost {
     this.signals.updateEmbedOption.on(embedOption => {
       this.embedOption = embedOption;
       this.requestUpdate();
+    });
+    tryUpdateGroupSize(this.page, 1);
+    this.addEventListener('keydown', e => {
+      if (e.ctrlKey || e.metaKey || e.shiftKey) return;
+      tryUpdateGroupSize(this.page, 1);
     });
 
     // TMP: clear selected rects on scroll
--- a/packages/blocks/src/page-block/edgeless/components.ts
+++ b/packages/blocks/src/page-block/edgeless/components.ts
@@ -116,7 +116,7 @@ function EdgelessBlockChild(
   };
 
   return html`
-    <div class="affine-edgeless-block-child" style=${styleMap(style)}>
+    <div test-id=${model.id} class="affine-edgeless-block-child" style=${styleMap(style)}>
       ${BlockElement(model, host)}
     </div>
   `;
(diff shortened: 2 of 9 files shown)


packages/blocks/src/page-block/edgeless/components.ts
```diff
@@ -188,3 +166,147 @@ export function EdgelessBlockChildrenContainer(
...
+  rect!: DOMRect;
+
+  private _dragStartInfo = {
+    startMouseLeft: 0,
+    absoluteLeft: 0,
+    width: 0,
+    direction: 'left',
+  };
+
+  private getHandles(rect: DOMRect) {
```
doodlewind, line 188: Just saying that we are using the `private _getHandles` naming convention 👀

packages/blocks/src/page-block/edgeless/components.ts
```diff
@@ -188,3 +166,147 @@ export function EdgelessBlockChildrenContainer(
...
+        width: rect.width,
+        direction: type,
+      };
+      // parent ele is the edgeless block container
+      this.parentElement?.addEventListener('mousemove', this._onDrag);
+      this.parentElement?.addEventListener('mouseup', this._onDragEnd);
+    }
+  };
+
+  private _onDrag = (e: MouseEvent) => {
```
doodlewind, line 239: `_onDragMove` matches our convention better :)

packages/blocks/src/page-block/edgeless/components.ts
```diff
@@ -188,3 +166,147 @@ export function EdgelessBlockChildrenContainer(
...
+      const groupContainer = groupBlock?.parentElement;
+      // first change container`s x/w directly for get groups real height
+      if (groupContainer) {
+        groupContainer.style.width = newW + 'px';
+        groupContainer.style.translate = `translate(${newX}px, ${y}px) scale(${this.zoom})`;
+      }
+      // reset the width of the container may trigger animation
+      requestAnimationFrame(() => {
+        // refresh xywh by model
+        space.updateBlock(selected, {
```
doodlewind: We need to call `store.captureSync()` once before the first time the drag move handler gets triggered. This can prevent the drag operation from being merged with previous operations, ensuring the current operation can be undone standalone.

himself65 (commented):
image

For the current commit, there's a UI bug on `edgeless-selected-rect`, that cannot sync the position and size. still investigating.

himself65:
> image
> 
> For the current commit, there's a UI bug on `edgeless-selected-rect`, that cannot sync the position and size. still investigating.

Fixed in https://github.com/toeverything/blocksuite/pull/247/commits/8c525fca27946b1298822493104766d9d76172bc

himself65:
Now I'm working on how to persist the size when switching mode

packages/blocks/src/page-block/edgeless/selection-manager.ts
```diff
@@ -216,6 +218,7 @@ export class EdgelessSelectionManager {
         this._blockSelectionState = {
           type: 'single',
           active: false,
+          viewpoint: this._container.viewport,
```
doodlewind, line 221: `viewport`?

packages/blocks/src/page-block/edgeless/components.ts
```diff
@@ -188,3 +166,159 @@ export function EdgelessBlockChildrenContainer(
...
+      }
+      // reset the width of the container may trigger animation
+      requestAnimationFrame(() => {
+        // refresh xywh by model
+        if (!this.lock) {
+          selected.page.captureSync();
+          this.lock = true;
+        }
+        if (this.state.type !== 'none') {
+          this.state.rect = getSelectionBoxBound(viewpoint, selected.xywh);
```
himself65: The reason is I have to sync the position of the selection rect. I didn't find any way to access the `this._document.viewpoint`

packages/blocks/src/page-block/edgeless/selection-manager.ts
himself65: https://github.com/toeverything/blocksuite/pull/247#discussion_r1038356323

packages/blocks/src/page-block/edgeless/components.ts
doodlewind: Just curious to know if `viewpoint` is a typo of `viewport` here?

himself65: 😨

himself65: I have only the primary school degree(

packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -481,11 +480,6 @@ export class DefaultPageBlockComponent extends LitElement implements BlockHost {
       this.embedOption = embedOption;
       this.requestUpdate();
     });
-    tryUpdateGroupSize(this.page, 1);
-    this.addEventListener('keydown', e => {
-      if (e.ctrlKey || e.metaKey || e.shiftKey) return;
```
himself65: I think we shouldn't modify the width and height in default block mode because the zoom is always `1`, and there's no edgeless UI in this mode

himself65:
The test case looks weird🤷

Himself65:
I think we need calc the xywh in the first time, otherwise, there will have major changes in the test cases

doodlewind:
Impressive work, thanks!

PR #248, opened 2022-12-02
fix: improve format quick bar

lawvs:
- fix paragraph panel text wrap on pathfinder
- fix some icons that can not be active
- fix when only one line is selected, the position of inline format menu may incorrect.


doodlewind:
Though it may be bikeshedding, can we consider renaming the `FormatQuickBar` component to `InlineHoverMenu` or `InlineFormatMenu`? This sounds semantically better to me. I'll appreciate it if you are open to renaming. Thanks!

lawvs:
> Though it may be bikeshedding, can we consider renaming the `FormatQuickBar` component to `InlineHoverMenu` or `InlineFormatMenu`? This sounds semantically better to me. I'll appreciate it if you are open to renaming. Thanks!

I will discuss this with PM to get a more ergonomic description.

doodlewind:
Let's merge this first. Renaming can happen anytime :)

PR #251, opened 2022-12-03
feat: BlockSchema type in space class

Himself65:
image


doodlewind:
Thanks for the enhancement!

PR #255, opened 2022-12-05
feat: Rewrite quill, adding support for line number and languange selection

thorseraq:


packages/blocks/src/code-block/components/SyntaxCodeBlock.ts
```diff
@@ -0,0 +1,112 @@
+// @ts-nocheck
```
doodlewind, line 1: I thinks this file should be named as `syntax-code-block.ts` and would be great if we can remove the `@ts-nocheck` in the future.

thorseraq: Get that, file should be named this style@ts-nocheck is a trade-off so far, Quill's declartion file doesn't correspond to its packed dist/index.js file, confusing

lawvs: If there are type errors, you should fix the type manually instead ignore all type

doodlewind: @lawvs I believe the original file is JavaScript. I can imagine the possible friction. We can do some type-refactoring after the critical features get stable or make it progressive.

packages/blocks/src/code-block/components/lang-list.ts
```diff
@@ -0,0 +1,277 @@
...
+  @property()
+  showLangList = 'hidden';
+
+  @state()
+  disposeTimer = 0;
+
+  @property()
+  delay = 150;
+
+  languages = [
```
himself65: Make this static constant?

thorseraq: Good idea!

doodlewind (approved):
Thanks for the hard work, let's merge this!

PR #257, opened 2022-12-05
fix: clipboard state when first block is divider

JimmFly:
When dragging and selecting, if the first block is a divider, copying and pasting will miss the first line

### DIFF
--- a/packages/editor/src/managers/clipboard/paste-manager.ts
+++ b/packages/editor/src/managers/clipboard/paste-manager.ts
@@ -197,8 +197,13 @@ export class PasteManager {
         selectedBlock?.text?.insertList(insertTexts, endIndex);
         selectedBlock &&
           this._addBlocks(blocks[0].children, selectedBlock, 0, addBlockIds);
-
-        parent && this._addBlocks(blocks.slice(1), parent, index, addBlockIds);
+        if (blocks[0].flavour === 'affine:divider') {
+          parent &&
+            this._addBlocks(blocks.slice(0), parent, index, addBlockIds);
+        } else {
+          parent &&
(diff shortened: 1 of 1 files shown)


packages/editor/src/managers/clipboard/paste-manager.ts
```diff
@@ -197,8 +197,13 @@ export class PasteManager {
         selectedBlock?.text?.insertList(insertTexts, endIndex);
         selectedBlock &&
           this._addBlocks(blocks[0].children, selectedBlock, 0, addBlockIds);
-
-        parent && this._addBlocks(blocks.slice(1), parent, index, addBlockIds);
+        if (blocks[0].flavour === 'affine:divider') {
```
lawvs, line 200: Maybe you need to add some comments to explain why the divider is special

JimmFly: > And please stick to the `matchFlavour` pattern.

But the block here is an OpenBlockInfo[], not a model

packages/editor/src/managers/clipboard/paste-manager.ts
```diff
@@ -197,8 +197,14 @@ export class PasteManager {
         selectedBlock?.text?.insertList(insertTexts, endIndex);
         selectedBlock &&
           this._addBlocks(blocks[0].children, selectedBlock, 0, addBlockIds);
-
-        parent && this._addBlocks(blocks.slice(1), parent, index, addBlockIds);
+        //This is a temporary processing of the divider block, subsequent refactoring of the divider will remove it
```
doodlewind, line 200: A tiny style suggestion: use `// hello` instead of `//hello`, adding a space after the double slashes. @JimmFly

PR #258, opened 2022-12-05
fix: select editing

### LINKED ISSUE #200: After selecting multiple lines, editing only modifies the first line
After selecting multiple lines, editing only modifies the first line
delete

himself65:
Fixes: https://github.com/toeverything/blocksuite/issues/200

### DIFF
--- a/packages/blocks/src/page-block/default/default-page-block.ts
+++ b/packages/blocks/src/page-block/default/default-page-block.ts
@@ -30,19 +30,23 @@ import {
   getContainerByModel,
   getPreviousBlock,
   getNextBlock,
-  focusNextBlock, getCurrentRange, isMultiBlockRange, getModelsByRange,
+  focusNextBlock,
+  getCurrentRange,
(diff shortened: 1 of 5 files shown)


packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -34,7 +34,7 @@ import {
   getCurrentRange,
   isMultiBlockRange,
   getModelsByRange,
-} from '../../__internal__'
+} from '../../__internal__';
```
doodlewind, line 37: Just saying that we have `pnpm format` for prettifying.

doodlewind (approved):
Great fix, thanks!

cc @lawvs

packages/blocks/src/page-block/default/default-page-block.ts
himself65: I would like to use [husky](https://www.npmjs.com/package/husky) to run it automatically before submitting

PR #266, opened 2022-12-06
fix: selectively join deleted rich text

### LINKED ISSUE #265: Redundant character shows after cursor position again
**Reproduction:**
1. press BACKSPACE at an empty block
2. type any character

**Recording**
recording

donaldxdonald:
Fix #265


doodlewind:
Good catch! Thanks for the fix!

PR #268, opened 2022-12-07
Fix/position edge case

lawvs:
- Fix some position edge case for quick format bar
- Extract the custom element `FormatQuickBar` to a new file
- Add a clamp util function
- Simplify format bar position overflow calculate

I recommend reviewing this pull request commit by commit.

## BUG Details

- drag on an empty line will trigger an error
- drag to an empty line will cause the format bar positing error

https://user-images.githubusercontent.com/18554747/206054629-6a19b690-a45f-499b-b042-b000e2294af0.mov

- drag to the start of the line will cause the format bar positing error

https://user-images.githubusercontent.com/18554747/206054942-9f859ec1-cac2-48c2-b256-24c722ec63a2.mov


packages/blocks/src/__internal__/utils/std.ts
```diff
@@ -90,3 +90,28 @@ export const throttle = <T extends (...args: any[]) => void>(
...
+ * otherwise, the function will return value.
+ *
+ * @example
+ * ```ts
+ * const x = clamp(10, 0, 5); // x will be 5
+ * const y = clamp(3, 0, 5); // y will be 3
+ * const z = clamp(-1, 0, 5); // z will be 0
+ * ```
+ */
+export const clamp = (value: number, min: number, max: number): number => {
```
doodlewind, line 109: I think we prefer `function` over `const` in util modules. This doesn't matter, though.

PR #269, opened 2022-12-08
fix: respond to newly added space & trigger pageUpdates

doodlewind:
- [x] fix collab test cases
- [x] update PageMeta on page block title updates (manually synced in `default-page-block`)
- [x] <del>update block title on PageMeta updates by `setPageMeta`</del> - HARD to sync state on undo/redo
- [x] support `createDate` and `trashDate`

cc @QiShaoXuan

### DIFF
--- a/packages/playground/src/main.ts
+++ b/packages/playground/src/main.ts
@@ -1,3 +1,4 @@
+// import * as Y from 'yjs';
 import '@blocksuite/blocks';
 import '@blocksuite/editor';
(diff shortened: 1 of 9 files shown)


packages/playground/src/main.ts
```diff
@@ -95,3 +98,5 @@ window.onload = () => {
     document.body.appendChild(debugMenu);
   }
 };
+
+// window.Y = Y;
```
himself65: I would like to use `Object.defineProperty` with a only getter or non-writable

QiShaoXuan (approved):

PR #271, opened 2022-12-10
fix: non-text block behavior of direction keys

DiamondThree:
about no-text keyboard features and bugs, for example, 'up', and 'down'.


packages/blocks/src/page-block/default/selection-manager.ts
```diff
@@ -379,42 +354,34 @@ export class DefaultSelectionManager {
...
       e.raw.pageX,
       e.raw.pageY
     );
-    if (clickBlockInfo?.model.type === 'image') {
+    if (clickBlockInfo && clickBlockInfo.model) {
+      const { model } = clickBlockInfo;
+      const page = getDefaultPageBlock(model);
+      page.lastSelectionPosition = 'start';
+    }
+    if (clickBlockInfo && NON_TEXT_ARR.includes(clickBlockInfo.model.type)) {
```
doodlewind, line 369: Please add a `matchTypes` in the same module of `matchFlavours`, so as to replace `NON_TEXT_ARR.includes` (and other similar places) with them, this benefits static analysis about the usage of block flavours and sub-types.

packages/blocks/src/page-block/default/selection-manager.ts
```diff
@@ -433,12 +400,17 @@ export class DefaultSelectionManager {
   };
 
   private _onContainerMouseMove = (e: SelectionEvent) => {
-    const hoverOption = getHoverBlockOptionByPosition(
+    const hoverOption = getBlockOptionByPosition(
```
doodlewind, line 403: The concept of `option` looks a bit weird here since `options` should be a static config instead of a dynamically updated state. Other block doesn't have options, and I don't expect a block to be defined together with `options`.

Quick thought: how about renaming `EmbedOption` to `EmbedState` or `EmbedEditingState`? Latter is preferred personally since future third-party blocks could have their own editing states, and concepts could be unified.

packages/blocks/src/page-block/default/utils.ts
```diff
@@ -3,29 +3,34 @@ import {
   assertExists,
   getBlockById,
   getBlockElementByModel,
-  matchFlavours,
 } from '../../__internal__/utils';
 
-export const getHoverBlockOptionByPosition = (
+export const NON_TEXT_ARR = ['image', 'divider'];
+
+export const getBlockOptionByPosition = (
```
doodlewind, line 10: Please use `export function` for in-module utils.

packages/blocks/src/page-block/utils/container-operations.ts
```diff
@@ -133,7 +132,6 @@ export function handleBackspace(page: Page, e: KeyboardEvent) {
       const { quill } = richText;
       if (isCollapsedAtBlockStart(quill)) {
         // use quill handler
-        noop();
```
doodlewind, line 136: This line should be kept. This is an explicit `noop`, telling people we intentionally do nothing here.

packages/playground/src/main.ts
```diff
@@ -95,5 +95,15 @@ window.onload = () => {
 
     document.body.appendChild(editor);
     document.body.appendChild(debugMenu);
+    // const idPage = page.addBlock({ flavour: 'affine:page' });
```
doodlewind, line 98: Please double-check before commit.

packages/blocks/src/__internal__/rich-text/rich-text.ts
```diff
@@ -47,6 +47,12 @@ export class RichText extends LitElement {
   }
 
   firstUpdated() {
+    // this._textContainer.addEventListener('keyup', e => {
```
doodlewind, line 50: Should also be removed.

packages/blocks/src/divider-block/divider-model.ts
```diff
@@ -2,14 +2,15 @@ import { Page, BaseBlockModel, IBaseBlockProps } from '@blocksuite/store';
 
 export interface DividerBlockProps extends IBaseBlockProps {
   flavour: 'affine:divider';
+  type: 'divider';
```
doodlewind, line 5: Why do we need this `type` field? If fixing the image block requires all other blocks to define a different `type` field, then the way of the fix seems to be problematic.

packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -48,13 +50,9 @@ import {
...
 import style from './style.css';
-import {
-  CaptionIcon,
-  CopyIcon,
-  DeleteIcon,
-  DownloadIcon,
-} from '../../image-block/icons';
+import { CaptionIcon, CopyIcon, DeleteIcon, DownloadIcon } from '../icons';
 import { downloadImage, focusCaption, copyImgToClip } from './utils';
+import { isAtLineEdge } from '../../__internal__/rich-text/rich-text-operations';
```
doodlewind, line 55: Just saying that we usually add a blank line after all `import` statements. Similarly, a blank line should also be added after we defined the `EmbedOption` interface (well `EmbedEditingState` seems to be a better fix here). It is recommended to refer to the existing code base to determine when crowded code should be avoided.

packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -212,63 +210,87 @@ function EmbedOptionContainer(
...
+    const model = getStartModelBySelection();
+    const activeContainer = getContainerByModel(model);
+    const activePreNodeModel = getPreviousBlock(activeContainer, model.id);
+    const editableContainer = getBlockElementByModel(model)?.querySelector(
+      '.ql-editor'
+    ) as HTMLElement;
+    const range = nativeSelection.getRangeAt(0);
+    const { height, left, top } = range.getBoundingClientRect();
+
+    // TODO resolve compatible problem
```
doodlewind, line 229: This looks like a false comment since we don't have the compatibility problem by having `caretRangeFromPoint` to support Firefox.

doodlewind (changes requested):

packages/blocks/src/page-block/default/selection-manager.ts
DiamondThree: That’s good idea

packages/blocks/src/divider-block/divider-model.ts
```diff
@@ -9,7 +9,7 @@ export class DividerBlockModel
   implements DividerBlockProps
 {
   flavour = 'affine:divider' as const;
-
+  type = 'divider' as const;
```
doodlewind, line 12: Why do we still have a `type` field here?

doodlewind (approved):

PR #276, opened 2022-12-12
chore: improve type

Himself65:
This PR is the first attempt at improving type.

Make BaseBlockModel an abstract class so that every code cannot directly create an instance from the class.

### DIFF
--- a/packages/store/src/base.ts
+++ b/packages/store/src/base.ts
@@ -12,13 +12,13 @@ export interface IBaseBlockProps {
   text?: TextType;
 }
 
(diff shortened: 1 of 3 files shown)


doodlewind:
I have to move the `page.register(BlockSchema)` API to `workspace.register`, since the previous `page = workspace.createPage()` usage was not usable for multi-player or page refresh (using IndexedDB). In these cases, the page instance should be automatically created by the provider, instead of being manually created. For the difference, you can check out the new `main.ts` in the playground.

So I guess if we don't register BlockSchema statically onto page, the previous typing trick won't work for now :(

PR #280, opened 2022-12-12
feat: change editor mode from state to property

QiShaoXuan:
Change editor mode from state to property, editor's mode should be controled by it container, no mater its initial value or changes


doodlewind (approved):
LGTM

PR #281, opened 2022-12-12
feat: return clearer search results

darkskygit:
~~add a util function help search bar find page space by block id~~

preprocess the search results and return them in the form of Map<BlockId, SpaceId>


darkskygit:
find an other method to store the page mapping: https://github.com/nextapps-de/flexsearch#document-store
I try to switch to this way to avoid adding api

darkskygit:
finished

doodlewind:
Thanks for the enhancement!

PR #284, opened 2022-12-12
Fix incorrect selection in deleteModels

### LINKED ISSUE #283: After selecting with line end, redundant character shows again
**Two issues:**
1. Select multiple blocks and drop the cursor at the end of line end. An error was threw out.
2. No need to delete firstRichText in `container-operations.ts` - `deleteModels` when the selected text in firstRicthText is empty.

**Recording:**
recording

donaldxdonald:
Fix #283

### DIFF
--- a/packages/blocks/src/page-block/utils/container-operations.ts
+++ b/packages/blocks/src/page-block/utils/container-operations.ts
@@ -49,10 +49,15 @@ export function deleteModels(page: Page, models: BaseBlockModel[]) {
     false
   );
 
-  firstRichText.model.text?.delete(
-    firstTextIndex,
-    firstRichText.model.text.length - firstTextIndex
-  );
+  if (
+    firstRichText.model.text &&
+    firstRichText.model.text.length !== firstTextIndex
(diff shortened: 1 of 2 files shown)


packages/blocks/src/page-block/utils/container-operations.ts
```diff
@@ -49,10 +49,15 @@ export function deleteModels(page: Page, models: BaseBlockModel[]) {
     false
   );
 
-  firstRichText.model.text?.delete(
-    firstTextIndex,
-    firstRichText.model.text.length - firstTextIndex
-  );
+  if (
```
doodlewind, line 52: Minor refactoring advice:

``` ts
const isFirstRichTextEmpty = firstRichText.model.text?.length !== firstTextIndex;
// See https://github.com/toeverything/blocksuite/issues/283
if (!isFirstRichTextEmpty) {
  firstRichText.model.text.delete(
    firstTextIndex,
    firstRichText.model.text.length - firstTextIndex
  );
}
```

PR #292, opened 2022-12-13
feat: image pasting and blob persistence

DiamondThree:
Fix image bugs about paste， and add a feature that publish image file.


doodlewind (changes requested):
Please write a semantic PR title (instead of default branch name) or short description at the very least. Otherwise it's hard for maintainers to know what it intends to do

packages/editor/src/managers/clipboard/paste-manager.ts
```diff
@@ -198,7 +191,10 @@ export class PasteManager {
         selectedBlock &&
           this._addBlocks(blocks[0].children, selectedBlock, 0, addBlockIds);
         //This is a temporary processing of the divider block, subsequent refactoring of the divider will remove it
-        if (blocks[0].flavour === 'affine:divider') {
+        if (
+          blocks[0].flavour === 'affine:divider' ||
```
doodlewind, line 195:
``` ts
matchFlavours(blocks[0], ['affine:divider', 'affine:embed'])
```

packages/blocks/src/image-block/image-block.ts
```diff
@@ -22,18 +23,26 @@ export class ImageBlockComponent extends LitElement {
...
 
-  override firstUpdated() {
+  override async firstUpdated() {
     this.model.propsUpdated.on(() => this.requestUpdate());
     this.model.childrenUpdated.on(() => this.requestUpdate());
     // exclude padding and border width
     const { width, height } = this.model;
+    // FIXME: after export global storage, we can use it directly
+    const storage = new BlobStorage();
+    const provider = await IndexedDBBlobProvider.init('test');
```
doodlewind, line 42: We should use workspace id here.

packages/blocks/src/page-block/default/selection-manager.ts
```diff
@@ -71,7 +71,12 @@ function createSelectionRect(
   return new DOMRect(left, top, width, height);
 }
 
-type PageSelectionType = 'native' | 'block' | 'none' | 'embed';
+type PageSelectionType =
+  | 'native'
+  | 'native_collapsed'
```
doodlewind, line 76: 1. This may overlap with other types since I guess a `native_collapsed` selection type is also `native`. This may lead to confusion and tricky bugs.
2. We don't use `snake_case`

packages/blocks/src/page-block/default/selection-manager.ts
```diff
@@ -425,6 +430,7 @@ export class DefaultSelectionManager {
       e.raw.pageX,
       e.raw.pageY
     );
+    if ((e.raw.target as HTMLElement).closest('.image-option')) return;
```
doodlewind, line 433: I thought we have replaced the `option` concept with `editing-state`.

packages/blocks/src/image-block/image-block.ts
DiamondThree: This is just a temporary solution， I will fix it when exporting global storage.

packages/editor/src/managers/clipboard/paste-manager.ts
DiamondThree: I tried to do it, but I found block[0]'s type is not BaseMoudel, that is OpenBlockInfo, so I can't use this function

packages/blocks/src/page-block/default/selection-manager.ts
DiamondThree: @SaikaSakura

packages/blocks/src/image-block/image-block.ts
doodlewind: We already have `this.host.page.id` to use here. A storage store should be under page level, instead of workspace level.

doodlewind (approved):

PR #299, opened 2022-12-14
fix: getPage maybe nullable

himself65:
upstream: 984ea012dec47f6dbe415aa7e875e9adb5cacae2

### DIFF
--- a/tests/edgeless.spec.ts
+++ b/tests/edgeless.spec.ts
@@ -22,16 +22,20 @@ async function getGroupSize(
   page: Page,
   ids: { pageId: string; groupId: string; paragraphId: string }
 ) {
-  const result = await page.evaluate(
+  const result: string | null = await page.evaluate(
     ([id]) => {
-      const block = window.workspace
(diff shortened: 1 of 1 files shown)


doodlewind (approved):
LGTM

doodlewind:
`assertExists` can avoid the `foo!` warning

himself65:
> `assertExists` can avoid the `foo!` warning

I use another workaround

himself65:
> `assertExists` can avoid the `foo!` warning

Do you mean `assertExists` from '../packages/blocks'?

doodlewind:
> Do you mean `assertExists` from '../packages/blocks'?

I think that may not work in a Node environment. But there are indeed some low-hanging fruits about our testing utils since we already intentionally have two `assertExists`. One is for browser runtime, and another is for unit testing the store (`test-utils-dom.ts`). Due to our complexity in testing (running tests across DOM and Node in a mixed way), it's challenging to decide the best practice to reuse test utils. I could chat later with you.

PR #300, opened 2022-12-14
fix: `affine-edgeless-frame-selection-rect` position

Himself65:
Fixes: https://github.com/toeverything/blocksuite/issues/297


doodlewind:
Good catch, thanks!

PR #303, opened 2022-12-14
fix: selection manager with shape mode

Himself65:


Himself65:
WIP, dont merge!!!

packages/blocks/src/page-block/edgeless/components.ts
```diff
@@ -33,12 +34,10 @@ function getCommonRectStyle(rect: DOMRect, zoom: number, isShape = false) {
...
 }
 
-export function EdgelessHoverRect(
-  rect: DOMRect | null,
-  zoom: number,
-  isShape = false
-) {
-  if (!rect) return html` <div></div>`;
+export function EdgelessHoverRect(hoverState: HoverState | null, zoom: number) {
+  if (!hoverState) return html` <div></div>`;
```
doodlewind, line 38: We can just `return null`

packages/blocks/src/page-block/edgeless/selection-manager/default.ts
```diff
@@ -0,0 +1,210 @@
...
+      resetNativeSelection(null);
+    }
+  }
+
+  onContainerContextMenu(e: SelectionEvent): void {
+    repairContextMenuRange(e);
+  }
+
+  onContainerDblClick(e: SelectionEvent): void {
+    // do nothing
```
doodlewind, line 107: We express this kind of intention with `noop()` explicitly. This applies to other `// do nothing` lines.

packages/blocks/src/page-block/edgeless/selection-manager/shape.ts
```diff
@@ -0,0 +1,83 @@
...
+import type { SelectionArea } from '../selection-manager';
+
+export class ShapeSelectionController extends SelectionController<ShapeMouseMode> {
+  readonly mouseMode = <ShapeMouseMode>{
+    type: 'shape',
+    color: 'black',
+    shape: 'rectangle',
+  };
+
+  private _draggingShapeBlockId: string | null = null;
```
doodlewind, line 13: What if we have multiple shapes being dragged together?

doodlewind (commented):
Impressive, this looks well-established!

packages/blocks/src/page-block/edgeless/selection-manager/shape.ts
himself65: oh, this is only for creating a new shape(drag from a point to another pont)

himself65: I think there's no way in human logic that create multiple shapes in one drag

PR #306, opened 2022-12-14
ci: add `codeql-analysis.yml`

himself65:


doodlewind:
@lawvs found a REGEX blocker here. It would be great if you could take a look. Thanks!

lawvs:
OK, I bypassed the scan by splitting the regex :)

himself65:
lgtm

doodlewind:
It costs 4min to complete CodeQL check, this slows us down. Can we make it run daily? For future multi-browser E2E testing, this kind of setup should be also useful.

himself65:
> It costs 4min to complete CodeQL check, this slows us down. Can we make it run daily? For future multi-browser E2E testing, this kind of setup should be also useful.

I think there should have some cache on the following incremental commit?

doodlewind:
> > It costs 4min to complete CodeQL check, this slows us down. Can we make it run daily? For future multi-browser E2E testing, this kind of setup should be also useful.
> 
> I think there should have some cache on the following incremental commit?

Well, let's just try committing some trivial refactoring here and see again 👀

doodlewind:
I'd say the incremental run speed is still not very satisfying, especially for a check whose report is not our concern in maybe 99% of the time :(

I really want to make the CI process *faster* and *easier*, lots of effort has been done for reaching the current state. Daily running the CodeQL check seems to be more acceptable, though.

himself65:
updated

doodlewind:
Great improvement for our CI, thanks!

PR #309, opened 2022-12-15
feat: support readonly

QiShaoXuan:


packages/blocks/src/__internal__/utils/components.ts
```diff
@@ -11,38 +11,48 @@ import type { EmbedBlockModel } from '../../embed-block';
...
           .host=${host}
+          .readonly=${readonly}
         ></paragraph-block>
       `;
     case 'affine:list':
       return html`
         <list-block
           .model=${model as ListBlockModel}
           .host=${host}
+          .readonly=${readonly}
```
doodlewind, line 39: I'm not sure if we should add arbitrary new props in this way. Can we `host.readonly` instead?

PR #310, opened 2022-12-15
Fix unexpected keyboard behaviours after native drag

thorseraq:


tests/utils/actions/misc.ts
```diff
@@ -171,6 +171,18 @@ export async function getQuillSelectionText(page: Page) {
   });
 }
 
+export async function getSelectedTextByQuill(page: Page) {
+  return await page.evaluate(() => {
+    const selection = window.getSelection() as Selection;
+    const range = selection.getRangeAt(0);
+    const component = range.startContainer.parentElement?.closest('rich-text');
+    // @ts-ignore
```
lawvs, line 179: Prefer to use `@ts-expect-error`

tests/selection.spec.ts
```diff
@@ -454,3 +455,39 @@ async function clickDivider(page: Page, i = 0) {
   const locator = page.locator('divider-block').nth(i);
   await locator.click();
 }
+
+test('Drag to select tagged text, and copy', async ({ page }) => {
```
doodlewind, line 459: 1. `withCtrlOrMeta` is a better fit here, since `Meta+B` won't trigger copy on Linux CI.
2. We don't use a capital letter at the beginning of a test case name :)
3. Suggestion on line break of a test case:

``` ts
test('drag to select tagged text and copy', async ({ page }) => {
  // put them together for init phrase
  await enterPlaygroundRoom(page);
  await initEmptyParagraphState(page);

  // separate statements by asserts, which generally represents a semantical operation
  await focusRichText(page);
  await page.keyboard.insertText('123456789');
  await assertRichTexts(page, ['123456789']);

  await dragBetweenIndices(page, [0, 1], [0, 3]);
  await withCtrlOrMeta(page, async (page: Page) => {
    await page.press('B');
  });
  await dragBetweenIndices(page, [0, 0], [0, 5]);
  await withCtrlOrMeta(page, async (page: Page) => {
    await page.press('B');
  });
  const textOne = await getSelectedTextByQuill(page);
  expect(textOne).toBe('12345');
});
```

The advice here applies other test cases :)

packages/blocks/src/page-block/default/utils.ts
```diff
@@ -401,3 +402,7 @@ async function getUrlByModel(model: BaseBlockModel) {
   const url = store?.get(model.sourceId);
   return url;
 }
+
+export function isControlledKeyBoardEvent(e: KeyboardEvent) {
```
doodlewind, line 406: `KeyBoard` -> `Keyboard` :)

tests/utils/actions/misc.ts
thorseraq: OK!

doodlewind (approved):
Nice fix!

tests/utils/actions/misc.ts
thorseraq: > // @ts-ignore and // @ts-expect-error do the exact same thing. They suppress the error the compiler would emit at the line below the comment.
But // @ts-ignore is very generic. It just says: "Compiler, ignore the next line."
But what did you exactly want to suppress?

tests/selection.spec.ts
thorseraq: Thx! Learned a lot :)

packages/blocks/src/page-block/default/utils.ts
thorseraq: They are reunioned now)

PR #312, opened 2022-12-15
feat: After pressing Enter, create a new block under the selected block

DiamondThree:


doodlewind:
The implementation looks good, but I guess this deserves a test case.

doodlewind (approved):
LGTM

PR #314, opened 2022-12-15
feat: adjust blob api with affine cloud

darkskygit:
Match blob sync api to newest affine cloud apis

Note1: Now we can access cloud api in same host based on vite proxy config, default disabled.
Note2: In the current backend design, the workspace id is a randomly generated int64 number, so if you need to test or enable blob synchronization, the provided workspace needs to be a number
Note3: current api server run at 100.77.180.48:11001


doodlewind (approved):
LGTM, a bit smelly though (

PR #316, opened 2022-12-15
fix: page title not sync readonly status

QiShaoXuan:


doodlewind (approved):
LGTM, do you need a new alpha?

PR #320, opened 2022-12-16
fix: point offset in edgeless default selection

himself65:


doodlewind (approved):
LGTM

himself65:
Now selection on the default page also has the same issue. I will fix it in the next PR

himself65:
I remove the padding on the body because It would cause test case to fail

himself65:
> I remove the padding on the body because It would cause test case to fail

Also, I think we should detect the range//xywh by the relative position, not the absolute position on the screen

doodlewind:
> > I remove the padding on the body because It would cause test case to fail
> 
> Also, I think we should detect the range//xywh by the relative position, not the absolute position on the screen

But `getBoundingClientRect` returns absolute position, and actually, we need to keep 2 coords systems in mind, one for model coords and one for final DOM absolute positions. Relative positions could be somewhat intermediate and interoperable with absolute ones.

PR #322, opened 2022-12-16
fix: replace focused blank paragraph on pasting image

DiamondThree:


packages/editor/src/managers/clipboard/paste-manager.ts
```diff
@@ -193,6 +194,15 @@ export class PasteManager {
         ) {
           parent &&
             this._addBlocks(blocks.slice(0), parent, index, addBlockIds);
+          const lastBlockModel = this._editor.page.getBlockById(lastBlock.id);
+          if (
```
doodlewind, line 198: Please add some comments for help understanding this complex `if`. Example:

``` ts
// On pasting image, replace the last focused paragraph instead of appending a new image block,
//  if this paragraph is empty.
```

This should apply to future code style if you are writing `if` with more than 3 expressions joined together.

DiamondThree: Ok, Thank you for your advice.

PR #327, opened 2022-12-16
fix: selection offset in default page

himself65:


himself65:
we need to improve test coverage in this case. I will do this later on this PR

himself65:
> we need to improve test coverage in this case. I will do this later on this PR

Also should test when the offset of the editor changes.

himself65:
test added /cc @doodlewind

doodlewind (approved):
LGTM

PR #329, opened 2022-12-16
feat: disable shadow dom of debug menu

linonetwo:
So overwriting style from the outside (for example, client app) is easier.

Move style to \<style> tag so it still takes effect.


himself65:
one question: why

doodlewind:
> one question: why

This allows for quick style adjusting in Tauri, we're prototyping in https://github.com/toeverything/devclient

PR #332, opened 2022-12-16
feat: support init with template

Himself65:
example: https://blocksuite-git-fork-himself65-featinitialdata-toeverything.vercel.app/?init=heavy


doodlewind (approved):
Looks very nice, I like it!

PR #338, opened 2022-12-17
build: export umd and esm module on `@blocksuite/store`

himself65:


himself65:
There're many issues that make these packages cannot be real production packages

himself65:
closing as this wont' be updated

PR #339, opened 2022-12-17
build: build both esm and umd

himself65:


himself65:
closing as this will not be updated

Issue #217, opened 2022-11-24
Wrong line break occurs during real-time collaboration


JimmFly:
I try to do real time collaboration in blocksuite. When I join the workspace, I press enter on the first line and a weird block appears. I think this is a bug and hopefully someone can fix it.

https://user-images.githubusercontent.com/102217452/203756249-16b750fb-f18f-4a76-bf71-2a71bc5327e2.mov

JimmFly:
<img width="1347" alt="898b2c704b47289fb4d0f2a14792d5e0" src="https://user-images.githubusercontent.com/102217452/203756529-4747c876-56be-4753-8b8d-f0c971930b04.png">

JimmFly:
Maybe I can try to handle it?

JimmFly:
> Maybe I can try to handle it myself?

Oh, I might not be able to handle this for some reason, can someone help me?

himself65:
I think the bug is the same as when I `Y.applyUpdate` directly https://github.com/toeverything/blocksuite/pull/318

himself65:
I think the reason is the state doesn't sync fully between users

Issue #287, opened 2022-12-12
Remove `idGenerator` config


doodlewind:
The `idGenerator` callback config for `Store` looks over-engineering. It's hard to imagine why another new ID generator (and a new customized algorithm) would be needed. A config to switch between auto-increment mode (this is a debug-only config) and UUIDv4 should be adequate.

doodlewind:
Implemented by @Himself65, thanks!

PR #342, opened 2022-12-17
refactor: id generator

Himself65:
I had a meeting with @doodlewind, and the conclusion is the error that the reason caused the collaborative mode and URL paste crash is the **default generator** in the playground. This will let all clients create a block starting from 0, and the editor will crash.

Fixes: https://github.com/toeverything/blocksuite/issues/217, https://github.com/toeverything/blocksuite/issues/287


doodlewind (approved):
LGTM

Issue #330, opened 2022-12-16
recalulate selected rect when resize/scroll the page


Himself65:

himself65:
https://user-images.githubusercontent.com/14026360/208166018-1699954f-e135-464b-a899-4e92f48d5f7f.mov

himself65:
https://user-images.githubusercontent.com/14026360/208190227-a6302371-2181-4f23-9707-53c6a6bbacd1.mov

himself65:
Related: https://github.com/toeverything/blocksuite/issues/616

PR #346, opened 2022-12-18
fix: bind `_clearSelection` to this

### LINKED ISSUE #345: Uncaught TypeError: Cannot read properties of undefined (reading 'state')
```bash
Uncaught TypeError: Cannot read properties of undefined (reading 'state')
    at HTMLDocument._clearSelection (default-page-block.ts:170:20)
```

himself65:
Fixes: https://github.com/toeverything/blocksuite/issues/345


himself65:
This PR also fixes https://github.com/toeverything/blocksuite/issues/330 in another way. But I think the correct behavior is not to cleanup the rect

doodlewind (approved):
LGTM

PR #349, opened 2022-12-18
fix: paste code sometimes go outside of the code block,  update UI of language selection

thorseraq:
Related to #352 #378


packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -285,7 +285,7 @@ export class DefaultPageBlockComponent extends LitElement implements BlockHost {
     });
 
     // TMP: clear selected rects on scroll
-    document.addEventListener('wheel', this._clearSelection);
+    document.addEventListener('wheel', this._clearSelection.bind(this));
```
doodlewind, line 288: I think the `_clearSelection` is refactored so `.bind` is no longer needed. Also, we should try to avoid `.call`/`.apply`/`.bind` in our codebase.

packages/blocks/src/code-block/components/lang-list.ts
```diff
@@ -229,49 +234,86 @@ export class LangList extends LitElement {
...
   }
 
+  protected updated() {
+    if (this.showLangList !== 'hidden') {
+      this.filterInput.focus();
+    }
+  }
+
+  protected firstUpdated() {
+    document.addEventListener('click', (e: MouseEvent) => {
```
doodlewind, line 244: Can we limit the scope of this listener to the `editor-container`?

doodlewind:
Also, I'm curious about when can we have test cases for the code block? Another PR for testing could be great :)

packages/blocks/src/code-block/components/lang-list.ts
thorseraq: If we need to detect the click outside code block language select container, add listener to document may be needed, but the handler will be removed after disposing, would this be viable?
https://github.com/toeverything/blocksuite/blob/d1b177466e5468a238f72ed0a1d38140af4b3f0f/packages/blocks/src/code-block/components/lang-list.ts#L258

I thought appending a fake dom of 100% width and height to `<body>`'s children, and add a listener to the fake dom. But since we can remove the handler of document, so I considered the current implementation finally)

packages/blocks/src/page-block/default/default-page-block.ts
thorseraq: OK!

thorseraq:
> Also, I'm curious about when can we have test cases for the code block? Another PR for testing could be great :)

OK! Test cases is more important than some remaining feature, I will add unit test first afterwards

packages/blocks/src/page-block/default/default-page-block.ts
himself65: this PR is the way of fixing `this` 

https://github.com/toeverything/blocksuite/commit/9f9739d75cc19afbdfd97b7d991f44722a4d3907

doodlewind (approved):
LGTM, thanks for passing the test!

PR #351, opened 2022-12-18
fix: enable ts check

himself65:


doodlewind (approved):
LGTM

himself65:
I found bugs in the new changes. We need a test for this part

packages/blocks/src/code-block/components/syntax-code-block.ts
```diff
@@ -102,15 +105,15 @@ class Syntax extends Module {
 
   constructor(quill: Quill, options: SyntaxCodeBlockOptions) {
     super(quill, options);
-    this.language = options.language;
+    this.language = options.lang;
```
thorseraq, line 108: Thanks! Now it corresponds to  type of `SyntaxCodeBlockOptions`, I think https://github.com/toeverything/blocksuite/blob/d1b177466e5468a238f72ed0a1d38140af4b3f0f/packages/blocks/src/code-block/code-block.ts#L110 need to be changed to `lang` as well.

himself65:
Due to quill doesn't have type declaration. So we use `any script`

doodlewind:
> Due to quill doesn't have type declaration. So we use `any script`

I think we have `@types/quill`

himself65:
> > Due to quill doesn't have type declaration. So we use `any script`
> 
> I think we have `@types/quill`

It's not fully. for example it doesn't have the type under `core/*`

PR #354, opened 2022-12-18
feat: init `@blocksuite/shared` package

himself65:


package.json
```diff
@@ -1,6 +1,9 @@
 {
   "name": "blocksuite",
   "private": true,
+  "keywords": [],
+  "license": "MPL-2.0",
+  "author": "toeverything",
```
himself65, line 6: I ran a `npx sort-package`

doodlewind (commented):
Not sure why the `shape.ts` module exists in the `@shared` package. Having shape-related modules in the `blocks` package looks fine.

himself65:
> Not sure why the `shape.ts` module exists in the `@shared` package. Having shape-related modules in the `blocks` package looks fine.

Once I need to import ShapeColor enum in next.js and currently the block package doesn't work well with next

himself65:
For example:

```diff
Subject: [PATCH] patch
---
Index: packages/app/src/components/edgeless-toolbar/index.tsx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/packages/app/src/components/edgeless-toolbar/index.tsx b/packages/app/src/components/edgeless-toolbar/index.tsx
--- a/packages/app/src/components/edgeless-toolbar/index.tsx	(revision 820d8d476aef2306016961585daa936ff84286da)
+++ b/packages/app/src/components/edgeless-toolbar/index.tsx	(date 1671414229291)
@@ -17,6 +17,7 @@
 import { Tooltip } from '@/ui/tooltip';
 import Slide from '@mui/material/Slide';
 import { useEditor } from '@/providers/editor-provider';
+import { ShapeColor } from '@blocksuite/shared/shape'
 
 const toolbarList1 = [
   {
@@ -50,7 +51,7 @@
         new CustomEvent('affine.switch-mouse-mode', {
           detail: {
             type: 'shape',
-            color: 'black',
+            color: ShapeColor.Black,
             shape: 'rectangle',
           },
         })

```

himself65:
I think the shared package should include all constant, enum, shared types, and environment independent functions by all modules

doodlewind:
Though I'm in favor of the `shared` package, IMO, it may be hard to be justified for sharing enums. Can we do some re-export in `blocks/index.ts` to reach this? Having the shape-related code encapsulated into the `shape-block` makes more sense to me. Otherwise, we may end up with a mess in `shared`.

The `std.ts` could be a good candidate since it lives in the `blocks` package, but the `store` needs it sometimes.

himself65:
> Though I'm in favor of the `shared` package, IMO, it may be hard to be justified for sharing enums. Can we do some re-export in `blocks/index.ts` to reach this? Having the shape-related code encapsulated into the `shape-block` makes more sense to me. Otherwise, we may end up with a mess in `shared`.
> 
> The `std.ts` could be a good candidate since it lives in the `blocks` package, but the `store` needs it sometimes.

Good point, I will look a way to that support `import '@blocksuite/blocks/std'`

himself65:
Closing this as we don’t really need this package

PR #359, opened 2022-12-18
build: compile all ts files

Himself65:
In this PR, we use `references` field to make sure cover all ts files


doodlewind (approved):
LGTM, it's very helpful covering type checks for test cases. Thanks!

PR #360, opened 2022-12-19
ci: upgrade github actions

himself65:


doodlewind (approved):
LGTM, thanks for maintaining this!

doodlewind:
I wonder if we can move the `tsc` type check part from the `Test` action to the `Lint` action. In this case, the `Lint` action could be renamed' Check'. Based on our current status, this could benefit the time cost of CI.

himself65:
> I wonder if we can move the `tsc` type check part from the `Test` action to the `Lint` action. In this case, the `Lint` action could be renamed' Check'. Based on our current status, this could benefit the time cost of CI.

That would possible and we can add cache on that to make it fater

PR #362, opened 2022-12-19
fix: `_scrollDisposable` is undefined

himself65:


himself65:
We should assert that there's no error occurs in the e2e test

doodlewind (approved):
LGTM

PR #365, opened 2022-12-19
 fix: enter shortcut on focusing embed block or its caption

DiamondThree:


himself65:
add a test case?

doodlewind:
Yes, and the case name could just be `enter shortcut on focusing embed block and its caption` 👀

DiamondThree:
> enter shortcut on focusing embed block and its caption
 thanks , that's a good idea,I will fix

doodlewind:
For quick iteration, we are going to merge this with the `housekeeping needed` label added. We'll plan a stage for cleaning up tech debts afterward.

PR #367, opened 2022-12-19
feat: improve tokenize result quality

darkskygit:
This pr will improve the tokenize quality of the existing word segmenter with latin words:
case insensitive
latin character normalizations
literal normalizations
morpheme normalizations


doodlewind (approved):
LGTM

PR #369, opened 2022-12-19
Fix: typo on format replace

lawvs:
Fix when editing the link text, it will split into pieces.

See more details:

https://user-images.githubusercontent.com/18554747/208391062-26c4450d-ec33-438e-b5e4-eff3a3d25266.mov

Related to https://github.com/toeverything/blocksuite/pull/148

Due to https://github.com/toeverything/blocksuite/pull/162

### DIFF
--- a/packages/blocks/src/__internal__/rich-text/rich-text.ts
+++ b/packages/blocks/src/__internal__/rich-text/rich-text.ts
@@ -101,7 +101,7 @@ export class RichText extends LitElement {
     this.quill.on('text-change', delta => {
       const selectorMap = {
         code: 'code',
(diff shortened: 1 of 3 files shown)


packages/blocks/src/__internal__/rich-text/rich-text.ts
```diff
@@ -101,9 +101,9 @@ export class RichText extends LitElement {
     this.quill.on('text-change', delta => {
       const selectorMap = {
         code: 'code',
-        link: 'link-code',
+        link: 'link-node',
```
lawvs, line 104: the key to the bug is a typo

PR #371, opened 2022-12-19
fix:image copy by click 

DiamondThree:


doodlewind (approved):
LGTM

packages/blocks/src/page-block/default/utils.ts
```diff
@@ -122,13 +122,25 @@ export async function downloadImage(model: BaseBlockModel) {
...
-  ]);
+
+  const newImage = new Image();
+  newImage.src = url;
+  newImage.onload = () => {
+    const canvas = document.createElement('canvas');
+    canvas.width = newImage.width;
+    canvas.height = newImage.height;
+    const ctx = canvas.getContext('2d');
+    ctx?.drawImage(newImage, 0, 0);
```
doodlewind, line 133: This will create a "deep copy" of the image. This can be merged for fast iteration, but I hope @darkskygit can take a look at what the best practice could be.

darkskygit: In any case, images that can be successfully loaded by img can be obtained by fetch() (this means that the corresponding link complies with cross-domain rules)If we use canvas screenshots, there will be slight differences in the original binary of each pasted image , which will invalidate the hash-based deduplication of the blob manager, and will also generate additional encoding and decoding overhead

darkskygit: simply say, it can be written like this:
``` js
fetch(url).then(r => r.blob()).then(blob => {
    // do something with blob
})
```

PR #372, opened 2022-12-19
fix:image download by click image-option

DiamondThree:


packages/blocks/src/page-block/default/utils.ts
```diff
@@ -108,15 +108,23 @@ function isPointIn(block: DOMRect, x: number, y: number): boolean {
...
+  const image = new Image();
+  imgSrc && (image.src = imgSrc);
+  image.setAttribute('crossOrigin', 'anonymous');
+  image.onload = function () {
+    const canvas = document.createElement('canvas');
+    canvas.width = image.width;
+    canvas.height = image.height;
+    const context = canvas.getContext('2d');
+    context && context.drawImage(image, 0, 0, image.width, image.height);
+    const url = canvas.toDataURL('image/png');
```
doodlewind, line 121: Images are blobs with our blob storage, why not convert the blob and export it using `File`? The canvas approach will erase all the metadata of the image, IMO that doesn't sound acceptable. See https://stackoverflow.com/questions/27159179/how-to-convert-blob-to-file-in-javascript

doodlewind (changes requested):

packages/blocks/src/page-block/default/utils.ts
DiamondThree: Because the file type was text when I  used blob download from storage.

doodlewind (approved):
For quick iteration, we are going to merge this with the `housekeeping needed` label added. We'll plan a stage for cleaning up tech debts afterward.

PR #358, opened 2022-12-18
fix: restore selection after format multiple lines

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/230

Also related to https://github.com/toeverything/blocksuite/discussions/231

https://user-images.githubusercontent.com/18554747/208319009-4a804023-4632-4d9d-8304-a8b0a5ffecad.mov

There still have some bugs, for example, the format will fail if there has an empty line

https://user-images.githubusercontent.com/18554747/208319163-5f02e066-91c0-4527-a40c-88163fa09b62.mov

Due to the id changed after `transformBlock`, it does not work at updateBlockType

PR #376, opened 2022-12-19
Fix: restore selection after transform block

lawvs:
- Restore selection after transform block

https://user-images.githubusercontent.com/18554747/209319944-db96fe6c-4350-4d10-84de-fe45e84dc6cc.mov

- Format bar will following current range always now.

- [x] Wait block update is too trick so we need a better solution.

Also mentioned at https://github.com/toeverything/blocksuite/pull/358

### DIFF
--- a/packages/blocks/src/__internal__/rich-text/rich-text.ts
+++ b/packages/blocks/src/__internal__/rich-text/rich-text.ts
@@ -103,7 +103,7 @@ export class RichText extends LitElement {
     // the character should not be inserted into the code or link node.
     // So we check and remove the corresponding format manually.
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    this.quill.on('text-change', delta => {
+    this.quill.on('text-change', (delta: any) => {
       const selectorMap = {
         code: 'code',
         link: 'link-node',
--- a/packages/blocks/src/__internal__/rich-text/rich-text.ts
+++ b/packages/blocks/src/__internal__/rich-text/rich-text.ts
@@ -91,7 +91,8 @@ export class RichText extends LitElement {
     });
 
     page.attachRichText(model.id, this.quill);
-    page.awareness.updateLocalCursor();
+    // TODO it can cause cursor flickering, by need to evaluate impact
+    // page.awareness.updateLocalCursor();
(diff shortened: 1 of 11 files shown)


packages/blocks/src/page-block/utils/container-operations.ts
```diff
@@ -73,22 +74,33 @@ export function deleteModels(page: Page, models: BaseBlockModel[]) {
...
+      // Replace selected block id
+      const blocks = selectedBlocks.filter(block => block.id === oldId);
+      // Because selectedBlocks maybe contains same block when only select one block, so we need to replace all of them
+      blocks.forEach(block => {
+        block.id = newId;
+      });
     }
   });
+  // XXX FIXME: wait for page update
+  await sleep(100);
```
himself65: Is there any signal for the page update?

packages/blocks/src/page-block/utils/container-operations.ts
```diff
@@ -73,22 +74,33 @@ export function deleteModels(page: Page, models: BaseBlockModel[]) {
...
   page.captureSync();
-  modelsInRange.forEach(model => {
+  modelsInRange.forEach(async model => {
     assertFlavours(model, ['affine:paragraph', 'affine:list']);
     if (model.flavour === flavour) {
       page.updateBlock(model, { type });
     } else {
-      transformBlock(page, model, flavour, type);
+      const oldId = model.id;
+      const newId = transformBlock(page, model, flavour, type);
```
himself65, line 120: Can transform block function returns a promise that we don't need `sleep`

packages/blocks/src/__internal__/rich-text/rich-text.ts
```diff
@@ -91,7 +91,8 @@ export class RichText extends LitElement {
     });
 
     page.attachRichText(model.id, this.quill);
-    page.awareness.updateLocalCursor();
+    // TODO it can cause cursor flickering, by need to evaluate impact
+    // page.awareness.updateLocalCursor();
```
lawvs, line 95: After removing the `asyncFocusRichText` after the transform block, the `updateLocalCursor` will cause cursor flickering after transform.

Is the update here necessary? @doodlewind

doodlewind (approved):
LGTM

PR #380, opened 2022-12-20
fix: Failed to resolve entry for package @blocksuite/store

linonetwo:
This bug happened after I upgrade from 0.3.0-alpha.7 to 0.3.0-alpha.13

This PR fix this when importing the package:

```
[ERROR] [plugin vite:dep-scan] Failed to resolve entry for package "@blocksuite/store". The package may have incorrect main/module/exports specified in its package.json.

    node_modules/.pnpm/esbuild@0.16.10/node_modules/esbuild/lib/main.js:1357:21:
      1357 │         let result = await callback({
```

I tried modify package.json in node_modules like this to fix it. But there may be better ways, like change this?

```json
  "publishConfig": {
    "access": "public",
    "main": "dist/index.js",
    "types": "dist/index.d.ts",
    "exports": {
      "./src/*": "./dist/*.js",
      ".": "./dist/index.js"
    }
  }
```

Seems you have changed the dist folder structure since 0.3.0-alpha.7 version

截屏2022-12-20 13 19 47

### DIFF
--- a/packages/store/package.json
+++ b/packages/store/package.json
@@ -2,7 +2,7 @@
   "name": "@blocksuite/store",
   "version": "0.3.0-alpha.13",
   "description": "BlockSuite data store built for general purpose state management.",
-  "main": "src/index.ts",
+  "main": "dist/store/src/index.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


linonetwo:
Emm, I have to change `publishConfig` instead.

PR #381, opened 2022-12-20
fix: package entry path after publish

linonetwo:
This bug happened after I upgrade from 0.3.0-alpha.7 to 0.3.0-alpha.13

This PR fix this when importing the package:

```
[ERROR] [plugin vite:dep-scan] Failed to resolve entry for package "@blocksuite/store". The package may have incorrect main/module/exports specified in its package.json.

    node_modules/.pnpm/esbuild@0.16.10/node_modules/esbuild/lib/main.js:1357:21:
      1357 │         let result = await callback({
```

I tried to modify package.json in devclient's node_modules like this to fix it.

Seems you have changed the dist folder structure of store package, since 0.3.0-alpha.7 version

截屏2022-12-20 13 19 47

And currently the index.d.ts is missing in the dist.

So a better solution is to restore the structure.

### DIFF
--- a/packages/store/package.json
+++ b/packages/store/package.json
@@ -39,11 +39,11 @@
   },
   "publishConfig": {
     "access": "public",
-    "main": "dist/index.js",
+    "main": "dist/store/src/index.js",
     "types": "dist/index.d.ts",
     "exports": {
-      "./src/*": "./dist/*.js",
-      ".": "./dist/index.js"
+      "./src/*": "./dist/store/src/*.js",
+      ".": "./dist/store/src/index.js"
     }
   }
 }


lawvs:
The struct of the published package is changed. I think it is related to https://github.com/toeverything/blocksuite/pull/359

Can you have a look @Himself65 ?

```
# before
dist/
  index.js
  xxx.js

# after
dist/
  store/
    index.js
    xxx.js
  blocks/
    index.js
```

lawvs (changes requested):
It's not a correct solution, please use 0.3.0-alpha.11 first, sorry for any inconvenience.

himself65:
> The struct of the published package is changed. I think it is related to https://github.com/toeverything/blocksuite/pull/359
> 
> 
> 
> Can you have a look @Himself65 ?
> 
> 
> 
> ```
> 
> # before
> 
> dist/
> 
>   index.js
> 
>   xxx.js
> 
> 
> 
> # after
> 
> dist/
> 
>   store/
> 
>     index.js
> 
>     xxx.js
> 
>   blocks/
> 
>     index.js
> 
> ```

The reason is I set baseUrl to . and this will compile all related packages

himself65:
The correct solution is use vite to build the package

lawvs:
Fix in https://github.com/toeverything/blocksuite/commit/9c20074a16cf8e6e0f9a6f4f079a9ce13edfdfc0

PR #383, opened 2022-12-20
fix: start backspace after list

### LINKED ISSUE #382: Some types of blocks cannot be entered at the beginning of the next line through Backspace
Cannot enter the Bulleted List, Numbered List, and To-do List at the beginning of the next line through Backspace

20221220153935_rec_

Flrande:
fix #382 

20221220174259_rec_

### DIFF
--- a/packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
+++ b/packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
@@ -168,7 +168,7 @@ export function handleLineStartBackspace(page: Page, model: ExtendedModel) {
       if (!parent || matchFlavours(parent, ['affine:group'])) {
         if (
           previousSibling &&
-          matchFlavours(previousSibling, ['affine:paragraph'])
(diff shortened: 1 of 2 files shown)


doodlewind:
Good catch! Although it may seem excessive, could you add a test case about this corner case? I think our test suite has covered all the utils that could help you write the operation sequence semantically. Thanks!

Flrande:
done

doodlewind (approved):
To speed up test case execution, merging these three cases into one case will be better (we can use `undo` to recover previous state), since the constant cost of opening a browser window is much more than doing common rich text editing operations. But that could be the content of another PR about optimization. let's land this one for now, thanks!

Flrande:
Do I need to create a pr for this optimization? I can do this work later.

doodlewind:
> Do I need to create a pr for this optimization? I can do this work later.

Just a suggestion, feel free to arrange it at your wish :)

Flrande:
All right.😇

PR #393, opened 2022-12-20
feat: use moduleResolution to `NodeNext`

himself65:


himself65:
re-assigned to @doodlewind

PR #400, opened 2022-12-21
mod: use webcrypto for blob

Austaras:


doodlewind (approved):
LGTM

cc @DiamondThree @darkskygit

PR #404, opened 2022-12-21
test: add test case for divider

JimmFly:
I think the divider function needs to add test cases to ensure its stability.


doodlewind (approved):
LGTM

PR #411, opened 2022-12-22
feat: update the select all shortcut key event

JimmFly:
Pressing Ctrl+A or Command+A now selects all blocks instead of text.


doodlewind:
It's a good start, I'll optimize based on your branch.

JimmFly:
Update the select all function. When there is focus, press Ctrl+A for the first time to select the current block, and press Ctrl+A for the second time to select all blocks. When there is no focus, Ctrl+A selects all blocks.

https://user-images.githubusercontent.com/102217452/209092266-e2a57329-06aa-4c3a-b130-692e1158b023.mov

doodlewind:
Closed in favor of #413, which includes some refactoring that makes this process simpler. Thanks for your initial work!

PR #413, opened 2022-12-22
feat: select-all hotkey with block level selection

doodlewind:
This continues from #411


tests/hotkey.spec.ts
```diff
@@ -37,13 +37,12 @@ test('rich-text hotkey scope on single press', async ({ page }) => {
...
 
-//TODO FIXME: select all feature is modified
-test.skip('single line rich-text inline code hotkey', async ({ page }) => {
+test('single line rich-text inline code hotkey', async ({ page }) => {
   await enterPlaygroundRoom(page);
   await initEmptyParagraphState(page);
   await focusRichText(page);
   await page.keyboard.type('hello');
-  await selectAllByKeyboard(page);
+  await dragBetweenIndices(page, [0, 0], [0, 5]);
```
doodlewind, line 45: @JimmFly we don't have to skip all these test cases 😄

JimmFly: > @JimmFly we don't have to skip all these test cases 😄

Sorry, I didn't read carefully the content of this test.

packages/blocks/src/page-block/utils/container-operations.ts
```diff
@@ -277,6 +281,19 @@ export function handleSelectAll() {
   resetNativeSelection(range);
 }
 
+export function handleSelectAll(selection: DefaultSelectionManager) {
+  if (selection.state.selectedBlocks.length === 0) {
+    const currentRange = getCurrentRange();
+    const rangeRect = currentRange.getBoundingClientRect();
+    selection.selectBlocksByRect(rangeRect);
+  } else {
+    const rect = document.body.getBoundingClientRect();
```
doodlewind, line 290: @JimmFly I added this API, which can be easier to work with :)

JimmFly: > @JimmFly I added this `selectBlocksByRect` API, which can be easier to work with :)

Oh, that seems more reasonable. Thanks for all you have done.

PR #419, opened 2022-12-23
fix: cannot paste outside the editor & skip divider when keyboard move

Himself65:


packages/editor/src/managers/clipboard/clip-event-dispatch.ts
```diff
@@ -47,15 +48,27 @@ export class ClipEventDispatch {
     );
   }
 
+  static containsEditorElement(composedPath: EventTarget[]): boolean {
+    return composedPath.reverse().some(element => {
```
himself65: why reverse? Because EditorContainer prefers on the bottom of the list

packages/editor/src/managers/clipboard/clip-event-dispatch.ts
```diff
@@ -54,19 +54,27 @@ export class ClipEventDispatch {
     });
   }
 
+  static path(event: ClipboardEvent): EventTarget[] {
+    if ('path' in event) {
+      // @ts-ignore
+      return event.path;
```
doodlewind, line 60: Can we use `event.path as unknown as EventTarget[]` to remove the `// @ts-ignore`?

himself65: no

doodlewind (approved):
LGTM, that's very speedy!

packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
```diff
@@ -245,6 +246,16 @@ export function handleKeyUp(model: ExtendedModel, editableContainer: Element) {
...
       !isAtLineEdge(range)
     ) {
+      if (preNodeModel) {
+        if (document.activeElement instanceof HTMLElement) {
+          if (!document.activeElement.closest('editor-container')) {
+            document.activeElement.blur();
+            const element = getBlockById(preNodeModel.id);
+            element?.focus();
+          }
+          console.log(document.activeElement);
```
lawvs, line 256:
```suggestion
```

JimmFly:
I don't think divider needs to be skipped, this is a requirement.

doodlewind:
> I don't think divider needs to be skipped, this is a requirement.

No worries, we've confirmed with the product team about this compromise.

PR #422, opened 2022-12-23
fix: select all fails when out of focus or focusing on the title

JimmFly:


himself65:
Does this relate to #425 ?

doodlewind (approved):
LGTM

PR #427, opened 2022-12-24
fix: yjs import multiple times for next.js

Himself65:


doodlewind (approved):
LGTM, thanks for all the efforts fixing this!

PR #429, opened 2022-12-24
build: use `dist` as exports

Himself65:


lawvs:
I prefer https://github.com/toeverything/blocksuite/pull/428

PR #433, opened 2022-12-25
fix: css might not work as expected

himself65:


doodlewind (approved):
LGTM, this really helps!

himself65:
Related comment: https://github.com/lit/lit/issues/1977#issuecomment-872645551

PR #448, opened 2022-12-26
feat: disable markdown shortcut in code block, code block option, hover option update

thorseraq:
1. disable markdown shortcut in code block
2. hover option now will not disappear after clicking
3. line wrap now has a activated state


packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
```diff
@@ -328,6 +328,9 @@ export function tryMatchSpaceHotkey(
   if (offset > prefix.length) {
     return ALLOW_DEFAULT;
   }
+  if (model.flavour === 'affine:code') {
```
doodlewind, line 331: `matchFlavour` would be appreciated here

doodlewind (approved):
LGTM! Thanks for these detailed improvements!

packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
thorseraq: Thanks! Applied

PR #454, opened 2022-12-26
refactor: enhance yjs type in store

himself65:


.github/workflows/test.yaml
```diff
@@ -39,3 +39,22 @@ jobs:
           name: test-results
           path: ./test-results
           if-no-files-found: ignore
+
+  test-unit:
+    name: unite test
```
doodlewind, line 44: `unite` -> `unit`  👀

Also, can we run `pnpm test:store` in the root directory, instead of `pnpm test:unit` in `packages/store`? This can include custom store E2E tests instead of unit tests.

himself65: I thought you don't wanna run the e2e test under store

packages/store/src/workspace/page.ts
```diff
@@ -14,21 +14,30 @@ import type { IdGenerator } from '../utils/id-generator.js';
...
+}
+
+export type YBlock = EnhancedYMap<YBlockData>;
+export type YBlocks = EnhancedYMap<Record<string, YBlock>>;
 
 /** JSON-serializable properties of a block */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type BlockProps = Record<string, any> & {
-  id: string;
+  // id: string;
```
himself65, line 39: I think we shouldn't let use to pass an id though props. because it is random generated

packages/store/src/workspace/workspace.ts
```diff
@@ -30,45 +39,39 @@ class WorkspaceMeta extends Space {
...
-    if (!this._yMetaRoot.has('avatar')) {
-      return null;
-    }
-    return this._yMetaRoot.get('avatar') as Y.Text;
+    return this._yMetaRoot.get('versions');
   }
 
   get name() {
-    return this._yName ? this._yName.toString() : '';
+    return this._yMetaRootProxy.name
```
doodlewind, line 66: Can we directly return `this._yMetaRootProxy.name`?

patches/yjs@13.5.43.patch
```diff
@@ -0,0 +1,217 @@
+diff --git a/dist/src/types/YMap.d.ts b/dist/src/types/YMap.d.ts
+index 00b8e8178827727172ed32930e6c76c1b1842f64..b8daa1898734caf05c2fe5807a98753978615e4d 100644
+--- a/dist/src/types/YMap.d.ts
++++ b/dist/src/types/YMap.d.ts
```
doodlewind, line 4: If other future y-provider packages (which may be maintained by us) depend on a non-patched version, will this lead to two different versions of yjs?

Or else maybe we can make these packages depend on `@blocksuite/store`, and get yjs from the static getter at `Workspace.Y`?

himself65: if downstream use Workspace.Y, they dont have this patch

himself65:
I'm gonna split this into different PRs to make sure it wont break anything

Issue #334, opened 2022-12-16
Bundle size significantly bloated by `highlight.js`


doodlewind:
In #255 we added `highlight.js` as a dependency. This leads to a significant bloat in bundle size.

Before:

```
packages/playground build$ tsc --project ./tsconfig.json && vite build
[2 lines collapsed]
│ Use of eval is strongly discouraged, as it poses security risks and may cause issues with minification
│ ✓ 273 modules transformed.
│ rendering chunks...
│ dist/index.html                  0.85 KiB
│ dist/assets/index.74233d91.css   8.83 KiB / gzip: 2.14 KiB
│ dist/assets/index.6d7e3a8c.js    965.16 KiB / gzip: 273.16 KiB
```

After:

```
packages/playground build$ tsc --project ./tsconfig.json && vite build
[2 lines collapsed]
│ Use of eval is strongly discouraged, as it poses security risks and may cause issues with minification
│ ✓ 479 modules transformed.
│ rendering chunks...
│ dist/index.html                  0.85 KiB
│ dist/assets/index.9044078f.css   12.04 KiB / gzip: 2.85 KiB
│ dist/assets/index.93f52fc6.js    2028.87 KiB / gzip: 577.41 KiB
```

@thorseraq can we optimization on this?

himself65:
I think in this PR, I've reduced the bundle size to about 270kb

doodlewind:
> I think in this PR, I've reduced the bundle size to about 270kb

The bundle size `@blocksuite/blocks` is intermediate, we need to also measure the size in `@blocksuite/playground`.

thorseraq:
OK! I’ll take a survey, see if we can reduce the bundle size increased by introducing highlight.js

lawvs:
We can further reduce the size of the highlight by registering only part of the language

See https://github.com/highlightjs/highlight.js#es6-modules--import

<img width="604" alt="Screenshot 2022-12-28 at 1 59 15 PM" src="https://user-images.githubusercontent.com/18554747/209765267-6d215f69-d5c6-414f-adbb-74315be29f93.png">

PR #455, opened 2022-12-27
feat: add lazy loading for highlight

lawvs:
fix https://github.com/toeverything/blocksuite/issues/334

- [x] fix tests

before
Screenshot 2022-12-27 at 12 55 34 PM

after
Screenshot 2022-12-27 at 12 55 10 PM

Screenshot 2022-12-27 at 3 43 21 PM

### DIFF
--- a/packages/blocks/src/code-block/code-block.ts
+++ b/packages/blocks/src/code-block/code-block.ts
@@ -10,9 +10,9 @@ import {
   BlockHost,
   NonShadowLitElement,
 } from '../__internal__/index.js';
-// @ts-ignore
-import highlight from 'highlight.js';
 import { ArrowDownIcon } from '../components/format-quick-bar/icons.js';
+
+let highlighGlobal: unknown = null;
 
 @customElement('affine-code')
 export class CodeBlockComponent extends NonShadowLitElement {
@@ -48,8 +48,22 @@ export class CodeBlockComponent extends NonShadowLitElement {
   @state()
   filterText = '';
 
-  @property()
-  delay = 150;
+  @state()
+  highlight: unknown;
+
+  connectedCallback() {
+    super.connectedCallback();
+    if (highlighGlobal) {
+      this.highlight = highlighGlobal;
+      return;
+    }
+    // @ts-ignore highlight.js has no types
+    import('highlight.js').then(highlightModule => {
+      const highlight = highlightModule.default.highlight;
+      this.highlight = highlight;
+      highlighGlobal = highlight;
+    });
+  }
 
   firstUpdated() {
     this.model.propsUpdated.on(() => this.requestUpdate());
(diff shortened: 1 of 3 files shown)


thorseraq (approved):

packages/blocks/src/code-block/code-block.ts
```diff
@@ -84,19 +98,22 @@ export class CodeBlockComponent extends NonShadowLitElement {
...
+                  highlight: this.highlight,
+                  codeBlockElement: this,
+                  language: this.model.language,
+                },
+              }}
+            >
+              <div id="line-number"></div>
+            </rich-text>`
+          : // TODO update loading state
+            `Loading coding block...`}
```
himself65, line 116: The loading state should be handled by page block.
Should we have something like `Suspense` in React? /cc @doodlewind

doodlewind: Can we use the [async rendering](https://lit.dev/docs/templates/directives/#asynchronous-rendering) directives in Lit?

himself65: wow this is great

packages/blocks/src/code-block/code-block.ts
```diff
@@ -10,10 +10,10 @@ import {
   BlockHost,
   NonShadowLitElement,
 } from '../__internal__/index.js';
-// @ts-ignore
-import highlight from 'highlight.js';
 import { ArrowDownIcon } from '../components/format-quick-bar/icons.js';
 
+let highlighGlobal: unknown = null;
```
himself65, line 15:
```ts
const highlightPromise = import('highlight.js').then(highlightModule => {
  const highlight = highlightModule.default.highlight;
  return highlight
});
// ...
@state()
highlight = highlightPromise

until(this.highlight, html`xxx`)
```

packages/blocks/src/code-block/code-block.ts
lawvs: The until directive will cause the UI to always have to wait for a trick because it returns nothing if the argument is not a promise.

So I will keep the existing code style.

tests/code.spec.ts
```diff
@@ -155,8 +157,9 @@ test('keyboard selection and copy paste', async ({ page }) => {
 
   await focusRichText(page);
   await addCodeBlock(page);
-  await page.keyboard.type('use');
+  await page.waitForSelector('rich-text');
```
doodlewind, line 160: I think we can always do the `waitForSelector` inside `addCodeBlock`

packages/blocks/src/code-block/code-block.ts
Himself65, line 15: I think you can just import('highlight.js') here

Himself65 (approved):

doodlewind (approved):

lawvs:
@thorseraq will continue to improve the loading style and test, thank you very much for his help.

PR #460, opened 2022-12-28
refactor: workspace meta

Himself65:


packages/store/src/workspace/workspace.ts
```diff
@@ -15,76 +16,58 @@ export interface PageMeta {
...
-    return this._yName ? this._yName.toString() : '';
+    return this.proxy.name;
   }
 
   get avatar() {
-    return this._yAvatar ? this._yAvatar.toString() : '';
+    return this.proxy.avatar;
   }
 
   setName(val: string) {
```
septs, line 57:
```suggestion
  setName(name: string) {
```

suggest, rename variable name to `name`

doodlewind (approved):
LGTM

PR #464, opened 2022-12-28
fix: prevent click event from making selection lost

lawvs:
This pull request fixes the issue with selecting losing when a user clicks the edge of the format bar.


doodlewind (approved):
LGTM

PR #481, opened 2022-12-29
fix: click elsewhere of language selector, language focus background does not disappear

thorseraq:


doodlewind (approved):
LGTM, thanks for unifying the convention!

PR #483, opened 2022-12-29
feat: add web-vitals in playground

Himself65:


doodlewind (changes requested):
My (subjective) concern is that the console data may be distracting in localhost build, and I'd rather keep the `main.ts` simple.

Also we use the `report-web-vitals.ts` convention as our module name. This doesn't have to apply to the Next example though.

PR #493, opened 2022-12-30
feat: support multiline in header

Himself65:


doodlewind (approved):
LGTM, thanks for the fix!

PR #494, opened 2022-12-30
fix: placeholder in default title

himself65:


himself65:
Related: https://github.com/toeverything/AFFiNE/actions/runs/3806007857/jobs/6474449744

PR #498, opened 2022-12-30
feat: run store testing if package change

### LINKED ISSUE #402: Run `pnpm test:store` in CI
Currently, we only run the `pnpm test` and `pnpm build` on CI. The test cases in the store package are ignored. It would be better to run `pnpm test:store` on CI, which can be a standalone check to improve CI speed.

darkskygit:
fix #402

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -9,8 +9,8 @@ on:
       - master
 
 jobs:
-  test:
-    name: test
+  main-test:
+    name: Unit test for main packages
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


Himself65 (changes requested):
Id like to run it in another yml file

darkskygit:
> Id like to run it in another yml file

what is your concern? time or maintainability

doodlewind:
Yes, I think having another YML entry can help make testing faster.

darkskygit:
> Yes, I think having another YML entry can help make testing faster.

We only need to split jobs to let the test run in parallel. Writing the similar type of actions in the same yaml helps us maintain it easily.

Action config now renew to parallel run test.

doodlewind (approved):
LGTM

Himself65 (approved):

PR #506, opened 2023-01-01
build: ignore HMR mode in dev

himself65:


doodlewind (commented):
Should this apply to all of the `store`, `blocks` and `editor` packages?

himself65:
> Should this apply to all of the `store`, `blocks` and `editor` packages?

I'm still investigating whether this could fixes affine debug log issue

PR #514, opened 2023-01-03
feat: support switch edgeless display mode

Himself65:


packages/blocks/src/__internal__/utils/types.ts
```diff
@@ -90,10 +90,12 @@ export type ShapeMouseMode = {
 };
 
 export type MouseMode = DefaultMouseMode | ShapeMouseMode;
+export type EdgelessDisplayMode = 'default' | 'grid';
```
doodlewind, line 93: We should be very cautious to introduce different "modes". Having `EdgelessConfig.showGrid: boolean` or `EdgelessDisplayOptions.showGrid` sounds better to me. In this case, we can define this field as a boolean rather than an enum.

doodlewind:
Also we can enable this grid config by default 👻

doodlewind:
Also we can enable this grid config by default 👻

doodlewind (approved):

PR #515, opened 2023-01-03
fix: when image size too large, option operation should  show inside

DiamondThree:


lawvs (approved):
You should add some comments to keep the code readable

packages/blocks/src/page-block/default/selection-manager.ts
```diff
@@ -409,6 +409,7 @@ export class DefaultSelectionManager {
 
     if (hoverEditingState?.model.type === 'image') {
       const { position } = hoverEditingState;
+      // when image size is too large, the option popup should show inside
```
doodlewind, line 412: /cc @DiamondThree

PR #519, opened 2023-01-03
fix: keep image meta fields in clipboard

DiamondThree:


packages/blocks/src/page-block/default/utils.ts
```diff
@@ -127,29 +128,67 @@ export async function downloadImage(model: BaseBlockModel) {
...
+}
+interface clipItem {
+  mimeType: string;
+  data: string;
+}
 
-  const newImage = new Image();
-  newImage.src = url;
-  newImage.onload = () => {
-    const canvas = document.createElement('canvas');
```
doodlewind, line 137: I thought the `downloadImage` function in this module should also be refactored.

DiamondThree: But this pr is about downloading, and I think we can open a new branch to solve this problem.

Issue #510, opened 2023-01-02
YDoc format compatibility test


doodlewind:
We need to ensure that after the [Abbey Wood](https://affine.pro/content/blog/New%20Year%20Gift-%20Alpha%202.0!%20More%20editorial%20features%20online./index) version in AFFiNE, the editor should be compatible with the user-generated data. For debugging we have added `workspace.exportYDoc()` in #509, or we can paste in the following snippet in the browser console:

```
function dumpYDoc() {
  const Y = workspace.constructor.Y;
  const binary = Y.encodeStateAsUpdate(workspace.doc);
  const file = new Blob([binary], { type: "application/octet-stream" });
  const fileUrl = URL.createObjectURL(file);

  const link = document.createElement("a");
  link.href = fileUrl;
  link.download = "workspace.ydoc";
  link.click();

  URL.revokeObjectURL(fileUrl);
}

dumpYDoc();
```

For now, some test binaries are manually created here:

[230102-abbey-wood.ydoc.zip](https://github.com/toeverything/blocksuite/files/10330127/230102-abbey-wood.ydoc.zip)

PR #520, opened 2023-01-03
refactor: `affine:group` -> `affine:frame`

doodlewind:
* [x] cleanup `group` concept
* [x] data compatibility & migration
* [x] unit test


himself65 (changes requested):
this is a breaking change. could you please try to make forward compatibility

lawvs:
I plan to implement an `assertDocMatchJSX` similar to `assertStoreMatchJSX` function but ignores all elements of edgeless( frame and shape), what do you think?

doodlewind:
> this is a breaking change. could you please try to make forward compatibility

Certainly, we need to do the `group` -> `frame` conversion in `syncFromExistingDoc`, and I'll use the binary in #510 for local testing.

doodlewind:
> I plan to implement an `assertDocMatchJSX` similar to `assertStoreMatchJSX` function but ignores all elements of edgeless( frame and shape), what do you think?

Will this feature require having YDoc binary as expected state? Since the binary is not readable, I'm not sure if committing YDoc binaries into our repo is an good idea.

himself65:
> > this is a breaking change. could you please try to make forward compatibility
> 
> Sure, we need to do the `group` -> `frame` conversion in `syncFromExistingDoc`, and I'll use the binary in #510 for local testing.

we could put a basic old version binary into test fixtures, and do a unit test into store

doodlewind:
@thorseraq is this test case flaky? https://github.com/toeverything/blocksuite/actions/runs/3835526880/jobs/6528878731

thorseraq:
> @thorseraq is this test case flaky? https://github.com/toeverything/blocksuite/actions/runs/3835526880/jobs/6528878731

end position of mouse drag seems exceeding text area in CI environment

himself65 (approved):

PR #521, opened 2023-01-03
fix: make shape block non-shadow element

himself65:


himself65:
Waiting for https://github.com/vercel/next.js/issues/44534

himself65:
use another workaround in other PRs

PR #534, opened 2023-01-04
feat: init handwriting mode

himself65:


himself65:
this feature has re-assigned to @doodlewind

PR #535, opened 2023-01-04
feat: save/read from local file

himself65:


doodlewind (changes requested):
I think it's currently a non-goal for us to use WASM-based SQLite in our Web version. We intend to use JSBridge-based PostgresDB (that runs natively in OctoBase) for local persistence in our desktop client.

packages/playground/src/components/debug-menu.ts
```diff
@@ -203,6 +217,38 @@ export class DebugMenu extends LitElement {
...
+    }
+    const input = <SlInput>document.getElementById('input-file-name');
+    const value = input.value;
+    const filename = value.endsWith('.affine') ? value : `${value}.affine`;
+    getFileHandle(filename, this.fsHandle).then(async handle => {
+      if (!handle) {
+        handle = await createFileHandle(filename, this.fsHandle);
+      }
+      if (handle) {
+        await writeContentToFile(handle, Workspace.toLocal(this.workspace));
```
doodlewind, line 244: Seems that we only serialize YDoc to disk for now?

himself65: yep, im still working on that

himself65: fixed

himself65:
use https://github.com/GoogleChromeLabs/browser-fs-access#readme

doodlewind:
Closing this since we are going to work on file format at https://github.com/toeverything/OctoBase/issues/173. Before the OctoBase implementation, we shouldn't support inner format in BlockSuite that may make migration harder.

PR #537, opened 2023-01-04
fix: select all with hot key at the start of the line

Flrande:
fix bug: CMD-A loses the cursor when the cursor is at the beginning of the line


doodlewind (approved):
Good catch!

PR #539, opened 2023-01-04
refactor: select code block

lawvs:
- merge similar logic
- now, when a code block is selected with the keyboard, the code block can be deleted with just one keystroke.

### DIFF
--- a/packages/blocks/src/page-block/default/selection-manager.ts
+++ b/packages/blocks/src/page-block/default/selection-manager.ts
@@ -74,7 +74,7 @@ function createSelectionRect(
   return new DOMRect(left, top, width, height);
 }
 
-type PageSelectionType = 'native' | 'block' | 'none' | 'embed' | 'focus';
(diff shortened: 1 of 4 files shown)


thorseraq:
Thx! But I find that after dragging to  select code block, we can't delete it using backspace. The Unit Test

https://github.com/toeverything/blocksuite/blob/68e2fcc3e4a8e3069d6aa95ea64a2963cd180af3/tests/code.spec.ts#L207

has passed, is also weird, I think it's maybe caused by the cursor remaining inside the quill context.

packages/blocks/src/page-block/default/selection-manager.ts
```diff
@@ -74,7 +74,7 @@ function createSelectionRect(
   return new DOMRect(left, top, width, height);
 }
 
-type PageSelectionType = 'native' | 'block' | 'none' | 'embed' | 'focus';
+type PageSelectionType = 'native' | 'block' | 'none' | 'embed';
```
doodlewind, line 77: This is great!

PR #540, opened 2023-01-05
fix: paste at the start of line

Flrande:
fix: The text will be inserted from the end of the line when pasting at the start of a line


doodlewind (approved):
Good catch!

PR #541, opened 2023-01-05
fix: use `animationend` event

Himself65:


doodlewind (approved):

lawvs (commented):
This change did not work, the `animationend` event is not triggered. You should use `addEventListener('transitionend', (event) => {});`


https://user-images.githubusercontent.com/18554747/210707685-16208dce-ae55-470f-a28f-775782b82b31.mov

PR #543, opened 2023-01-05
feat: support drag handle in default page

Himself65:


packages/blocks/src/page-block/default/selection-manager.ts
```diff
@@ -420,6 +423,60 @@ export class DefaultSelectionManager {
...
+              ) as BaseBlockModel;
+              const idx = currentParentModel.children.findIndex(
+                m => m.id === currentModel.id
+              );
+              const nextIdx = nextParentModel.children.findIndex(
+                m => m.id === nextModel.id
+              );
+              if (
+                recursiveFindParent(this.page, currentModel, nextModel) ||
+                recursiveFindParent(this.page, nextModel, currentModel)
```
colelawrence, line 453: Break this out into a named value like `const isParent___ = ` and check with `if (isParent___) return;`

(I don't know what this is checking, so obviously `isParent___` is just a stand in.

packages/blocks/src/page-block/default/selection-manager.ts
```diff
@@ -420,6 +423,60 @@ export class DefaultSelectionManager {
...
+              ) as BaseBlockModel;
+              const nextParentModel = this.page.getParent(
+                nextModel
+              ) as BaseBlockModel;
+              const idx = currentParentModel.children.findIndex(
+                m => m.id === currentModel.id
+              );
+              const nextIdx = nextParentModel.children.findIndex(
+                m => m.id === nextModel.id
+              );
```
colelawrence: Looks like all these lines can be moved after the `if (
                recursiveFindParent(this.page, currentModel, nextModel) ||
                recursiveFindParent(this.page, nextModel, currentModel)` check

packages/store/src/workspace/page.ts
```diff
@@ -276,6 +276,28 @@ export class Page extends Space<PageData> {
     this.updateBlock(model, props);
   }
 
+  moveBlock(
+    model: BaseBlockModel,
+    oldParent: BaseBlockModel,
+    oldIndex: number,
+    newParent: BaseBlockModel,
+    newIndex: number
+  ) {
```
colelawrence: Lots of positional arguments of the same types is usually a code smell. I'll defer to @doodlewind and other code owners for whether this is actually aligned with the codebase's maintainability guidelines.

Alternatively, pass in an
```
  opts: {
    target: BaseBlockModel,
    from: { parent: BaseBlockModel, idx: number },
    to: { parent: BaseBlockModel, idx: number }
  }
```

colelawrence (commented):
A few comments on different TypeScript patterns. I do not know all the parts of this codebase, so feel free to discard any of my ideas.

doodlewind (approved):
Awesome! we can leave future improvements in following PRs 🍗

PR #555, opened 2023-01-06
fix: cursor lose outside container

Flrande:
fix: Cursor lost when clicking on the blank position on the left and right side of the page

### DIFF
--- a/packages/blocks/src/__internal__/utils/gesture.ts
+++ b/packages/blocks/src/__internal__/utils/gesture.ts
@@ -1,3 +1,6 @@
+import { restoreSelection, saveBlockSelection } from './selection.js';
+import type { SelectedBlock } from './types.js';
+
 export interface IPoint {
   x: number;
   y: number;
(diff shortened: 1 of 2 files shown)


doodlewind (approved):
It's a very detailed but significant improvement to the experience, and it's great!

Next challenge: can we *reposition* the cursor on outside click (rather than keeping the cursor at the previous position)? For example, for following 3 paragraphs:

```
123
456
789
```
 
Clicking at the left of `456` will position the cursor at `|4`, and clicking at the right of `456` will position the cursor at `6|`. This is what happens in Notion.

doodlewind:
Also, adding a test case for this feature could be great. Otherwise, it will possibly be *you* who has to look into future regression (rather than the one who breaks your code).

Flrande:
I will add a test after completing this improvement because the current fix in PR conflicts with the improvement just mentioned. If I fail to complete this improvement today, I will first add a test for this fix. :)

PR #558, opened 2023-01-06
test: add test case on drag handle

Himself65:


tests/drag.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+    return { x: box.right, y: box.bottom - 2 };
+  }, []);
+
+  await page.mouse.move(topLeft.x, topLeft.y);
+  const length1 = await page.evaluate(() => {
+    const handles = document.querySelectorAll('affine-drag-handle');
+    return handles.length;
+  }, []);
+  expect(length1).toBe(1);
```
lawvs, line 35: you can use the locator API

```ts
const dragHandle = await page.locator('affine-drag-handle')
await expect(dragHandle).toHaveCount(1);
```

https://playwright.dev/docs/locators

Issue #545, opened 2023-01-05
Support Shift Select


ShenQingchuan:
I did some competitive research.

- Notion/Wolai/Feishu support when user press Shift and use arrow keys to update selection.
- But AFFiNE (BlockSuite) doesn't support.

Notion and Wolai use the 'anchor' side of Selection to locate the float menu's position. whatever you change the selection range, the float menu stays still. 

- In Notion, if you make the selection area scrolled out of view, you will see it's fading out, and if you scroll back, it will fade in again.
- In Wolai, it has the same logic but without the fade animation
- In Feishu, it doesn't has this feature, the float menu always stay still when scrolling out of view.

<details>
  <summary> 中文 </summary>

竞品调研了一下：
- Notion/Wolai/飞书云文档   都支持了 键盘制造选区之后 呼出菜单栏
- 很遗憾 AFFiNE 还没有呢 😂  估计 BlockSuite 也有同样的问题

上述三种竞品支持的情况是
- Notion 和 Wolai 都是针对 文档的 Anchor 也就是选取不动的那一边的 绝对位置 进行浮动 Menu 的定位的，所以无论你是向上选择、向下选择，无论怎么拉大拉小选区，浮动菜单的位置都暂时是不变的。（我还特意做好事者看了一下，在 Notion 里，如果滚动页面让选区离开画面，浮动 Menu 是 Fixed 并且过 100ms 左右渐隐消失，而滚动回来又会显示 .... 体验很好，不得不说是这波非常细了。 Wolai 是有这个逻辑，但没有渐隐的动画，也差不多 ）
- 飞书好像有点傻，是每次选区变化后都重新刷新一下位置、看得出会闪一下

</details>

doodlewind:
Thanks for your feedback! This is in the plan and should be a technically low-handing fruit to BlockSuite. We would let you know about future progress.

Flrande:
It works fine now.

PR #569, opened 2023-01-06
feat: show format bar on selection change

Innei:
#545


himself65 (commented):
I don't suggest to bind event on `document` as it will affect the upstream app like AFFiNE

doodlewind:
For making the preview available, the Vercel error is the same as the errors in the `Linter` job.

Innei:
> I don't suggest to bind event on `document` as it will affect the upstream app like AFFiNE

Seem event `selectionchange` only work on document.

himself65:
> > I don't suggest to bind event on `document` as it will affect the upstream app like AFFiNE
> 
> 
> 
> Seem event `selectionchange` only work on document.

I see, so we should take care of this

packages/blocks/src/__internal__/utils/std.ts
```diff
@@ -109,6 +109,38 @@ export const throttle = <
...
+  limit: number,
+  { leading = true, trailing = true } = {}
+): T => {
+  let timer: ReturnType<typeof setTimeout> | null = null;
+  let lastArgs: Args | null = null;
+
+  const setTimer = () => {
+    if (lastArgs && trailing) {
+      fn(...lastArgs);
+      lastArgs = null;
```
himself65, line 126: `lastArgs` is always null

packages/blocks/src/__internal__/utils/gesture.ts
```diff
@@ -236,16 +238,26 @@ export function initMouseEventHandlers(
...
+    }
+
+    onSelectionChange(e as Event);
+  }, 100);
+
   container.addEventListener('mousedown', mouseDownHandler);
   container.addEventListener('mousemove', mouseMoveHandler);
   container.addEventListener('contextmenu', contextMenuHandler);
   container.addEventListener('dblclick', dblClickHandler);
+  document.addEventListener('selectionchange', selectionChangeHandler);
```
doodlewind, line 253: This looks a bit overkill, since the input elements outside of BlockSuite may be unexpectedly handled. Why not add this listener to the `container`?

Innei: Because event selectionchange only work on document.🥲

doodlewind:
@lawvs Seems that the only failing case is in `format-quick-bar.spec.ts`, do we have some chance to land this? Thanks.

lawvs:
This implementation still has many areas that need to be improved (It will frequently call `showFormatQuickBar` when the selection change), but it works now.

@Innei Do you have time to add some tests or fix quick-bar directions? (incorrect direction when selecting from back to front, you can get the direction based on `window.getSelection()` in most cases)

Innei:
> window.getSelection()

`window.getSelection()` seem can not judge the direction of mouse selection, but do not affect feature.

himself65 (approved):
LGTM

doodlewind (approved):

doodlewind:
Thanks for the contribution!

PR #571, opened 2023-01-06
feat: add slash menu

lawvs:
Add `slash` to url params to enable feature

for example `http://localhost:5173/?init&slash`


packages/blocks/src/components/slash-menu/slash-menu-node.ts
```diff
@@ -0,0 +1,238 @@
+import { getRichTextByModel } from '@blocksuite/blocks/std.js';
```
himself65: don't import itself

Himself65 (changes requested):

packages/blocks/src/components/slash-menu/slash-menu-node.ts
lawvs: Blame the incorrect auto import!

himself65: We could have a eslint rule to avoid this

himself65: See https://github.com/toeverything/blocksuite/pull/762

lawvs:
I think it is ready for review.

CI fails due to https://github.com/toeverything/blocksuite/pull/777, I will rebase after it is merged.

Himself65 (changes requested):
fix the build error

packages/blocks/src/components/slash-menu/index.ts
```diff
@@ -0,0 +1,134 @@
...
+}: {
+  model: BaseBlockModel;
+  range: Range;
+  direction?: DragDirection;
+  container?: HTMLElement;
+  abortController?: AbortController;
+}) => {
+  // Abort previous format quick bar
+  globalAbortController.abort();
+  globalAbortController = abortController;
```
himself65, line 112: I think we could use a singleton pattern on this too, like drag-handle

PR #580, opened 2023-01-06
fix: keep cursor to title when deleting at the start of the first block

### LINKED ISSUE #567: Keep cursor to title when deleting at the start of the first block
Can the cursor be moved to the end of the title by pressing backspace at the beginning of the first line?（google translate）

<details>
<summary>中文</summary>
<br/>
在第一行的行头按下backspace光标能否移动到标题尾部。

</details>

# notion

https://user-images.githubusercontent.com/83949429/210953495-85e90cbd-8d65-414a-acca-a698a9953ca7.mp4

# feishu

https://user-images.githubusercontent.com/83949429/210953501-aee5ba73-b659-4d02-89e1-2847fb3cb31e.mp4

# blocksuite

https://user-images.githubusercontent.com/83949429/210953544-447317ff-78b9-48f5-9fc4-66d962e69037.mp4

zoulamZero:
Fixes #567

### DIFF
--- a/packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
+++ b/packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
@@ -1,7 +1,6 @@
 // operations used in rich-text level
 
 import { Page, Text } from '@blocksuite/store';
-import autosize from 'autosize';
 import type { Quill } from 'quill';
 import {
   ExtendedModel,
@@ -21,16 +20,8 @@ import {
   focusPreviousBlock,
   focusBlockByModel,
   supportsChildren,
-  getPageTitle,
+  getModelByElement,
 } from '../utils/index.js';
-
-function focusTextEndAppendText(input: HTMLTextAreaElement, text = '') {
-  const current = input.value + text;
-  input.focus();
-  input.value = '';
-  input.value = current;
-  autosize.update(input);
-}
 
 export function handleBlockEndEnter(page: Page, model: ExtendedModel) {
   const parent = page.getParent(model);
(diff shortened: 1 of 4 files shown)


tests/paragraph.spec.ts
```diff
@@ -90,6 +91,33 @@ test('backspace and arrow on title', async ({ page }) => {
   await assertTitle(page, 'hll');
 });
 
+test('backspace on line start of the first block', async ({ page }) => {
```
doodlewind, line 94: This is a test case that is expected for this edge case. You can use the [playwright VSCode plugin](https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright) to debug it visually. Quality-related takeaways here:

* To ensure the state is stable, we require that the editor state should be the same after one undo and one redo, just like the [principle](https://www.figma.com/blog/how-figmas-multiplayer-technology-works/#:~:text=if%20you%20undo%20a%20lot%2C%20copy%20something%2C%20and%20redo%20back%20to%20the%20present%20(a%20common%20operation)%2C%20the%20document%20should%20not%20change.) shared by Figma.
* If no paragraph remains on the page, there will be a cursor blind zone. So in the asserter, we expect `['\n']` (one empty paragraph or list block) instead of `[]` (no content block) here.

packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
```diff
@@ -276,8 +286,15 @@ export function handleLineStartBackspace(page: Page, model: ExtendedModel) {
             page.deleteBlock(model);
           }
         });
+      } else {
+        const richText = getRichTextByModel(model);
+        if (richText) {
+          page.captureSync();
+          page.deleteBlock(model);
```
doodlewind, line 293: Deleting this starting block may bring a cursor blind zone. Instead, `model.text.delete(0, model.text.length)` can avoid this.

Also, IMO we need a `matchFlavours(model, ['affine:list', 'affine:paragraph'])` guard _here,_ so that other block flavours (types) won't be affected.

packages/blocks/src/__internal__/utils/query.ts
```diff
@@ -394,3 +394,8 @@ export function getAllBlocks() {
     );
   });
 }
+
+export const getPageTitle = () =>
```
doodlewind, line 398: We use the `export function` convention :)

packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
```diff
@@ -20,8 +21,17 @@ import {
   focusPreviousBlock,
   focusBlockByModel,
   supportsChildren,
+  getPageTitle,
 } from '../utils/index.js';
 
+function focusTextEndAppendText(input: HTMLTextAreaElement, text = '') {
```
doodlewind, line 27: In this case, you are directly mutating the DOM. But when it comes to non-native undo/redo, the single source of truth is always the model state (see https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/page-block/default/default-page-block.ts#L293-L298). So instead, calling `page.updateBlock(model, { title: newTitle })` and waiting for the state update should be the best practice.

doodlewind (commented):
I think you're pretty close, go for it!

himself65 (approved):
Make sense to me, nice work

himself65:
I think the redo undo now works well. What about your thought? /cc @doodlewind

doodlewind (approved):

PR #581, opened 2023-01-07
chore: hide edgeless debug menu entry

doodlewind:


Himself65 (approved):
Breaking changes on e2e test

packages/playground/src/components/debug-menu.ts
```diff
@@ -410,7 +409,9 @@ export class DebugMenu extends LitElement {
 
         <div
           class="edgeless-toolbar"
-          style=${'display:' + (this.mode === 'edgeless' ? 'flex' : 'none')}
+          style=${
+            'display:none' /*'display:' + (this.mode === 'edgeless' ? 'none' : 'none')*/
```
himself65, line 413: remove?

doodlewind: Well I'd rather keep it here till the new self-contained toolbar is landed.

/cc @thorseraq

thorseraq: Thanks, current debug menu is a good reference

Issue #583, opened 2023-01-07
Incorrect multi-line title size after undo/redo


zoulamZero:
Recorded on the latest [playground](https://block-suite.pages.dev/?init)

https://user-images.githubusercontent.com/83949429/211130531-8927b40a-c7b1-40fa-8ea9-79340811a29b.mp4

append `autosize.update(this._title);` at [this line](https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/page-block/default/default-page-block.ts#L296) can fix it

doodlewind:
Good catch! Having a standalone PR fixing this would be great!

PR #584, opened 2023-01-07
fix: multi-line title size after undo/redo

zoulamZero:
fix: #583


Himself65 (approved):
LGTM, but is there possible to add a test?

doodlewind:
> LGTM, but is there possible to add a test?
>
> For example, in the empty title, the height should be `x`. and you type something to `2x` height. and undo, the expected height should be `x`

I think we can be a bit permissive for styling issues :D

doodlewind (approved):

Issue #582, opened 2023-01-07
DragHandle may be positioned to parent during a same-level drag


Flrande:
When moving between siblings, the button will move to the upper node briefly.

blocksuite:

![20230107114117_rec_](https://user-images.githubusercontent.com/50035259/211129683-2a0d5dce-a80e-4c40-a643-b428caa3e6d5.gif)

feishu:

![20230107114156_rec_](https://user-images.githubusercontent.com/50035259/211129708-f654dc69-b985-4ce3-9a81-3dbd68d62033.gif)

himself65:
Good catch, because the block rect box is smaller than you thought. So there's a small gap between each block

himself65:
hot patch

```
.affine-block-children-container {
    display: flex;
    flex-direction: column;
}
```

Flrande:
We can redesign the generating function of the button position. For example, we can determine which block-level node is in the horizontal direction by calculating getBoundingClientRect().bottom in mousemove. (Like this [function](https://github.com/toeverything/blocksuite/blob/cb27920033b719b5943fa39d4c2a786926911bc4/packages/blocks/src/__internal__/utils/selection.ts#L836)) This can also improve the user experience because users do not have to put the mouse into the node to change the button position.

blocksuite:

![20230107125133_rec_](https://user-images.githubusercontent.com/50035259/211131672-5d91a097-7048-431f-b9f4-adbded103451.gif)

feishu:

![20230107125205_rec_](https://user-images.githubusercontent.com/50035259/211131692-246622c7-b7a8-49a6-80b0-5a621d4f1a42.gif)

himself65:
> We can redesign the generating function of the button position. For example, we can determine which block-level node is in the horizontal direction by calculating getBoundingClientRect().bottom in mousemove. (Like this [function](https://github.com/toeverything/blocksuite/blob/cb27920033b719b5943fa39d4c2a786926911bc4/packages/blocks/src/__internal__/utils/selection.ts#L836)) This can also improve the user experience because users do not have to put the mouse into the node to change the button position.
> 
> blocksuite:
> 
[...]

PR #586, opened 2023-01-07
fix: use flex in children container

Himself65:
Fixes: https://github.com/toeverything/blocksuite/issues/582


Flrande (approved):

doodlewind (approved):
We may need to adjust the size of frame block

packages/blocks/src/page-block/default/selection-manager.ts
```diff
@@ -34,10 +34,10 @@ import { showDragHandle } from '../../components/drag-handle.js';
 
 function intersects(rect: DOMRect, selectionRect: DOMRect, offset: IPoint) {
```
himself65, line 35: list item now will right next to each other

PR #588, opened 2023-01-07
fix: native selection

thorseraq:
Before:
#527

After:

https://user-images.githubusercontent.com/20554850/211153684-7ee06cea-1a9c-44f2-9ab7-072f7e0ee2dc.mov

### DIFF
--- a/packages/blocks/src/__internal__/utils/query.ts
+++ b/packages/blocks/src/__internal__/utils/query.ts
@@ -1,9 +1,10 @@
 import type { BaseBlockModel } from '@blocksuite/store';
 import type { DefaultPageBlockComponent, SelectedBlock } from '../../index.js';
+import { ShapeBlockTag } from '../../index.js';
 import type { RichText } from '../rich-text/rich-text.js';
 import { BLOCK_ID_ATTR as ATTR } from './consts.js';
 import { assertExists, matchFlavours } from './std.js';
-import { ShapeBlockTag } from '../../index.js';
+import type { IPoint } from './gesture.js';
 
 type ElementTagName = keyof HTMLElementTagNameMap;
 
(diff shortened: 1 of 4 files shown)


himself65:
What is the bug you fixed? Could you please give a short description for the future people who use git blame

packages/blocks/src/__internal__/utils/selection.ts
```diff
@@ -414,7 +425,16 @@ export function handleNativeRangeDragMove(
   startRange: Range | null,
   e: SelectionEvent
 ) {
-  const isForward = e.x > e.start.x || e.y > e.start.y;
+  const startBlock = getBlockByPoint(e.start);
+  assertExists(startBlock);
+  const startRect = startBlock.getBoundingClientRect();
+
+  const smallDistance = 6;
```
himself65: why 6?

thorseraq: No particular criterion, any suggestion?

himself65: nope

himself65: I think, at least, make all `6` into a constant to declare that they are pointing to the same 6 under same meaning

doodlewind (approved):

packages/blocks/src/__internal__/utils/selection.ts
thorseraq: Thanks! Done, Let's make it 2px, which is also used in isFarEnough()

himself65 (approved):

packages/blocks/src/__internal__/utils/query.ts
```diff
@@ -28,6 +29,12 @@ export function getBlockById<T extends ElementTagName>(
   return ele.querySelector<T>(`[${ATTR}="${id}"]` as T);
 }
 
+export function getBlockByPoint(point: IPoint): Element | null | undefined {
+  return document
+    .elementFromPoint(point.x, point.y)
+    ?.closest('[data-block-id]');
```
lawvs, line 35: you should use `BLOCK_ID_ATTR`

thorseraq: ok

PR #589, opened 2023-01-07
build: restrict node version and add packageManager to force agent

fi3ework:
Set node version's lower limit to `16.9.0` as `corepack` is added since then.


doodlewind (approved):

doodlewind:
Nice improvement, thanks!

PR #592, opened 2023-01-08
feat: improve drag handle

### LINKED ISSUE #582: DragHandle may be positioned to parent during a same-level drag

himself65:
Fixes: https://github.com/toeverything/blocksuite/issues/582#issuecomment-1374378880


himself65:
The drag handle won't appear on the first render; I will fix it in the next pr which uses `singleton mode`

PR #594, opened 2023-01-08
refactor: use random room by default

doodlewind:
We should make collaborative opt-in in the playground, so as to avoid the `?init` conflict. So in this case, only the playground URL with the same `room` param will be explicitly assigned to the same room.


himself65:
Please also remove the code in vite config

PR #597, opened 2023-01-08
fix: `captureSync` on creating code block & filter hotkey listener 

thorseraq:


packages/blocks/src/__internal__/rich-text/shortcuts.ts
```diff
@@ -286,6 +286,7 @@ export class Shortcuts {
         };
         page.deleteBlock(model);
         page.addBlock(blockProps, parent, index);
+        page.captureSync();
```
lawvs, line 289: How it works

thorseraq: It's not related to hotkey manager, Just unify behavior of after creating a block, use page.captureSync() :)

lawvs: But in most cases, we call `page.captureSync()` before block operation

thorseraq: `page.captureSync()` is now before block operation

thorseraq:
This PR covers #619

packages/blocks/src/__internal__/rich-text/shortcuts.ts
```diff
@@ -278,6 +278,7 @@ export class Shortcuts {
           return PREVENT_DEFAULT;
         }
         const page = getDefaultPageBlock(model).page;
```
doodlewind, line 280: This looks a bit weird here:

1. I think `getDefaultPageBlock` should return `page` rather than `{ page: Page }`.
2. Why not use `model.page`?

thorseraq: 1. `getDefaultPageBlock()` returns the `DefaultPageBlockComponent` rather than `Page`, which is a property of `DefaultPageBlockComponent`.
2. YES!!!

PR #607, opened 2023-01-08
perf: use binary search on blocks rect hit test

### LINKED ISSUE #599: performance improve on blocks rect detect
We could use binary search on every code that uses for-loop in blocks, which will improve the algorithm complexity from `O(n)` to `O(log(n))`.

This could make 1000 blocks (heavy) example much faster than ever.

Example:

https://github.com/toeverything/blocksuite/blob/c9785dd905a05154cfb4c100d80d3e5ae18629ab/packages/blocks/src/page-block/default/utils.ts#L44-L85

fundon:
Fixes: https://github.com/toeverything/blocksuite/issues/599

1. use `binary search` to find the `block index` where the current mouse is (that is the position of `affine-drag-handle`) and cache it. #599 
2. use `Seek` to move in a way to further narrow down the search with the cursor.

``` js
  const size = options?.size || 5;
  const start = Math.max(cursor - size, 0);
  const end = Math.min(cursor + size, blocks.length - 1);
 ```

3. In terms of user experience, also need to consider page scrolling.

### CODE REVIEW

**Review by doodlewind — COMMENTED** [review]
Very impressive improvement! Last challenge here: how can we query the nested child? At following cursor position, seems that the drag handle doesn't found the leaf block. I think this is why the test case failed. Did you try searching reversely from end to start?

Screenshot 2023-01-09 at 07 21 23

**Review by himself65 — DISMISSED** [review]
I'm now thinking that the best solution is to only detect the blocks that appear on the screen, which means the overflowed blocks will be ingored.

So the algorithm would be:

1. Find the last block on the bottom of the screen (use binary search `O(log(n))`)
2. Use linear search from the one from behind to the front. (`O(1)`), return null if the current one is overflowing the screen. return model state if find the correct one

So the overall algorithm complexity is `O(log(n)),` but if we could maintain the bottom block every time the user wheel and resize the document. the complexity would be `O(1)`

**Comment by lawvs** [comment]
IMO we should extract the binarySearch function into a shared module, it can avoid duplicating the same code in multiple places.

And we can more easily test the function in isolation and ensure that it is working correctly.

This can help to improve the reliability and maintainability of the code.

We also used binary search here

https://github.com/toeverything/blocksuite/blob/5143005c9e6e95ead4be57b1c360c2abb98f1182/packages/blocks/src/__internal__/utils/selection.ts#L845-L863


Here is a simple template for binarySearch
```ts
/**
 * BinarySearch
 * @param {number[]} array
 * @param {(midVal: number, midIdx: number) => boolean} isLow
 * @returns {number} idx
 * @see https://www.youtube.com/watch?v=JuDAqNyTG4g
 */
const binarySearch = <T = number >(array: T[], isLow = () => true) => {
  let low = -1;
  let high = array.length;
  while (low + 1 !== high) {
    const mid = Math.floor((low + high) / 2);
    if (isLow(array[mid], mid)) {
      low = mid;
    } else {
      high = mid;
    }
  }
  // return low
  return high;
};
```

fundon:
https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/page-block/default/selection-manager.ts#L225

The `unshift` method should be used to keep the blocks in order.

fundon:
@Himself65 I tested it in Firefox, found this issue https://stackoverflow.com/questions/13110349/pagex-and-pagey-are-always-set-to-0-in-firefox-during-the-ondrag-event.

himself65 (approved):

doodlewind (approved):
LGTM, congratulations  🎉

himself65:
Thanks for the significant improvement on the performance

PR #611, opened 2023-01-09
feat: export in safe file name

AkaraChen:
When export file, the browser will replace unsafe characters in the file name with `_`, which is not expected.

For example:

`Are you sure?.md` => `Are you sure_.md`
`Are you sure?????.md` => `Are you sure______.md`

This pr try to make export file name look more comfortable:

`Are you sure?.md` => `Are you sure.md`
`Are you sure?????.md` => `Are you sure.md`

### DIFF
--- a/packages/editor/package.json
+++ b/packages/editor/package.json
@@ -14,6 +14,7 @@
   "dependencies": {
     "@blocksuite/blocks": "workspace:*",
     "@blocksuite/store": "workspace:*",
+    "filenamify": "^5.1.1",
     "lit": "^2.5.0",
     "marked": "^4.2.5",
     "turndown": "^7.1.1"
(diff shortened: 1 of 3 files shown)


doodlewind (commented):
It seems that the `filenamify` package doesn't work with our ESM setup (we are supporting ESM in both Next and Vite environment) and caused the build error. But since its source code (and also the source of its deps) are pretty simple, how about inlining them directly into the FileManager directory?

AkaraChen:
> It seems that the `filenamify` package doesn't work with our ESM setup (we are supporting ESM in both Next and Vite environment) and caused the build error. But since its source code (and also the source of its deps) are pretty simple, how about inlining them directly into the FileManager directory?

Okay, I'll do it later.

doodlewind (approved):

doodlewind:
Thanks for your contribution!

PR #615, opened 2023-01-09
chore: update code block unit test

thorseraq:
Update: Press ArrowUp after code block now should select code block, rather than jumping into it

### DIFF
--- a/tests/code.spec.ts
+++ b/tests/code.spec.ts
@@ -258,7 +258,7 @@ test('press backspace after code block can jump into start of code block', async
   expect(index).toBe(0);
 });
 
-test.only('press ArrowUp after code block can select code block', async ({
+test('press ArrowUp after code block can select code block', async ({
   page,
 }) => {
   await enterPlaygroundRoom(page);
@@ -269,10 +269,10 @@ test.only('press ArrowUp after code block can select code block', async ({
   await page.keyboard.press('Enter');
   await page.keyboard.press('ArrowUp');
 
-  const IfSelectRectExist = await page.evaluate(() => {
+  const selectedRectsExist = await page.evaluate(() => {
     return !!document.querySelector('.affine-page-selected-rects-container');
   });
-  expect(IfSelectRectExist).toEqual(true);
+  expect(selectedRectsExist).toEqual(true);
 });
 
 test('undo and redo works in code block', async ({ page }) => {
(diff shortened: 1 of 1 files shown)


tests/code.spec.ts
```diff
@@ -258,7 +258,7 @@ test('press backspace after code block can jump into start of code block', async
   expect(index).toBe(0);
 });
 
-test('press ArrowUp after code block can jump into start of code block', async ({
+test.only('press ArrowUp after code block can select code block', async ({
```
doodlewind, line 261: Please remove `test.only`, and `IfSelectRectExist` -> `selectedRectsExist`, it's always a good habit to pay attention to word properties and plural cases 👻

thorseraq: Thanks for reminding!!!

doodlewind (approved):

tests/code.spec.ts
```diff
@@ -268,8 +268,11 @@ test('press ArrowUp after code block can jump into start of code block', async (
   await page.keyboard.press('Enter');
   await page.keyboard.press('Enter');
   await page.keyboard.press('ArrowUp');
-  const index = await getQuillSelectionIndex(page);
-  expect(index).toBe(0);
+
+  const selectedRectsExist = await page.evaluate(() => {
+    return !!document.querySelector('.affine-page-selected-rects-container');
+  });
+  expect(selectedRectsExist).toEqual(true);
```
lawvs, line 275: you can use playwright locator API next time

```ts
const selectedRects = page.locator('.affine-page-selected-rects-container');
await expect(selectedRects).toBeVisible();
await expect(selectedRects).toHaveCount(1);
```

ref https://playwright.dev/docs/api/class-locatorassertions

doodlewind: @Himself65 your PR time 😈

lawvs: The `toBeVisible` also will check the element attached and visible. The zero-height container broke the test. I think expecting count is enough in this scene.



> Element is considered visible when it has non-empty bounding box and does not have visibility:hidden computed style. Note that elements of **zero size** or with display:none are not considered visible.


Screenshot 2023-01-09 at 3 30 17 PM

thorseraq: 🐮🍺

PR #624, opened 2023-01-09
feat: adding toolbar in edgeless page block

thorseraq:
Now toolbar is hidden by default, adding `toolbar` in search param will activate edgeless toolbar, for example http://localhost:5173/?init&toolbar.


packages/blocks/src/page-block/edgeless/components.ts
```diff
@@ -187,6 +190,117 @@ export function EdgelessBlockChildrenContainer(
   `;
 }
 
+@customElement('edgeless-toolbar')
+export class EdgelessToolBar extends LitElement {
+  static styles = css`
```
lawvs, line 195: Extract styles to a separate file

thorseraq:
move to #794

PR #628, opened 2023-01-09
feat: type-safe `addBlockByFlavour` API

Himself65:


packages/global/src/index.d.ts
```diff
@@ -31,3 +45,61 @@ declare namespace BlockSuiteInternal {
     'affine:surface': SurfaceBlockModel;
   };
 }
+
+type EmbedType = 'image' | 'video' | 'audio' | 'file';
+type ListType = 'bulleted' | 'numbered' | 'todo';
+type ParagraphType = 'text' | 'quote' | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+
+namespace BlockSuiteModelProps {
```
doodlewind, line 53: Do we have a chance to place this definition in the module of different blocks? Otherwise, we may end up defining block-related entities in multiple places.

himself65: No, but I think it should be a best practice for the people who want to add a new block.

PR #630, opened 2023-01-09
feat: support bracket autocomplete for native range selected text

lawvs:
fix https://github.com/toeverything/AFFiNE/issues/632

~~but incorrect in CJK input method.~~

~~It may also fail on other keyboard layouts.~~

~~Block by https://github.com/quilljs/quill/issues/1975~~

It does not work with multiple block selection now

### DIFF
--- a/packages/blocks/src/__internal__/rich-text/keyboard.ts
+++ b/packages/blocks/src/__internal__/rich-text/keyboard.ts
@@ -11,6 +11,7 @@ import {
   isMultiBlockRange,
   noop,
 } from '../utils/index.js';
+import { createBracketAutoCompleteBindings } from './bracket-complete.js';
 import { markdownConvert } from './markdown-convert.js';
 import {
   handleBlockEndEnter,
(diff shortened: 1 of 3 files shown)


packages/blocks/src/__internal__/rich-text/keyboard.ts
```diff
@@ -44,6 +45,7 @@ type KeyboardBinding = {
    * See https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values
    */
   key: number | string | string[];
+  collapsed?: boolean;
```
doodlewind, line 48: Why using `satisfies` could match the type here without adding this field?

doodlewind (approved):

packages/blocks/src/__internal__/rich-text/keyboard.ts
lawvs: Yes, It's unexpected.

PR #635, opened 2023-01-10
fix: return the correct delta result in `sliceToDelta`

### LINKED ISSUE #601: Incorrect copied result with more than one delta operation
## Description
It's a bug of the [`sliceToDelta`](https://github.com/toeverything/blocksuite/blob/3543ce264be856958312a38b0165e158a326558a/packages/store/src/text-adapter.ts#LL233C8-L233C8) method. When more than one delta operation was copied, the `contentText` variable of the last operation was [updated](https://github.com/toeverything/blocksuite/blob/3543ce264be856958312a38b0165e158a326558a/packages/store/src/text-adapter.ts#L250) without being [pushed](https://github.com/toeverything/blocksuite/blob/3543ce264be856958312a38b0165e158a326558a/packages/store/src/text-adapter.ts#L261) into `result`.

## Record
eUQjoMvZF5

donaldxdonald:
fixed #601

### DIFF
--- a/packages/store/src/text-adapter.ts
+++ b/packages/store/src/text-adapter.ts
@@ -231,28 +231,31 @@ export class Text {
   }
 
   sliceToDelta(begin: number, end?: number) {
+    const result: DeltaOperation[] = [];
     if (end && begin >= end) {
-      return [];
+      return result;
     }
 
     const delta = this.toDelta();
     if (begin < 1 && !end) {
       return delta;
     }
-    const result = [];
+
     if (delta && delta instanceof Array) {
       let charNum = 0;
       for (let i = 0; i < delta.length; i++) {
-        const content = delta[i];
-        let contentText = content.insert || '';
+        const content: DeltaOperation = delta[i];
+        let contentText: string = content.insert || '';
         const contentLen = contentText.length;
-        if (end && charNum + contentLen > end) {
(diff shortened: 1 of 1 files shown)


packages/store/src/text-adapter.ts
```diff
@@ -231,38 +231,43 @@ export class Text {
...
       return delta;
     }
-    const result = [];
+
     if (delta && delta instanceof Array) {
       let charNum = 0;
       for (let i = 0; i < delta.length; i++) {
-        const content = delta[i];
-        let contentText = content.insert || '';
+        const content: DeltaOperation = delta[i];
```
himself65, line 247:
```suggestion
        const content = delta[i];
```

doodlewind (commented):
Nice catch! Looks like this case can be easily reproduced as a test case, would you like to give it a try? Otherwise, there may be regressions in our future refactoring.

packages/store/src/text-adapter.ts
donaldxdonald: I think using `any` everywhere is not a good idea. 🤔

himself65: sorry if misleading, I mean make `toDelta` to

```ts
toDelta(): DeltaOperation[] {
```

PR #638, opened 2023-01-10
fix: dispose quill keyboard handler before block selection

### LINKED ISSUE #633: Unable to delete in block-level selection after selecting code block
The code block can't be deleted by backspace if it's under block-level selection:

image

thorseraq:
Fixes #633

### DIFF
--- a/packages/blocks/src/page-block/default/selection-manager.ts
+++ b/packages/blocks/src/page-block/default/selection-manager.ts
@@ -257,7 +257,7 @@ export class DefaultSelectionManager {
     this.state.refreshRichTextBoundsCache(this._mouseRoot);
     resetNativeSelection(null);
     // deactivate quill keyboard event handler
-    (document.activeElement as HTMLTextAreaElement).blur();
+    (document.activeElement as HTMLDivElement).blur();
   }
 
   private _onBlockSelectionDragMove(e: SelectionEvent) {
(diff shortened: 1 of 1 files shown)


packages/blocks/src/page-block/default/selection-manager.ts
```diff
@@ -256,6 +256,8 @@ export class DefaultSelectionManager {
     this.state.resetStartRange(e);
     this.state.refreshRichTextBoundsCache(this._mouseRoot);
     resetNativeSelection(null);
+    // deactivate quill keyboard event handler
+    (document.activeElement as HTMLTextAreaElement).blur();
```
doodlewind, line 260: I think the Quill instance is an `HTMLDivElement`.

thorseraq: How about this

```typescript
if (activeElement && 'blur' in activeElement && typeof activeElement['blur'] === 'function') {
  // deactivate quill keyboard event handler
    activeElement.blur();
  }
```

doodlewind: This looks too verbose. `as HTMLDivElement` may be semantically better, otherwise, it may indicate that the instance here is the page title.

thorseraq: OK, Get that!

PR #639, opened 2023-01-10
refactor: abstract block-specific methods into `BlockService`

himself65:


himself65:
Cannot be merged because too many conflicts

PR #640, opened 2023-01-10
fix: optimize `isForward` detection in `SelectionManager`

thorseraq:
1. Related to #588, in #588, multiple lines of one content block was not considered, which will cause back select failure
2. Fix: When BlockSuite is embedded in other container, can't get correct start container by `SelectionEvent.start.x`, `SelectionEvent.start.y`,

### DIFF
--- a/packages/blocks/src/__internal__/utils/query.ts
+++ b/packages/blocks/src/__internal__/utils/query.ts
@@ -29,9 +29,6 @@ export function getBlockById<T extends ElementTagName>(
   return container.querySelector<T>(`[${ATTR}="${id}"]` as T);
 }
 
-/*
-Be cautious!! If BlockSuite is embedded in other container, MouseEvent.clientX, MouseEvent.clientY will change
- */
 export function getBlockByPoint(point: IPoint): Element | null | undefined {
   return document.elementFromPoint(point.x, point.y)?.closest(`[${ATTR}]`);
 }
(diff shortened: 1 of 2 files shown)


packages/blocks/src/__internal__/utils/query.ts
```diff
@@ -29,6 +29,9 @@ export function getBlockById<T extends ElementTagName>(
   return container.querySelector<T>(`[${ATTR}="${id}"]` as T);
 }
 
+/*
+Be cautious!! If BlockSuite is embedded in other container, MouseEvent.clientX, MouseEvent.clientY will change
```
lawvs, line 33: What is the meaning of `BlockSuite` here? Can you give an example? Because this function and MouseEvent are not related.

lawvs (changes requested):

lawvs:
What is the bug you fixed? Could you please give a short video about bug reports?

thorseraq:
Having a P0 in hand, please wait

packages/blocks/src/__internal__/utils/query.ts
himself65: I think e.raw.clientX Y is always what it is on the screen.

doodlewind: I'd rather not add redundant comment here :)

doodlewind: But this value can't be compared with the staring position since they're not inside the same coord system.

PR #641, opened 2023-01-10
fix: dispose quill keyboard handler before block selection

thorseraq:


packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
```diff
@@ -224,18 +224,15 @@ export function handleLineStartBackspace(page: Page, model: ExtendedModel) {
...
     const index = parent.children.indexOf(model);
+    const blockProps = {
+      flavour: 'affine:paragraph',
+      type: 'text',
+      text: model.text?.clone(),
+      children: model.children,
+    };
     page.captureSync();
     page.deleteBlock(model);
-    const id = page.addBlockByFlavour(
```
himself65, line 229: why dont use this?

thorseraq: It's will cause unit test fail

himself65: Seems like a bug

himself65: Could you please create a issue?

doodlewind (approved):

PR #643, opened 2023-01-10
fix: use manual shallow copy on `DOMRect`

### LINKED ISSUE #632: Image selection state shelters the blocks above
When an image is inserted into the page and is selected, the blocks above (red frame in the screenshot) are not clickable.

image

This is introduced in #606

JimmFly:
Fixes #632

### DIFF
--- a/packages/blocks/src/page-block/default/utils.ts
+++ b/packages/blocks/src/page-block/default/utils.ts
@@ -48,6 +48,7 @@ function hasOptionBar(block: BaseBlockModel) {
 }
 
 // Workaround native DOMRect clone issue in #632
+// See https://stackoverflow.com/questions/42713229/getboundingclientrect-object-properties-cannot-be-copied
 function copyRect(rect: DOMRect): DOMRect {
   const { top, right, bottom, left, width, height, x, y } = rect;
   return { top, right, bottom, left, width, height, x, y } as DOMRect;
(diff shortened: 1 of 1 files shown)


packages/blocks/src/page-block/default/utils.ts
```diff
@@ -68,12 +68,15 @@ export function getBlockEditingStateByPosition(
...
+    const copyRect = (rect: DOMRect) => {
+      const { top, right, bottom, left, width, height, x, y } = rect;
+      return { top, right, bottom, left, width, height, x, y };
+    };
     if (hasOptionBar(block)) {
       const currentHoverDom =
         block.type === 'image' ? hoverDom?.querySelector('img') : hoverDom;
       blockRect = currentHoverDom?.getBoundingClientRect() as DOMRect;
-      detectRect = { ...blockRect } satisfies DOMRect;
+      detectRect = copyRect(blockRect) as DOMRect;
```
lawvs, line 79: why does it work?

lawvs (changes requested):

Himself65 (approved):

packages/blocks/src/page-block/default/utils.ts
Himself65: https://stackoverflow.com/questions/42713229/getboundingclientrect-object-properties-cannot-be-copied

himself65: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties#querying_object_properties

PR #652, opened 2023-01-10
test: add test case for copying multiple delta ops

donaldxdonald:
related #635


doodlewind (approved):
This looks solid, thanks!

PR #659, opened 2023-01-11
fix: parse inline html into single paragraph

### LINKED ISSUE #650: Content pasted from the GitHub README is not correctly recognized
Reproduction:

Paste the following content in BlockSuite:

> 💠 BlockSuite is an open-source collaborative editor project behind [AFFiNE](https://github.com/toeverything/AFFiNE), an open-source knowledge base that attempts to combine the best of Notion and Miro altogether. In this monorepo, we maintain multiple abstraction layers of the editor and some core modules that can be reused externally.

Recognized result:

image

This also happens for the HTML pasted from Wikipedia.

Flrande:
Fixes #650

### DIFF
--- a/packages/editor/src/managers/clipboard/content-parser/parse-html.ts
+++ b/packages/editor/src/managers/clipboard/content-parser/parse-html.ts
@@ -170,6 +170,26 @@ export class HtmlParser {
     if (result && result.length > 0) {
       return result;
     }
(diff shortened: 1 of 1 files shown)


doodlewind:
@Flrande, hope to see the test case to guard this high-frequency code path, thanks!

PR #661, opened 2023-01-11
fix(blocks): set supressWarning to true of quill-cursors

fi3ework:
## why
clean blocking for https://github.com/toeverything/blocksuite/pull/621, stop the re-registry warning when performing HMR. And, all the other quill registry has set `supressWarning` to `true` so I guess this is not an exception. I tested the official example of quill-curosrs https://github.com/reedsy/quill-cursors/tree/main/example and tweaked it with many re-register, seems like everything is fine.


doodlewind (approved):
Nice improvement, thanks!

PR #666, opened 2023-01-11
fix: query code block container when loading

Himself65:


packages/blocks/src/__internal__/utils/components.ts
```diff
@@ -74,7 +74,7 @@ function BlockElementWithService(
         state.then(() => {
           onLoaded();
         });
-        return html`<loader-element />`;
+        return html` <loader-element .hostModel=${model}> </loader-element> `;
```
lawvs, line 77: WebStorm will automatically add spaces in front of string templates, do you have any ideas on this?

himself65: Nope

PR #677, opened 2023-01-12
fix: cant reach the last block in nesting

fundon:
## Before
If we want to move A after B, but not a child of B. The current move can only be a child of B, if B is the last block.

Screenshot 2023-01-12 at 4 30 29 PM

## After

May not be very good in user experience

https://user-images.githubusercontent.com/27926/212127387-83f54f4e-44f7-4b52-8aa8-9f4f68f2b6a2.mov

### DIFF
--- a/packages/blocks/src/page-block/default/utils.ts
+++ b/packages/blocks/src/page-block/default/utils.ts
@@ -109,6 +109,7 @@ export function getBlockEditingStateByCursor(
   options?: {
     size?: number;
     skipX?: boolean;
+    dragging?: boolean;
   }
 ): EditingState | null {
   const size = options?.size || 5;
@@ -126,8 +127,10 @@ function binarySearchBlockEditingState(
   end: number,
   options?: {
     skipX?: boolean;
+    dragging?: boolean;
   }
 ): EditingState | null {
+  const dragging = Boolean(options?.dragging);
   while (start <= end) {
     const mid = start + Math.floor((end - start) / 2);
     const { block, blockRect, detectRect, hoverDom } = getBlockAndRect(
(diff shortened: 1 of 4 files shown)


packages/blocks/src/page-block/default/utils.ts
```diff
@@ -178,8 +181,30 @@ function binarySearchBlockEditingState(
...
+
+                return {
+                  index: mid,
+                  position: result.blockRect,
+                  model: result.block,
+                };
+              }
+            }
+          } else {
+            if (x < detectRect.left || x > detectRect.left + detectRect.width) {
```
himself65, line 205: use `isPointIn`?

packages/blocks/src/page-block/default/utils.ts
```diff
@@ -178,8 +181,30 @@ function binarySearchBlockEditingState(
         assertExists(blockRect);
 
         if (!options?.skipX) {
-          if (x < detectRect.left || x > detectRect.left + detectRect.width) {
-            return null;
+          if (dragging) {
+            if (block.depth) {
+              let depth = Math.floor((blockRect.left - x) / 26);
```
himself65, line 186: what is 26?

fundon: padding-left

packages/blocks/src/page-block/default/utils.ts
fundon: Ok, add it back

packages/blocks/src/page-block/default/utils.ts
himself65: I think we should put everything related into a constant

PR #680, opened 2023-01-13
fix: line number for wrapped lines

thorseraq:
Before change, the code block line number is calculated by number of `\n`, when there are long lines, and when they are wrapped, the line number will be more than number of `\n`, it causes line number less than actual lines, as shown below.
image

After, line numbers only cares how much start lines there are, exceeding part are filled up by calculating

https://user-images.githubusercontent.com/20554850/212226891-11c8075d-5a54-4bb5-be89-3197a6979493.mov

### DIFF
--- a/packages/blocks/src/code-block/components/syntax-code-block.ts
+++ b/packages/blocks/src/code-block/components/syntax-code-block.ts
@@ -16,9 +16,9 @@ function addLineNumber(
 ) {
   const node = document.createElement('div');
   if (i) {
-    node.innerHTML = `${i}`;
+    node.textContent = `${i}`;
   } else {
-    node.innerHTML = '';
+    node.textContent = '';
     node.style.height = lineHeight;
   }
   container.appendChild(node);
@@ -70,21 +70,9 @@ class SyntaxCodeBlock extends CodeBlock {
     }
   }
 
-  private _getCtx() {
-    const fontSize = window.getComputedStyle(this.domNode).fontSize;
-    const fontFamily = window
-      .getComputedStyle(this.domNode)
-      .fontFamily.split(',')[0];
-    const canvas = document.createElement('canvas');
-    const ctx = canvas.getContext('2d') as CanvasRenderingContext2D;
-    ctx.font = `${fontSize} ${fontFamily}`;
-    return ctx;
-  }
-
   updateLineNumber(codeBlockElement: HTMLElement, forceRefresh = false) {
     const text = this.domNode.textContent;
     assertExists(text);
-    const ctx = this._getCtx();
     const hasWrap = this.domNode.classList.contains('wrap');
 
     if (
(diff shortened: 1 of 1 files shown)


packages/blocks/src/code-block/components/syntax-code-block.ts
```diff
@@ -80,13 +137,15 @@ SyntaxCodeBlock.className = 'ql-syntax';
...
   private _language = 'javascript';
+
   static register() {
     Quill.register(CodeToken, true);
     Quill.register(SyntaxCodeBlock, true);
   }
 
   constructor(quill: QuillType, options: SyntaxCodeBlockOptions) {
     super(quill, options);
+    // console.log('(quill: QuillType, options: SyntaxCodeBlockOptions)');
```
himself65: remove?

thorseraq: done

packages/blocks/src/code-block/components/syntax-code-block.ts
```diff
@@ -9,10 +9,25 @@ const Module = Quill.import('core/module');
...
 const CodeToken = Quill.import('modules/syntax');
 
+function addLineNumber(
+  container: HTMLElement,
+  lineHeight: string,
+  i: number | null
+) {
+  const node = document.createElement('div');
+  if (i) {
+    node.innerHTML = `${i}`;
```
lawvs, line 19: For safety reasons, use `textContent` please.

doodlewind:
Minor issue found: if one line wraps into more than 2 lines, the count would be incorrect:

image

Context:

```
111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111
22222
```

packages/blocks/src/code-block/components/syntax-code-block.ts
thorseraq: Thanks, I will change this

thorseraq:
> Minor issue found: if one line wraps into more than 2 lines, the count would be incorrect:
> 
> image
> 
> Content:
> 
> ```
> 111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111
> 22222
> ```

It seems the that line length calculated by canvas is not stable across devices, which is used to calculate line numbers of a very long line, I should change to a stable way

thorseraq:
Now uses a fake DOM simulating the code block container to calculate text width, how about try again with

testtesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttesttest
test

doodlewind (approved):
LGTM

PR #681, opened 2023-01-13
fix: collect wrong BaseBlockModel[] at _block()

### LINKED ISSUE #669: First block not reachable in drag handle
Refs: https://github.com/toeverything/AFFiNE/pull/714

thorseraq:
close #669


Himself65 (commented):
I think unshift is correct because we want to collect the blocks in the correct direction

thorseraq:
It's now changed to dfs to fix unexpectedly remove "content block" when first layer has more than one frame block, at the meantime keep the `result` index ascending with block position from upside to downside

Himself65 (approved):

PR #685, opened 2023-01-13
fix: adapting parent index of blocks to binarySearchBlockEditingState

thorseraq:
After #681 in which #669 is fixed, the parent index of each block are 1 larger, which is used in `binarySearchBlockEditingState`, causing horizontal dragging fails. This PR is to restore the behavior in #677, ie, supporting dragging in nested level


doodlewind (approved):
LGTM

PR #689, opened 2023-01-13
feat: add `readonly` flag in store

### LINKED ISSUE #686: Read-only mode can edit
In read-only mode, shortcut keys can still be used for editing, floating buttons will appear for pictures and code blocks, and the code type of code blocks can also be switched.

https://user-images.githubusercontent.com/102217452/212292534-18a11a40-0a15-4d99-99a4-1e6abff1ec96.mov

himself65:
Fixes: #686

### DIFF
--- a/packages/store/src/workspace/page.ts
+++ b/packages/store/src/workspace/page.ts
@@ -258,6 +258,9 @@ export class Page extends Space<PageData> {
     parent?: BaseBlockModel | string | null,
     parentIndex?: number
   ): string {
+    if (this.awareness.getFlag('readonly')) {
+      throw new Error('cannot create block in readonly mode');
+    }
     if (!blockProps.flavour) {
       throw new Error('Block props must contain flavour');
     }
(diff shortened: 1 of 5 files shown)


packages/store/src/workspace/page.ts
```diff
@@ -299,11 +302,18 @@ export class Page extends Space<PageData> {
   }
 
   updateBlockById(id: string, props: Partial<BlockProps>) {
+    if (this.awareness.getFlag('readonly')) {
+      console.log('1');
```
lawvs, line 306:
```suggestion
```

himself65:
where is vercel build? /cc @doodlewind

doodlewind:
Vercel error log:

```
Done in 1.1s
--
21:37:15.061 | Running "pnpm run build"
21:37:15.510 |  
21:37:15.510 | > @blocksuite/playground@ build /vercel/path0/packages/playground
21:37:15.510 | > tsc --project ./tsconfig.json && vite build
21:37:15.510 |  
21:37:19.790 | src/utils.ts(52,5): error TS2741: Property 'readonly' is missing in type '{ enable_drag_handle: true; }' but required in type 'BlockSuiteFlags'.
21:37:19.807 | ELIFECYCLE  Command failed with exit code 2.
21:37:19.823 | Error: Command "pnpm run build" exited with 1
```

packages/store/src/workspace/page.ts
```diff
@@ -258,6 +258,9 @@ export class Page extends Space<PageData> {
     parent?: BaseBlockModel | string | null,
     parentIndex?: number
   ): string {
+    if (this.awareness.getFlag('readonly')) {
```
himself65, line 261: **Caution**: when you only enable read-only in the store but not on the block page. Press 'Enter' in the paragraph could cause unexpected result

himself65: which means you need set read-only value both in page and store

PR #692, opened 2023-01-13
refactor: readonly mode for multiple spaces

Himself65:


Himself65 (commented):
I added `console.error` because modifying data should be prohibited at the UI level. The check in-store level is just a backup. DONT drop the job to the store. Otherwise, the UI would be broken because the status needs consistency.

Issue #694, opened 2023-01-13
Support `page.awareness.setRemoteFlag(clientID, field, value)`


Himself65:
Upstream: https://github.com/yjs/y-protocols/issues/25

PR #695, opened 2023-01-13
feat: support set remote flag

Himself65:
Fixes: https://github.com/toeverything/blocksuite/issues/694

Since modifying awareness source code is harder than I thought, I implemented a version that uses `local state` to update others flags

### DIFF
--- a/packages/store/src/awareness.ts
+++ b/packages/store/src/awareness.ts
@@ -115,7 +115,7 @@ export class AwarenessAdapter<
   public isReadonly(): boolean {
     const rd = this.getFlag('readonly');
     if (rd && typeof rd === 'object') {
-      return (rd as Record<string, boolean>)[this.space.prefixedId];
+      return Boolean((rd as Record<string, boolean>)[this.space.prefixedId]);
     } else {
       return false;
     }
--- a/packages/store/src/awareness.ts
+++ b/packages/store/src/awareness.ts
@@ -38,8 +38,8 @@ interface AwarenessState<
   cursor?: SelectionRange;
   user?: UserInfo;
   flags: Flags;
-  changeRequest?: Request<Flags>[];
-  changeResponse?: Response[];
+  request?: Request<Flags>[];
+  response?: Response[];
 }
 
 interface AwarenessMessage<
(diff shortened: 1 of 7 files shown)


packages/store/src/awareness.ts
```diff
@@ -144,8 +182,77 @@ export class AwarenessAdapter<
     } else {
       this._resetRemoteCursor();
     }
+    if (this.getFlag('enable_set_remote_flag')) {
+      this._handleRemoteFlags();
+    }
   };
 
+  private _handleRemoteFlags() {
```
himself65, line 196: This is unsafe. Make sure you only enable this in development mode

packages/store/src/awareness.ts
```diff
@@ -18,12 +18,28 @@ interface UserInfo {
...
+  id: string;
+};
+
 interface AwarenessState<
   Flags extends Record<string, unknown> = BlockSuiteFlags
 > {
   cursor?: SelectionRange;
   user?: UserInfo;
   flags: Flags;
+  changeRequest?: Request<Flags>[];
```
doodlewind, line 41: Where are `changeRequest` and `changeResponse` being used? Didn't see them being used in this PR.

himself65: https://github.com/toeverything/blocksuite/pull/695/files#diff-1f168a5b1666540a4a8099cdf382e91e2590e563317a62e99fcb93b88591adafR198-R205

himself65: https://github.com/toeverything/blocksuite/pull/695/files#diff-1f168a5b1666540a4a8099cdf382e91e2590e563317a62e99fcb93b88591adafR247-R253

himself65: Let me clarify the whole process:
1. One use set a remote flag, then we push a request item into `state.changeRequest`
2. Others get `changeRequest` and check if he needs to update the flag
3. If the request accepted, push a `changeResponse`
4. cleanup the duplicated `changeResponse` and `changeRequest` which has same id

doodlewind: Well but the actual awareness state is using `request` and `response` as field names, rather than `changeRequest` and `changeResponse`.

himself65: my bad

PR #696, opened 2023-01-13
improve selection change

lawvs:
Improve https://github.com/toeverything/blocksuite/pull/569

- add test
- Range should be directionless, so directions should not be added to Range


himself65 (approved):

himself65:
Forgot to update the commit message. Trying to force push

PR #698, opened 2023-01-14
feat: update the css effect of drag handle

Himself65:
Preview: https://blocksuite-git-20230114-drag-handle-css-toeverything.vercel.app/?init

## Regression

The left side of the list block is a drag handle with an unpredictable position; the test cases will somehow be broken due to its position


packages/blocks/src/components/drag-handle.ts
```diff
@@ -64,19 +68,34 @@ export type DragHandleGetModelStateWithCursorCallback = (
...
-      background: rgba(55, 53, 47, 0.08);
+    }
+
+    .affine-drag-handle-rect {
+      width: 11px;
+      height: 11px;
+      border-width: 1px;
+      border-style: solid;
+      border-color: rgba(61, 61, 61, 1);
+      transform: rotate(45deg);
```
himself65: copy from axshare

PR #706, opened 2023-01-14
refactor: use subscribe with filter and selector

himself65:


himself65:
deprecated: use another API in awareness

Issue #648, opened 2023-01-10
Children lost after deleting at line start


doodlewind:
Initial state:

```
123
  456
789
```

After deleting at the beginning of `123`, its child `456` is lost.

https://user-images.githubusercontent.com/7312949/211546024-f4cfec04-13f5-4571-a9cf-254495202b72.mov

SteinsHead:
And I found that if u try to add carriage return, the current line will change to under the children line. And I will try to figure out why dose this problem occer.


https://user-images.githubusercontent.com/50788736/212471013-a1b2d05a-3c80-4e90-b6c0-a36593cee2af.mov

SteinsHead:
Sorry for the video so small, I will re-upload a clearer version.


https://user-images.githubusercontent.com/50788736/212471181-3decb028-eb73-4eab-bc93-c9faa28bde3e.mov

himself65:
I've seen this bug in the database block. This need to be fixed ASAP

himself65:
> Sorry for the video so small, I will re-upload a clearer version.
> 
>  2023-01-14.20-15-04.mov

This is an expected behavior but not correct. Because `456` is a child of the original `123`. You press enter, then `123` go to the next line which is should be top of the `789`

himself65:
We should carry the children of a model when the user press enter or backspace

himself65:
I think the correct logic is when you append a block, you should bring the old data into the new one, and when you delete a block, you should bring data to the father

SteinsHead:
> I think the correct logic is when you append a block, you should bring the old data into the new one, and when you delete a block, you should bring data to the father

I tried to modify the same file in the code, but did not fix the problem. Do I need to make additional changes？

himself65:
I'm fixing this. PR will coming soon

SteinsHead:
Ok, thank you so much 😘

himself65:
PR has opened, but it needs some E2E tests. Welcome everyone who wants to contribute it

doodlewind:
I haven't noticed that this has been fixed, great work here!

Also, according to my inaccurate memory, I thought maybe this was possibly introduced by #479, so /cc @lawvs

PR #707, opened 2023-01-14
fix: bring children to father model

himself65:
Refs: https://github.com/toeverything/blocksuite/issues/648


himself65:
This could also fix the bug in database block

PR #708, opened 2023-01-14
fix: handle children when delete a block

### LINKED ISSUE #648: Children lost after deleting at line start

himself65:
Fixes: https://github.com/toeverything/blocksuite/issues/648


himself65:
Need e2e test in the future

PR #710, opened 2023-01-14
feat: init support for debug log

himself65:


doodlewind (commented):
How can we selectively enable the console debug output? For example, the `debug` package can set `debug.disable()` or configure `*` wildcards ( https://www.npmjs.com/package/debug ). Otherwise, the log may be too verbose.

himself65:
> 

This is a good question. I will make this possible.

But I think the API would something be like

```js
globalThis.debugger.enable('store')
globalThis.debugger.disable('*')
const dispose = globalThis.debugger.traceModel(model)
globalThis.debugger.showHistory(model)
```

packages/playground/vite.config.ts
```diff
@@ -7,6 +7,9 @@ import { hmrPlugin } from './scripts/hmr-plugin';
 // https://vitejs.dev/config/
 export default defineConfig({
   plugins: [react(), hmrPlugin],
+  build: {
+    sourcemap: true,
```
himself65, line 11: enable the source map to help the debugging

PR #712, opened 2023-01-14
fix: delete selected text on enter

pionxzh:
### Before
Lo|**ngT**|ext <- Enter
Lo
|ngText

### After
Lo|**ngT**|ext <- Enter
Lo
|ext


Himself65 (approved):
LGTM

Issue #197, opened 2022-11-10
Incorrect format for copying and deleting of list


JimmFly:
The dots before the first line of the list are not copied and are not deleted together when deleted
![delete](https://user-images.githubusercontent.com/102217452/201037437-b55a38f0-7784-4f75-b2d8-802080ebd051.gif)

doodlewind:
Closed by #715

PR #715, opened 2023-01-15
fix: should keep first line format when pasted into a new line

fundon:
Related: #197 

`insertBlocks` method may need to be refactored.


Himself65 (approved):

doodlewind (approved):
Thanks for fixing this! The prerequisite of refactoring `insertBlocks` issue may by #639

PR #718, opened 2023-01-15
fix: replace text when paste to a block

SteinsHead:
I think the correct logic is that the selected block should be removed first, and then the pasted block should be added to the location of the removed block


SteinsHead:
This code need test in the future

doodlewind (commented):
Thanks for your contribution!

TBO the simple `delete` approach looks a bit overkill. Suppose I copied this list:

```
* 123
* 456
* 789
```

Then I pasted this list into following range selection state:

image

The `Hello foo world` content will be deleted.

But partial deletion of block is also supported in our codebase. How about trying `deleteModelsByRange` instead? Thanks.

P.S. To insert the pasted content into the right position, we may need to split one paragraph block into two, so as to "make room" for the pasted content. To do this we also have `handleBlockSplit` that may help without re-inventing the wheel 🤔.

Certainly what mentioned above could be supported progressively. We can merge this one if test is included, then make room for new issues and PRs for other enhancements.

SteinsHead:
> Thanks for your contribution!
> 
> TBO the simple `delete` approach looks a bit overkill. Suppose I copied this list:
> 
> ```
> * 123
> * 456
> * 789
> ```
> 
> Then I pasted this list into following range selection state:
> 
> image
> 
> The `Hello foo world` content will be deleted.
> 
> But partial deletion of block is also supported in our codebase. How about trying `deleteModelsByRange` instead? Thanks.
> 
> P.S. To insert the pasted content into the right position, we may need to split one paragraph block into two, so as to "make room" for the pasted content. To do this we also have `handleBlockSplit` that may help without re-inventing the wheel 🤔.
> 
> Certainly what mentioned above could be supported progressively. We can merge this one if test is included, then make room for new issues and PRs for other enhancements.

Thanks for your guidance, I think I should know how to deal with this situation!

SteinsHead:
And I want to know why my code can't pass the E2E test 😳
I haven't used testing tools before, so I don't understand what's going on here

himself65:
https://github.com/toeverything/blocksuite/actions/runs/3922314536/jobs/6705441959#step:7:73

SteinsHead:
Sorry, I still don't understand the cause of the error when I see the error. I don't seem to have modified the corresponding file 😵

doodlewind:
> And I want to know why my code can't pass the E2E test 😳 I haven't used testing tools before, so I don't understand what's going on here

This part in our docs should help 👀 https://github.com/toeverything/blocksuite/blob/master/BUILDING.md#testing

Just reproducing `pnpm test` errors locally can be of great help :) Also, the VSCode plugin can be handy for debugging.

SteinsHead:
Ok, thank you very much. I will review my code 😀

himself65:
Please read https://github.com/toeverything/blocksuite/blob/master/BUILDING.md#testing to learn how to test the code. If you have any question, post here

SteinsHead:
> Please read https://github.com/toeverything/blocksuite/blob/master/BUILDING.md#testing to learn how to test the code. If you have any question, post here

Hi, I want to know about the test function `keyboard selection and copy paste` . Does the test function in this part of the function conflict with what paste does? I think the correct answer for this part of the test code should be 'use'. This is line 237 of the test code in the 'code.spec.ts' file

image

himself65:
see 248 and 249

SteinsHead:
But ’use‘ should be selected when the shift key is not raised, so ’use‘ should be replaced by clipboard's content?

himself65:
good idea, I think maybe the test case is wrong

SteinsHead:
I don't think I'm necessarily right, and that might be something to discuss 😣

himself65:
I try the same thing on the notion. our test case is wrong.

himself65:
> I don't think I'm necessarily right, and that might be something to discuss 😣

I talked with my teammates. They say some e2e test is technically right but actually wrong. So...

SteinsHead:
I am also not familiar with the Playwright test tool (actually I have not used the test tool). I think participating in this project has given me a lot of knowledge and promotion. Thank you! I will continue to modify my code 😊

himself65:
The last one test is the PR issue I think

himself65:
> The last one test is the PR issue I think

Because you shouldn't delete the whole block if you just select just a part of the text

packages/editor/src/managers/clipboard/paste-manager.ts
```diff
@@ -255,6 +255,9 @@ export class PasteManager {
             this._addBlocks(blocks.slice(0, 1), parent, index, addBlockIds);
             this._editor.page.deleteBlock(lastBlockModel);
           } else {
+            if (currentSelectionInfo.type === 'Range') {
+              selectedBlock?.text?.delete(0, selectedBlock.text.length);
```
himself65: Delete the block if and only if you select the whole text

himself65 (changes requested):

packages/editor/src/managers/clipboard/paste-manager.ts
SteinsHead: Would I be able to achieve the above effects if I used this method? I'm trying to use `deleteModelsByRange` but it still seems to have some problems

image

himself65: I think you just need to delete the section part and insert

SteinsHead:
I found another test problem in the `clipboard.spec.ts file` : in line 177, function should have the letter `b` checked, so the final string should be `['atext', 'h1c']`

image

tests/clipboard.spec.ts
```diff
@@ -176,10 +176,10 @@ test('split block when paste', async ({ page }) => {
   await page.keyboard.type('abc');
   await setQuillSelection(page, 1, 1);
   await pasteContent(page, clipData);
-  await assertRichTexts(page, ['abtext', 'h1c']);
+  await assertRichTexts(page, ['atext', 'h1', 'c']);
   await assertSelection(page, 1, 2, 0);
   await undoByClick(page);
-  await assertRichTexts(page, ['\n']);
```
himself65, line 187: why it's ac?

SteinsHead: When I use single step debugging in vscode, I find that this step is a retraction, so the result should be 'ac'

SteinsHead: To split the words, I used the ready-made `handleBlockSplit` method, which splits the block of text into two pieces

SteinsHead: I'll upload a video to do the same operation

himself65:
https://user-images.githubusercontent.com/14026360/212605708-2870e4a3-b911-4f15-aeae-42e54669f675.mov

I think you paste a singe line. There shouldn't have a new block. and when you redo it should be `abc`

himself65:
Click redo twice is ok for now

SteinsHead:
https://user-images.githubusercontent.com/50788736/212605980-a656d555-4c7f-42bb-88ee-d0c8e6548578.mov

`b` will be overwritten during operation in the first step above, so `b` will also disappear during redo operation

SteinsHead:
> Click redo twice is ok for now

yeap, I think we need to do a double retraction

SteinsHead:
This part of my code still has problems pasting multiple lines of content, so I think I can continue to improve 😉

himself65 (changes requested):
Nope.

see https://youtu.be/fo5STZvlkRc

Idk why GitHub upload always failed

SteinsHead:
> Nope.
> 
> see https://youtu.be/fo5STZvlkRc
> 
> Idk why GitHub upload always failed

ok, I'll look at what's wrong with my code.

SteinsHead:
I haven't fixed the error when pasting multiple lines of content, it might take a while, this code fixes the problem when pasting a single line of content, I've modified the logic to get the content of the clipboard 😣

himself65:
https://user-images.githubusercontent.com/14026360/212682167-223a2ae4-90a2-45d4-a214-05ad696ac80a.mov

SteinsHead:
I'm working on it 😢

SteinsHead:
Sorry, I am currently unable to fix the bug where the content disappears when pasting multiple lines. I have tried several methods but still cannot solve the problem. But at present, my code may be quite correct when pasting a line of content. I hope I can get some help or modification suggestions. In the next few days, I may not be able to continue to improve the code due to personal reasons. If this affects the code modification of the community, I will close this pr. Thanks again for the community's guidance to me 😣

doodlewind:
I think fallback to 809da06f4cd51f6607eb36d93b1b2cbb78599365 should be adequate, we could handle the multi line pasting in a standalone PR. This looks more tricky than expected, and the `insertBlocks` part needs to be refactored to make more dedicated clipboard handling to happen. Thanks for you engagement!

PR #719, opened 2023-01-15
fix: should not create link in readonly mode

lawvs:


tests/link.spec.ts
```diff
@@ -177,6 +177,23 @@ test('type character in link should not jump out link node', async ({
   );
 });
 
+test('readonly mode should not trigger create link popup', async ({ page }) => {
```
doodlewind, line 180: Can we merge this case with the read-only test case below?

Himself65 (approved):

PR #642, opened 2023-01-10
feat: init block tag & database block

himself65:
Upstream bug: https://github.com/toeverything/blocksuite/issues/648

Preview URL: https://blocksuite-git-20220110-tag-toeverything.vercel.app/?init=database

PR #720, opened 2023-01-15
feat: init block tag & database block

himself65:
Upstream PR: https://github.com/toeverything/blocksuite/pull/642

In the first implementation, the block column will be ignored

## Discussion

https://github.com/toeverything/blocksuite/discussions/717

## Regression

- Drag Handle should prohibit in this part
- Range selection should prohibit in this part

## Limitation

- Each cell should and only can have one model as its child

### DIFF
--- a/packages/blocks/src/page-block/utils/container-operations.ts
+++ b/packages/blocks/src/page-block/utils/container-operations.ts
@@ -415,6 +415,9 @@ export function handleBlockSelectionBatchDelete(
   const parentModel = getModelByElement(parent);
   const index = parentModel?.children.indexOf(models[0]);
   for (let i = 0; i < models.length; i++) {
+    if (models[i].flavour === 'affine:database') {
+      models[i].children.slice().forEach(model => page.deleteBlock(model));
+    }
     page.deleteBlock(models[i]);
   }
   const id = page.addBlock(
--- a/packages/store/src/__tests__/workspace.unit.spec.ts
+++ b/packages/store/src/__tests__/workspace.unit.spec.ts
@@ -102,6 +102,7 @@ describe.concurrent('addBlock', () => {
         'sys:children': [],
         'sys:flavour': 'affine:page',
         'sys:id': '0',
+        'sys:tags': {},
       },
     });
   });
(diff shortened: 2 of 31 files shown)


packages/store/src/__tests__/workspace.unit.spec.ts
```diff
@@ -88,6 +89,7 @@ describe.concurrent('basic', () => {
         versions: {},
       },
       [spaceId]: {},
+      [tagsId]: {},
```
doodlewind, line 92: Here is not using the page-level tags, but the workspace level. Is that expected?

himself65: Yep

himself65: Because tagsId is `tags:page0` which is correct, I think. put the tag map under YDoc is not a big deal

doodlewind: The top-level entries will be SubDocs, so putting tags into the PageBlock under `space:page0` should make more sense here.

himself65: Make sense, but I will do this in the end as this not a big deal for now

doodlewind:
Good to merge once we add an opt-in `enable_database` feature flag to ensure that these changes won't affect the YDoc data structure in AFFiNE (the top-level `getMap` could be a breaking change). API breaking changes can happen in nightly versions at any time, but for data format, we need to be more cautious.

packages/blocks/src/page-block/utils/container-operations.ts
```diff
@@ -415,6 +415,9 @@ export function handleBlockSelectionBatchDelete(
   const parentModel = getModelByElement(parent);
   const index = parentModel?.children.indexOf(models[0]);
   for (let i = 0; i < models.length; i++) {
+    if (models[i].flavour === 'affine:database') {
+      models[i].children.slice().forEach(model => page.deleteBlock(model));
+    }
     page.deleteBlock(models[i]);
```
himself65, line 421: cleanup the row block

himself65:
see https://github.com/toeverything/blocksuite/pull/770

PR #721, opened 2023-01-15
feat: attach surface elements to page ydoc

doodlewind:
Previously the contents in the canvas context were not attached to YDoc. For now, these elements could be collaboratively updated.

image

### DIFF
--- a/packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
+++ b/packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
@@ -278,12 +278,14 @@ export class EdgelessPageBlockComponent
         </div>
         ${hoverRect} ${selectionRect}
         ${selectionState.type !== 'none'
-          ? html` <edgeless-selected-rect
-              .state=${selectionState}
-              .rect=${selectionState.rect}
-              .zoom=${zoom}
-              .readonly=${this.readonly}
-            ></edgeless-selected-rect>`
+          ? html`
+              <edgeless-selected-rect
+                .state=${selectionState}
+                .rect=${selectionState.rect}
+                .zoom=${zoom}
+                .readonly=${this.readonly}
+              ></edgeless-selected-rect>
+            `
           : null}
       </div>
     `;
--- a/packages/blocks/src/surface-block/surface-block.ts
+++ b/packages/blocks/src/surface-block/surface-block.ts
@@ -44,14 +44,10 @@ export class SurfaceBlockComponent extends LitElement {
     const container = new SurfaceContainer(this._canvas, yContainer);
     this._container = container;
 
-    const params = new URLSearchParams(location.search);
-    if (params.get('surface') !== null) {
-      page.awareness.setFlag('enable_surface', true);
-    }
-
     if (page.awareness.getFlag('enable_surface')) {
       bindWheelEvents(this._container.renderer, this.mouseRoot);
 
+      const params = new URLSearchParams(location.search);
       if (params.get('init') !== null) {
         const element0 = new RectElement('0');
         element0.setBound(0, 0, 100, 100);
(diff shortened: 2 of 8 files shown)


packages/store/src/workspace/page.ts
```diff
@@ -106,6 +106,18 @@ export class Page extends Space<PageData> {
     return Array.isArray(this._root) ? this._root[1] : null;
   }
 
+  /** @internal used for getting surface block elements for phasor */
+  get ySurfaceContainer() {
```
doodlewind, line 110: Here looks tricky, but I didn't find a better way to provide a `yContainer` to the `SurfaceContainer` that manages the canvas context. The `SurfaceContainer` that consumes this YMap is driven by YEvent updates, so this is a must-have.

packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
```diff
@@ -278,12 +278,14 @@ export class EdgelessPageBlockComponent
...
-              .rect=${selectionState.rect}
-              .zoom=${zoom}
-              .readonly=${this.readonly}
-            ></edgeless-selected-rect>`
+          ? html`
+              <edgeless-selected-rect
+                .state=${selectionState}
+                .rect=${selectionState.rect}
+                .zoom=${zoom}
+                .readonly=${this.readonly}
```
himself65, line 286: Not related to the PR, but should we access `readonly` by accessing `page.awareness.isReadonly`?

packages/blocks/src/surface-block/surface-block.ts
```diff
@@ -38,12 +42,22 @@ export class SurfaceBlockComponent extends LitElement {
...
-      initMockData(this._renderer, 1000000, 100000, 100000);
+    if (params.get('surface') !== null) {
+      bindWheelEvents(this._container.renderer, this.mouseRoot);
+
+      if (params.get('init') !== null) {
+        const element0 = new RectElement('0');
+        element0.setBound(0, 0, 100, 100);
+        element0.color = 'red';
+        container.addElement(element0);
+        // initMockData(this._container.renderer, 1000000, 100000, 100000);
```
himself65: remove

packages/blocks/src/surface-block/surface-block.ts
```diff
@@ -38,12 +42,22 @@ export class SurfaceBlockComponent extends LitElement {
...
+    const { page } = this.model;
+    const yContainer = page.ySurfaceContainer;
+    const container = new SurfaceContainer(this._canvas, yContainer);
+    this._container = container;
 
     const params = new URLSearchParams(location.search);
-    if (params.get('phasor') !== null) {
-      bindWheelEvents(this._renderer, this.mouseRoot);
-      initMockData(this._renderer, 1000000, 100000, 100000);
+    if (params.get('surface') !== null) {
```
himself65: Will it be better if we use `flag`?

packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
doodlewind: We can have a new PR that removes the `readonly` prop.

packages/blocks/src/surface-block/surface-block.ts
```diff
@@ -38,29 +38,37 @@ export class SurfaceBlockComponent extends LitElement {
...
-    this.model.childrenUpdated.on(() => this.requestUpdate());
-
+  private _initContainer() {
     const { page } = this.model;
     const yContainer = page.ySurfaceContainer;
     const container = new SurfaceContainer(this._canvas, yContainer);
     this._container = container;
 
     const params = new URLSearchParams(location.search);
     if (params.get('surface') !== null) {
```
himself65: Put this logic into `playground`🤔

doodlewind: Good catch, I think this can be a pattern to have not-enabled-by-default features in the playground.

Issue #619, opened 2023-01-09
keyboard operation is broken in image caption


lawvs:
https://user-images.githubusercontent.com/18554747/211246141-22044784-1d5d-4013-b0fb-affca2e55f55.mov

Issue #682, opened 2023-01-13
Undo/redo not avaliable in title


Himself65:

doodlewind:
The undo/redo hotkey doesn't trigger `page.undo|redo` when editing `textarea`. Using the `undoByClick` should work but the cursor is not kept.

lawvs:
> The undo/redo hotkey doesn't trigger `page.undo|redo` when editing `textarea`. Using the `undoByClick` should work but the cursor is not kept.

I skipped title undo/redo in https://github.com/toeverything/blocksuite/pull/548 (https://github.com/toeverything/blocksuite/pull/548/commits/01e55bee5f88e8f4b92b4a845ea50c7969db8f27) to prevent undo sometimes triggers multiple times（native textarea undo and hotkey undo occur together）

https://github.com/toeverything/blocksuite/blob/b09db146d9f282b7d095e130b6ac68f260ad74d3/packages/blocks/src/page-block/utils/bind-hotkey.ts#L28-L39

doodlewind:
> > The undo/redo hotkey doesn't trigger `page.undo|redo` when editing `textarea`. Using the `undoByClick` should work but the cursor is not kept.
> 
> I skipped title undo/redo in #548 ([01e55be](https://github.com/toeverything/blocksuite/commit/01e55bee5f88e8f4b92b4a845ea50c7969db8f27)) to prevent undo sometimes triggers multiple times（native textarea undo and hotkey undo occur together）
> 

This should do the trick, is anyone trying for a PR?

``` ts
if (isPageTitle(e)) { 
  e.preventDefault();
} 
page.undo();
```

PR #724, opened 2023-01-15
fix: uniform event handler outside of rich text

lawvs:
fix https://github.com/toeverything/blocksuite/issues/619 https://github.com/toeverything/blocksuite/issues/682

- Filter most of the events whose target is non-rich-text from hotkey (fix most of the keyboard issues in non rich text input)
- Clear workaround for input and title in event hander
- Add test for #682 
- Add test for non-rich-text input, including
  - create/edit link input
  - code block language search input
  - image caption input


doodlewind (approved):
This looks great, we are moving towards a common focus filter that doesn't over-capture events outside the editor context, thanks!

PR #728, opened 2023-01-16
test: use github reporter

lawvs:
#TIL 

but the flaky test is also annotated.

see https://playwright.dev/docs/test-reporters#github-actions-annotations

Screenshot 2023-01-16 at 2 15 53 PM


doodlewind (approved):
LGTM, we can try for a while and see if it can help locating flaky test cases.

PR #729, opened 2023-01-16
feat: keep deleting forward with cmd+backspace

chenmoonmo:
This PR adds the same handling to "command + backspace" as "backspace" to support smooth paragraph deletion.
The following video contains a comparison with the notion and obsidian operations.

https://user-images.githubusercontent.com/36295999/212618163-f9e46652-7bec-46a0-b450-d5199ed6214e.mp4

### DIFF
--- a/packages/blocks/src/page-block/default/utils.ts
+++ b/packages/blocks/src/page-block/default/utils.ts
@@ -1,9 +1,6 @@
 import type { BaseBlockModel, Page } from '@blocksuite/store';
 import { toast } from '../../components/toast.js';
-import {
-  handleLineStartBackspace,
-  isAtLineEdge,
-} from '../../__internal__/rich-text/rich-text-operations.js';
+import { isAtLineEdge } from '../../__internal__/rich-text/rich-text-operations.js';
 import {
   asyncFocusRichText,
   focusNextBlock,
@@ -21,7 +18,6 @@ import {
   resetNativeSelection,
   isCaptionElement,
   doesInSamePath,
-  isCollapsedAtBlockStart,
 } from '../../__internal__/utils/index.js';
 import type { PageBlockModel } from '../page-model.js';
 import {
@@ -542,17 +538,8 @@ export function bindHotkeys(
     const { state } = selection;
     if (state.type === 'none') {
       // Will be handled in the `keyboard.onBackspace` function
-      const target = e.target as HTMLElement;
-      const model = getModelByElement(target);
-      const richText = getRichTextByModel(model);
-      assertExists(richText);
-      const quill = richText?.quill;
-      if (isCollapsedAtBlockStart(quill)) {
-        handleLineStartBackspace(page, model);
-        return;
-      }
+      return;
     }
-
     if (state.type === 'native') {
       handleMultiBlockBackspace(page, e);
       return;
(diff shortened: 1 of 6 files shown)


doodlewind (commented):
Thanks for your contribution! If I understand correctly, our current playground already supports `cmd-backspace` for single line deletion. The difference is that for custom behavior, e.g., pressing backspace at the start of an empty list, `cmd+backspace` doesn't "keep deleting", and IMO we only need to modify the `HOTKEYS` binding to make this happen (rather than adding quill binding). Also, would you mind adding a test case to verify this? Thanks.

chenmoonmo:
Yes, that's what I'm trying to say, I've modified the code as I understand it and added test cases, but I'm not sure if this meets your requirements.
BTW, this is the first test case I wrote in my life.

himself65:
rebase please?

packages/blocks/src/page-block/default/utils.ts
```diff
@@ -337,7 +341,15 @@ export function bindHotkeys(
 
   hotkey.addListener(BACKSPACE, e => {
     const { state } = selection;
-    if (state.type === 'native') {
+    const target = e.target as HTMLElement;
```
doodlewind, line 344: This doesn't feel right since it adds an `isCollapsedAtBlockStart` before the simple conditions based on `state.type`. 🤔 Try merging the latest refactored branch and you'll see the inconsistence.

chenmoonmo:
I feel screwed up.

chenmoonmo:
I feel screwed up.

himself65:
？

himself65:
I think you git histry crashed, let me handle this

himself65:
I also feel lost, don't know what the code should be...

himself65:
I think now the code is clean @chenmoonmo
does this code make sense to you?

himself65 (approved):

packages/blocks/src/__internal__/utils/consts.ts
```diff
@@ -6,7 +6,7 @@ export const ALLOW_DEFAULT = true;
 export const HOTKEYS = {
   UNDO: 'command+z,ctrl+z',
   REDO: 'command+shift+z,ctrl+shift+z,ctrl+y',
-  BACKSPACE: 'backspace',
+  BACKSPACE: 'backspace,command+backspace',
```
lawvs, line 9: Is `crtl+backspace` should be added?

chenmoonmo:
I tried to merge and rebase the branches @Himself65, sorry for my mistakes

lawvs:
I refactor the code and now it is ready to merge.

tests/slash-menu.spec.ts
```diff
@@ -38,9 +37,17 @@ test.describe('slash menu should show and hide correctly', () => {
...
       await page.mouse.click(0, 50);
     }
-    await focusRichText(page);
     // Clear input
-    await page.keyboard.press(`${SHORT_KEY}+Backspace`);
+    await page.evaluate(paragraphId => {
+      const { page } = window;
+      const block = page.getBlockById(paragraphId);
+      if (!block) {
+        throw new Error('Block not found, id: ' + paragraphId);
```
himself65, line 45: use `assertExists`

himself65 (changes requested):

tests/slash-menu.spec.ts
lawvs: This is in the test file

doodlewind:
Thanks @chenmoonmo for the enhancement! BTW we are also hosting Chinese community chat in https://t.me/blocksuite_cn, welcome!

PR #731, opened 2023-01-16
feat: init blockhub

thorseraq:


packages/blocks/src/components/blockhub.ts
```diff
@@ -0,0 +1,648 @@
...
+  },
+  {
+    flavour: 'affine:list',
+    type: 'todo',
+    name: 'To-do List',
+    description: 'Track tasks with a to-do list.',
+    icon: TodoIcon,
+    toolTip: 'Drag to insert To-do List.',
+  },
+];
```
lawvs, line 150: There is a very similar configuration at `packages/blocks/src/page-block/utils/const.ts`. Will it come in handy?

thorseraq: If I extend the configuration at `packages/blocks/src/page-block/utils/const.ts` with two more filed: `description`, `toolTip`, will it break it's original usage? It seems safe to do so, and it will be handy with only one more divider manually extended needed

thorseraq: I find the item order is slightly different, this will affect the order to display, and list items are categorized differently, how about just keep it

himself65: Put it to the `packages/globals/const.ts`?

himself65:
rebase/

packages/blocks/src/components/blockhub.ts
thorseraq: Now the icons and consts in `packages/blocks` have been moved to `packages/globals/config`

packages/global/package.json
```diff
@@ -26,6 +27,7 @@
     ]
   },
   "dependencies": {
-    "ansi-colors": "^4.1.3"
+    "ansi-colors": "^4.1.3",
+    "lit": "^2.6.1"
```
himself65, line 33: Do not put lit into global? because I want to make it pure

thorseraq: lit is installed to use `import { svg } from 'lit';`, as I think export icons in global is good 👀, what do you think

himself65: Looks fine? I'm not sure whether it has the side effect

himself65 (approved):

packages/global/package.json
thorseraq: It seems fine when building, I'll track this further

himself65 (changes requested):
Also update https://github.com/toeverything/blocksuite#current-status-blocksuiteeditor

packages/blocks/src/page-block/default/selection-manager.ts
```diff
@@ -242,6 +289,9 @@ export class DefaultSelectionManager {
     if (this.page.awarenessAdapter.getFlag('enable_drag_handle')) {
       createHandle();
     }
+    if (this.page.awarenessAdapter.getFlag('enable_block_hub')) {
```
himself65, line 312: Subscribe to disable the flag?

packages/store/src/workspace/page.ts
```diff
@@ -458,6 +460,34 @@ export class Page extends Space<PageData> {
     });
   }
 
+  insertBlock(
```
himself65, line 462: add a @debug('CRUD') on top

himself65 (changes requested):

packages/store/src/workspace/page.ts
```diff
@@ -458,6 +460,34 @@ export class Page extends Space<PageData> {
...
+    blockProps: OptionalBlockProps,
```
himself65: use `Partial<BlockProps>`

himself65 (changes requested):

thorseraq:
Requested changed updated here:
https://github.com/toeverything/blocksuite/pull/731/commits/07527fffdc1c61da76a8f2b60ad7008952232688
/cc @Himself65

himself65 (approved):
LGTM

PR #735, opened 2023-01-16
feat: discontinuous image blobs uploading support

doodlewind:
This comes from #733


doodlewind:
@SaikaSakura @darkskygit any idea how can we test this feature?

doodlewind:
@SaikaSakura @DiamondThree if necessary, you can use this nightly version (published from this branch, with current master merged in) to test in AFFiNE https://github.com/toeverything/blocksuite/actions/runs/3935500461

packages/store/src/blob/providers.ts
```diff
@@ -117,19 +146,59 @@ export class BlobCloudSync {
...
+    onUploadFinished: Signal<BlobId>
+  ) {
+    this._onUploadFinished = onUploadFinished;
     this._fetcher = ky.create({
       prefixUrl,
       signal: this._abortController.signal,
       throwHttpErrors: false,
     });
 
+    const uploadingProxy = new Proxy(new Set<string>(), {
```
darkskygit, line 170: Why can't we emit signal at the code where set uploading is located?

packages/store/src/blob/providers.ts
```diff
@@ -152,11 +221,24 @@ export class BlobCloudSync {
...
+        }
+      }
+    });
   }
 
   private async _handleTaskRetry(task: SyncTask, status?: BlobStatus) {
+    this._uploadingIds.delete(task.id);
     if (status?.exists) {
       await this._pending.delete(task.id);
+      await this._uploading.set(task.id, true);
```
darkskygit, line 240: Maybe we can combine the var of padding and uploading? padding uses `type = 'add' | 'delete'` to indicate that it is being added or deleted, and we can add a type of 'updated' to indicate that the upload has been successful

darkskygit (changes requested):

packages/store/src/blob/providers.ts
SaikaSakura: Nice I will try only use _pending database .

doodlewind:
Closed in favor of #750

darkskygit:
> Closed in favor of #750

that pr merge into 0117/blob branch, so this pr still works. i will turn it on again.

PR #742, opened 2023-01-17
refactor: move drag indicator to a global singleton pattern

thorseraq:
Now drag indicator lifecycle is created and and removed in Drag handler, as we have may have more drag scenarios using drag indicator, as blockHub so far, and there will be only one dragging at a time. So how about making it globally exists, and after a component finished using it, hide it rather than removing it. In this way any components with dragging behavior can share one indicator

### DIFF
--- a/packages/blocks/src/components/drag-handle.ts
+++ b/packages/blocks/src/components/drag-handle.ts
@@ -131,6 +131,12 @@ export class DragHandle extends LitElement {
     this._indicator = <DragIndicator>(
       document.querySelector('affine-drag-indicator')
     );
+    if (!this._indicator) {
+      this._indicator = <DragIndicator>(
+        document.createElement('affine-drag-indicator')
+      );
+      document.body.appendChild(this._indicator);
+    }
   }
 
   public show(startModelState: EditingState) {
--- a/packages/blocks/src/components/drag-handle.ts
+++ b/packages/blocks/src/components/drag-handle.ts
@@ -128,6 +128,9 @@ export class DragHandle extends LitElement {
   protected firstUpdated() {
     this.setAttribute('draggable', 'true');
     this.style.display = 'none';
+    this._indicator = <DragIndicator>(
+      document.querySelector('affine-drag-indicator')
+    );
   }
 
   public show(startModelState: EditingState) {
@@ -158,10 +161,6 @@ export class DragHandle extends LitElement {
     );
     document.body.addEventListener('wheel', this._onWheel);
     window.addEventListener('resize', this._onResize);
-    this._indicator = <DragIndicator>(
-      document.createElement('affine-drag-indicator')
-    );
-    document.body.appendChild(this._indicator);
     this.addEventListener('mousedown', this._onMouseDown);
     isFirefox &&
       document.addEventListener('dragover', this._onDragOverDocument);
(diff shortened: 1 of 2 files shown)


packages/blocks/src/components/drag-handle.ts
```diff
@@ -128,6 +128,9 @@ export class DragHandle extends LitElement {
   protected firstUpdated() {
     this.setAttribute('draggable', 'true');
     this.style.display = 'none';
+    this._indicator = <DragIndicator>(
+      document.querySelector('affine-drag-indicator')
+    );
```
lawvs, line 133: Use the query API in lit.

```ts
@query('affine-drag-indicator')
_indicator!: DragIndicator
```

doodlewind (approved):

packages/blocks/src/components/drag-handle.ts
doodlewind: I guess `@query` only works in the same component?

thorseraq: I can't get `affine-drag-indicator` this way as it's outside of DragHandle now (globally)

packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -385,6 +385,7 @@ export class DefaultPageBlockComponent
           </div>
           ${childrenContainer}
         </div>
+        <affine-drag-indicator></affine-drag-indicator>
```
lawvs, line 388: IMO it is an anti-pattern to putting `affine-drag-indicator` outside of `drag-handle`

lawvs (changes requested):

packages/blocks/src/page-block/default/default-page-block.ts
thorseraq: The cost of each components with dragging behavior maintaining their own `affine-drag-indicator`, can you suggest a better way to resolve this?

packages/blocks/src/components/drag-handle.ts
```diff
@@ -158,10 +161,6 @@ export class DragHandle extends LitElement {
     );
     document.body.addEventListener('wheel', this._onWheel);
     window.addEventListener('resize', this._onResize);
-    this._indicator = <DragIndicator>(
-      document.createElement('affine-drag-indicator')
-    );
-    document.body.appendChild(this._indicator);
```
lawvs, line 164: I think it's better 

```ts
this._indicator = <DragIndicator>querySelector('affine-drag-indicator')
if (this._indicator) {
  const indicator = <DragIndicator>(
    document.createElement('affine-drag-indicator')
  );
  document.body.appendChild(indicator);
  this._indicator = 
}
```

packages/blocks/src/page-block/default/default-page-block.ts
doodlewind, line 388: Querying this static component using DOM API doesn't look very right. This may also break in the edgeless mode. Is there a way for us to use slot-based events?

thorseraq: > Slot-based events are a type of event that is associated with a specific time and location. These events are often used in calendar or scheduling applications and are designed to be scheduled or rescheduled as needed. The "slot" refers to the specific time and location that the event is scheduled to take place. This allows users to easily view and manage their schedule and see when and where events are taking place.

Does slot-based events mean as above? I think I get the point, I'll try this

lawvs (approved):

Himself65 (approved):

PR #751, opened 2023-01-17
fix: use `requestAnimationFrame`

himself65:


lawvs (commented):
I think we can add a `nextFrame` function to utils

himself65:
> 

I think this's fine for now

PR #762, opened 2023-01-17
build: add eslint to avoid import package itself

Himself65:
Refs: https://github.com/toeverything/blocksuite/pull/571#discussion_r1072332350

/cc @lawvs


doodlewind (approved):
LGTM

PR #763, opened 2023-01-17
test: make test pass in local

fi3ework:
## why
I found that I can't make the `pnpm test` pass following the `BUILDING.md` tutorial, even on a clean Codespaces on GitHub. 🫠 Guessing some details are missing in the documentation so I tweak the playwright config file. And now, I can pass the `pnpm test` on my mac.

why remove `actionTimeout`?
I found when running on my mac (6 core, 12 workers), playwright will be failed on `actionTimeout` timeout. I think maybe too many workers are running concurrently so it stuck.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -56,7 +56,7 @@ When you contribute new features and bug fixes, the test cases are encouraged an
 
 We use [Playwright](https://playwright.dev/) for E2E test, and [vitest](https://vitest.dev/) for unit test.
 
-To test locally, please make sure browser binaries are already installed via `npx playwright install` and Vite playground is started with `pnpm dev`. Then there are multi commands to choose from:
+To test locally, please make sure browser binaries are already installed via `npx playwright install`. Then there are multi commands to choose from:
 
 ```sh
 # run tests in headless mode in another terminal window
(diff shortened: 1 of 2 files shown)


doodlewind (changes requested):
Looks like this will run `pnpm dev` before each `pnpm test`, but that's not our current workflow. We generally keep `pnpm dev` in a separate terminal locally, and I thought this is mentioned in BUILDING.md 👀

> To test locally, please make sure browser binaries are already installed via `npx playwright install` and Vite playground is started with `pnpm dev`.

lawvs:
Maybe `reuseExistingServer: !process.env.CI` option can help us simplify workflow

https://playwright.dev/docs/test-advanced#launching-a-development-web-server-during-the-tests

fi3ework:
😅 I did miss some text when reading `BUILDING.md`, sorry about that. Following @lawvs 's suggestion and simplified the local test command.

Himself65 (approved):

fi3ework:
And still, need to raise the timeout limitation when running on a 6-cores Intel mac. 🤔

doodlewind (approved):

PR #773, opened 2023-01-18
chore: add animation for blockHub, fix mouse cursor bug

thorseraq:
Add animation for blockHub menu when expanding and shrinking

https://user-images.githubusercontent.com/20554850/213158323-92fac303-4e02-487c-bb81-77ab1da2d4b3.mov

### DIFF
--- a/packages/blocks/src/components/blockhub.ts
+++ b/packages/blocks/src/components/blockhub.ts
@@ -39,6 +39,9 @@ export class BlockHub extends NonShadowLitElement {
   @query('.new-icon')
   private _blockHubButton!: HTMLElement;
 
+  @query('.block-hub-icons-container')
+  private _blockHubIconsContainer!: HTMLElement;
+
   private _onDropCallback: (e: DragEvent, lastModelState: EditingState) => void;
   private _getBlockEditingStateByPosition: DragHandleGetModelStateCallback | null =
     null;
@@ -56,6 +59,7 @@ export class BlockHub extends NonShadowLitElement {
   private _isCardListVisiable = false;
   private _cardvisibleType = '';
   private _showToolTip = true;
+  private _timer: number | null = null;
 
   static styles = css`
     .affine-block-hub-container {
@@ -79,6 +83,7 @@ export class BlockHub extends NonShadowLitElement {
       top: unset;
       bottom: 0px;
       transform: unset;
+      right: calc(100% + 5.5px);
     }
 
     .visible {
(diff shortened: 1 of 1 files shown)


packages/blocks/src/components/blockhub.ts
```diff
@@ -394,13 +415,32 @@ export class BlockHub extends NonShadowLitElement {
...
     }
     this.requestUpdate();
   };
 
+  /**
+   * This is currently a workaround, as the height of the .block-hub-icons-container is determined by the height of its
+   * content, and if its child's opacity is set to 0 during a transition, its height won't change, causing the background
+   * to exceeds its actual visual height. So currently we manually set the height of those whose opacity is 0 to 0px.
+   */
+  private _toggleHide = () => {
```
himself65: use `animationend` event

https://developer.mozilla.org/en-US/docs/Web/API/Element/animationend_event

Himself65 (changes requested):

packages/blocks/src/components/blockhub.ts
thorseraq: updated:
https://github.com/toeverything/blocksuite/pull/773/commits/69f483743dbdb44ea33cb57f6f38ca52de98fb87

Himself65 (approved):

packages/blocks/src/components/blockhub.ts
```diff
@@ -303,7 +350,10 @@ export class BlockHub extends NonShadowLitElement {
 
   private _blockHubMenuTemplate = () => {
     return html`
-      <div style=${this._expanded ? 'display:block' : 'display:none'}>
+      <div
+        class="block-hub-icons-container"
+        transition=${this._expanded ? 'true' : 'false'}
```
lawvs, line 355: lit support [Boolean Attributes](https://lit.dev/docs/templates/expressions/#boolean-attribute-expressions)

```ts
html`<div ? transition=${this._expanded}></div>`
```

thorseraq: updated:
https://github.com/toeverything/blocksuite/pull/773/commits

thorseraq:
> At slow speeds, the transition still has some glitches.
> 
>  Screen.Recording.2023-01-18.at.9.28.37.PM.mov

I changed to delay from 50ms to 200ms. But you can still see this.

The height is set to 0px by `this._blockHubIconsContainer.style.height = '0px';` 200ms after the transition start, during which time, the downward transition is finished. After 200ms, the container height should be decreased, as explained  below, to let the background show correctly.

```ts
  /**
   * This is currently a workaround, as the height of the _blockHubIconsContainer is determined by the height of its
   * content, and if its child's opacity is set to 0 during a transition, its height won't change, causing the background
   * to exceeds its actual visual height. So currently we manually set the height of those whose opacity is 0 to 0px.
   */
```

doodlewind (approved):

PR #775, opened 2023-01-18
feat: support resizing surface element

doodlewind:


himself65:
How to resize it? I cannot resize in playground

doodlewind:
This only updates the element bound (in shape mode) dynamically. I will open a new PR to pick and resize element in default mode.

PR #780, opened 2023-01-18
refactor: `AwarenessAdapter` -> `AwarenessStore`

doodlewind:
`AwarenessStore` reflects the definition of this module better. It's our data model API built on top of `YAwareness` (that produces `RawAwarenessState`).

### DIFF
--- a/packages/store/src/awareness.ts
+++ b/packages/store/src/awareness.ts
@@ -1,6 +1,6 @@
 import * as Y from 'yjs';
 import type { RelativePosition } from 'yjs';
-import type { Awareness } from 'y-protocols/awareness.js';
+import type { Awareness as YAwareness } from 'y-protocols/awareness.js';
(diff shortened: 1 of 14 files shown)


packages/store/src/awareness.ts
```diff
@@ -43,35 +44,27 @@ export type AwarenessState<
...
+interface AwarenessEvent<
   Flags extends Record<string, unknown> = BlockSuiteFlags
 > {
   id: number;
   type: 'add' | 'update' | 'remove';
-  state?: AwarenessState<Flags>;
+  state?: RawAwarenessState<Flags>;
 }
 
-export interface AwarenessMetaMessage<
```
doodlewind, line 54: @Himself65 I removed this interface since looks like it's not being used. Subjectively, the `MetaMessage` naming also makes it a bit too opaque for me to understand.

PR #783, opened 2023-01-19
fix: reuse comparePoint to decide dragging direction during native selection

### LINKED ISSUE #697: Vertical vibration affects the recognization of native range selection
https://user-images.githubusercontent.com/18554747/212409958-98b24f86-40df-4c8d-a477-964a117940e1.mov

thorseraq:
Fixes #697 

The problem of `comparePoint()` is when the cursor is not in the text area, its behavior is not stable. But when the cursor is in the text area, it seems very ideal

https://user-images.githubusercontent.com/20554850/213331852-fbb3f174-f49f-424b-97e0-9aef442c524a.mov

https://user-images.githubusercontent.com/20554850/213332527-944909b6-b8e0-4838-8427-c39e33cc9270.mov


thorseraq:
Combine comparePoint with vertical offset, as when the cursor moves to none-text area causing comparePoint unstable, the vertical offset is already big enough

doodlewind (approved):
Thanks for the quick fix!

PR #785, opened 2023-01-19
feat: init drag handle ui

xell:


doodlewind:
Let's continue this in #698

PR #788, opened 2023-01-19
fix: embed component lifecycle warning

### LINKED ISSUE #285: Lit lifecycle console warning
Warning message thrown from `pnpm test`:

> Element **embed-block** scheduled an update (generally because a property was set) after an update completed, causing a new update to be scheduled. This is inefficient and should be avoided unless the next update can only be scheduled as a side effect of the previous update. See https://lit.dev/msg/change-in-update for more information.

@DiamondThree any ideas?

doodlewind:
Fixes #285 

Console warning message:

> reactive-element.ts:76 Element **affine-embed** scheduled an update (generally because a property was set) after an update completed, causing a new update to be scheduled. This is inefficient and should be avoided unless the next update can only be scheduled as a side effect of the previous update. See https://lit.dev/msg/change-in-update for more information.

/cc @DiamondThree

### DIFF
--- a/packages/blocks/src/embed-block/embed-block.ts
+++ b/packages/blocks/src/embed-block/embed-block.ts
@@ -47,10 +47,15 @@ export class EmbedBlockComponent extends NonShadowLitElement {
   _caption!: string;
 
   firstUpdated() {
-    this._caption = this.model?.caption ?? '';
-    if (this._caption) {
-      this._input.classList.add('caption-show');
-    }
+    requestAnimationFrame(() => {
+      this._caption = this.model?.caption ?? '';
+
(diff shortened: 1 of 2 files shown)


packages/blocks/src/embed-block/embed-block.ts
```diff
@@ -47,10 +47,15 @@ export class EmbedBlockComponent extends NonShadowLitElement {
   _caption!: string;
 
   firstUpdated() {
-    this._caption = this.model?.caption ?? '';
-    if (this._caption) {
```
doodlewind, line 51: @DiamondThree please avoid DOM mutation in `firstUpdated`.

PR #791, opened 2023-01-19
feat: add append-flavor slash menu behavior

lawvs:
This PR:

- Enabled the slash menu in the playground by default.
- Adds append-flavor slash flag (with URL param `?slash=1`). In this mode, pressing enter will create a new block.
- Simplifies `updateBlockType`.


doodlewind (approved):
LGTM

Issue #335, opened 2022-12-16
Browser support tiers and on-demand E2E browser matrix testing


doodlewind:
For speeding up CI, we only run E2E tests against Chromium by default, which is our Tier 0 supported browser.

But the situation is that we have 1/3 of seed users using Firefox. And with Tauri, we need to support WebKit on macOS (WebView 2 on Windows is Chromium-based, good news for us). So Firefox and Safari should be our Tier 1 browsers.

The basic support strategy could be:

* For the latest Tier 0 browsers, all master commits should not break its E2E test cases.
* For the latest Tier 1 browsers, we run E2E tests against them daily ([fully supported by Playwright](https://playwright.dev/docs/test-configuration#multiple-browsers)), fixing regressions ASAP. Meanwhile, for major refactoring or compatibility-related PRs, this CI process can be manually triggered by commenting on PR.
* For other browsers and older versions of T0/T1 browsers, we accept bug reports and decide the priority case by case.

Tasks here:

1. Support daily scheduled E2E browser support matrix testing, just like what we do in #306.
2. Support triggering this test suite manually on GitHub.

cc @tzhangchi

doodlewind:
Closed by #793 and #796, thanks @fi3ework for the help!

PR #793, opened 2023-01-19
ci: add browser compatibility cron job test

fi3ework:
## why
The first step for #335 .
The naming is a bit ugly somehow. The prefix `TEST_` is to avoid collision with playwright's official environment naming convention(in the future). Any good naming ideas are welcomed.

image


Himself65 (approved):
/cc LGTM but need @doodlewind approval

doodlewind (approved):
LGTM, thanks for initiating our testing matrix!

PR #795, opened 2023-01-20
chore: bump version

Himself65:


packages/store/src/doc-providers.ts
```diff
@@ -48,16 +48,8 @@ const signaling = isLocalhost ? LOCAL_SIGNALING : DEFAULT_SIGNALING;
 export class DebugDocProvider extends WebrtcProvider implements DocProvider {
   constructor(room: string, doc: Y.Doc, options?: { awareness?: Awareness }) {
     super(room, doc, {
-      awareness: options?.awareness ?? null,
+      awareness: options?.awareness ?? undefined,
```
doodlewind, line 51: Personally I prefer explicitly passing in `null`, but here `options?.awareness` without `??` looks adequate.

himself65: the type of y-webrtc has changed which disallow passing a `null` here

PR #798, opened 2023-01-21
fix: parse partially select text in `sliceToDelta`

### LINKED ISSUE #764: Inaccurate copy range when partially select text inside a paragraph
For a paragraph with content `hello world 123`, range select `world` and `cmd-c` will wrongly copy `hello world` into the clipboard.

This is not related to #718 since it existed in the previous master.

https://user-images.githubusercontent.com/7312949/213059887-cffdfac7-7f1e-4182-9fa2-580afd02f2fe.mov

pingren:
Will close #764 

CleanShot 2023-01-21 at 12 44 34


doodlewind (approved):
Thanks for the fix, hope next time we can see your test case that makes your contribution solid. Happy new year!

PR #808, opened 2023-01-22
feat: switch to handwriting style canvas rect shape

doodlewind:
This follows #806 

To test online, the preview page should be opened with the `&surface` param.

image


doodlewind:
## Quick Note

To convert SVG path command string into canvas `Path2D`, following code snippet could be used (needs further debugging):

``` ts
function parse(pathCommand: string): Path2D | null {
  const commands = pathCommand.match(/[a-df-z][^a-df-z]*/gi);
  const path = new Path2D();

  if (!commands) return null;

  for (const command of commands) {
    console.log(command);
    const type = command.charAt(0);
    const coordStr = command.slice(1).trim();
    const coords = coordStr.split(/[\s,]+/).map(a => parseFloat(a));

    let currX = 0;
    let currY = 0;

    switch (type) {
      case 'M':
        path.moveTo(coords[0], coords[1]);
        currX = coords[0];
        currY = coords[1];
        break;
      case 'L':
        path.lineTo(coords[0], coords[1]);
        currX = coords[0];
        currY = coords[1];
        break;
      case 'H':
        path.lineTo(coords[0], currY);
        break;
      case 'V':
        path.lineTo(currX, coords[0]);
        break;
      case 'C':
        path.bezierCurveTo(
          coords[0],
          coords[1],
          coords[2],
          coords[3],
          coords[4],
          coords[5]
        );
        currX = coords[4];
        currY = coords[5];
        break;
      case 'S':
        path.bezierCurveTo(coords[0], coords[1], coords[2], coords[3]);
        break;
      case 'Q':
        path.quadraticCurveTo(coords[0], coords[1], coords[2], coords[3]);
        break;
      case 'T':
        path.quadraticCurveTo(coords[0], coords[1]);
        break;
      case 'A':
        path.arcTo(coords[0], coords[1], coords[2], coords[3], coords[4]);
        break;
      case 'Z':
        path.closePath();
        break;
      default:
        throw new Error(`Unsupported command: ${type}`);
    }
  }

  return path;
}
```

But this doesn't have to be implemented manually in our codebase, since modern browsers supports `new Path2D(commandStr)`. Or else we can use https://github.com/akira-cn/svg-path-to-canvas

PR #812, opened 2023-01-23
refactor: clean usage for cross-env

fi3ework:
## why

1. use pnpm's shell-emulator to replace usage of `cross-env` for Windows. https://pnpm.io/cli/run#shell-emulator
2. add missing process env for CI by using `environment` property.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+shell-emulator=true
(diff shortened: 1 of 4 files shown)


.npmrc
```diff
@@ -0,0 +1 @@
+shell-emulator=true
```
fi3ework, line 1: Guess no need to add a new line here as Prettier does not do that automatically. 🤔

doodlewind (approved):
LGTM

PR #826, opened 2023-01-23
style: check subject in commitlint

Himself65:
image


doodlewind:
Adding this check makes sense, but I'm not sure if it deserves adding about 1/3 in our dependency list for this feature. Can we have configurable or hard-coded packages scope instead?

PR #858, opened 2023-01-26
build(playground): support lan connection

doodlewind:


Himself65 (approved):

lawvs:
I think there may be potential security risks in having it on by default.

doodlewind:
> I think there may be potential security risks in having it on by default.

This doesn't apply to our online playground, so I guess it should be just fine. This change helps testing collab-editing between different devices under the same Wi-Fi. I'm using this for sharing clipboard between my Mac and Steam Deck 🤪

Issue #151, opened 2022-10-31
Remove hard-coded fields in store


doodlewind:
All the code path with `TODO use schema` comments in the store package should be refactored.

doodlewind:
Closed by #859

PR #859, opened 2023-01-26
feat(blocks,store): use schema in block model

himself65:
Related: https://github.com/toeverything/blocksuite/issues/151


packages/store/src/base.ts
```diff
@@ -2,6 +2,29 @@ import type { Page } from './workspace/index.js';
...
+  version: z.number(),
+  schema: z.object({
+    flavour: z.string(),
+    tag: z.string(),
+    type: z.string().optional(),
+    defaultProps: z.record(z.any()),
+  }),
+});
+
+const CodeBlockSchema = {
```
doodlewind, line 17: I think we should not put the per-block definition in the `store` package.

himself65: just a example here

himself65:
blocking because there is need to discussion about the design

himself65:
After discussion. This code needs a little change

packages/store/src/base.ts
```diff
@@ -10,22 +10,15 @@ const TagSchema = z.object({
...
 
-const StateSchema = z.any();
-
 export const BlockSchema = z.object({
   version: z.number(),
   model: z.object({
     flavour: FlavourSchema,
     tag: TagSchema,
-    state: StateSchema,
+    state: z.function().returns(z.record(z.any())),
```
himself65, line 18: we use `() => state` here to avoid the Inconsistent state

doodlewind (approved):
LGTM, a good start!

PR #860, opened 2023-01-26
refactor(blocks): remove redundant `type` field in code block

doodlewind:
/cc @thorseraq

### DIFF
--- a/packages/blocks/src/code-block/code-model.ts
+++ b/packages/blocks/src/code-block/code-model.ts
@@ -7,7 +7,6 @@ export class CodeBlockModel
 {
   static version = 1;
   flavour = 'affine:code' as const;
-  type = 'code' as const;
   tag = literal`affine-code`;
 
   language: string;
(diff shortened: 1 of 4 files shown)


packages/blocks/src/code-block/code-model.ts
```diff
@@ -7,7 +7,6 @@ export class CodeBlockModel
 {
   static version = 1;
   flavour = 'affine:code' as const;
-  type = 'code' as const;
```
lawvs, line 10: In `BaseBlockModel`, the `type` field is required, removing it directly is unsafe. For example, it will break the format bar in the code block(I can fix it by switching to `flavour+type`).

Screenshot 2023-01-27 at 2 03 00 AM

Screenshot 2023-01-27 at 2 05 17 AM

Screenshot 2023-01-27 at 2 06 37 AM

https://github.com/toeverything/blocksuite/blob/4f085f4cc4aa0f7ba436afd117efa9486b324b3f/packages/blocks/src/components/format-quick-bar/format-bar-node.ts#L82

doodlewind: It's just a workaround in `BaseBlockModel`, only the `flavour` field should be required. I don't want the default `type` field actually, which should have been commented 🤦‍♂️

PR #867, opened 2023-01-27
fix: update code type

lawvs:
Fix https://github.com/toeverything/blocksuite/pull/860#discussion_r1088186641

### DIFF
--- a/tests/format-quick-bar.spec.ts
+++ b/tests/format-quick-bar.spec.ts
@@ -413,7 +413,9 @@ test('should format quick bar be able to change to heading paragraph type', asyn
 
   const paragraphBtn = page.locator(`.paragraph-button`);
   await paragraphBtn.hover();
-  const h1Btn = page.locator(`.format-quick-bar [data-testid=h1]`);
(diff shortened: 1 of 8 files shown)


tests/format-quick-bar.spec.ts
```diff
@@ -556,6 +563,8 @@ async function scrollToBottom(page: Page) {
     .evaluate(node =>
       node.scrollTo({ left: 0, top: 1000, behavior: 'smooth' })
     );
+  // TODO switch to `scrollend`
+  // See https://developer.chrome.com/en/blog/scrollend-a-new-javascript-event/
```
himself65, line 567: Scrolled is a too latest feature. We need a polyfill if to do so

https://github.com/argyleink/scrollyfills/blob/main/src/scrollend.js

lawvs: This is a function for the e2e test only, we can use it when the test environment is ready.

doodlewind (approved):
LGTM

PR #868, opened 2023-01-27
refactor(store): use schema in `syncBlockProps`

Himself65:
Related: https://github.com/toeverything/blocksuite/issues/151

### DIFF
--- a/packages/store/src/__tests__/workspace.unit.spec.ts
+++ b/packages/store/src/__tests__/workspace.unit.spec.ts
@@ -96,6 +96,7 @@ describe.concurrent('addBlock', () => {
       '0': {
         'meta:tags': {},
         'meta:tagSchema': {},
+        'prop:title': '',
         'sys:children': [],
         'sys:flavour': 'affine:page',
         'sys:id': '0',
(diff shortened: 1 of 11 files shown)


packages/store/src/__tests__/workspace.unit.spec.ts
```diff
@@ -96,6 +96,7 @@ describe.concurrent('addBlock', () => {
       '0': {
         'meta:tags': {},
         'meta:tagSchema': {},
+        'prop:title': '',
```
himself65, line 99: we didn't set `page.title` as a default value; that's the reason the previous test works

packages/store/src/utils/utils.ts
```diff
@@ -41,72 +43,42 @@ export function initInternalProps(yBlock: YBlock, props: Partial<BlockProps>) {
...
-    // TODO compare with current yBlock value
-    if (props[key] !== undefined) {
-      yBlock.set('prop:' + key, props[key]);
+    if (Array.isArray(value)) {
+      yBlock.set(`prop:${key}`, Y.Array.from(value));
+    } else {
+      yBlock.set(`prop:${key}`, value);
+    }
+    if (!(key in defaultState)) {
+      console.warn(`unexpected props: ${key}`);
```
himself65: please check the warning output in https://blocksuite-git-20230127-schema-toeverything.vercel.app/?init

PR #869, opened 2023-01-27
feat: support `focusTitle` & skip container in `getNextBlock`

lawvs:
- Refactor `getNextBlock` and `getPreviousBlock`
  - Skip `affine:frame`/`affine:page`
  - Sunset `getSiblingsById`/`getNextSiblingById`/`getPreviousSiblingById`
- Add new utility function `focusTitle(offset)`
- Other misc changes

https://user-images.githubusercontent.com/18554747/215123880-295c82db-35db-4a03-a495-3d1a21108dce.mov


Himself65 (commented):
sounds good, but I think we shouldn't throw an error as it a expected behavior and we could control it

doodlewind (approved):
LGTM! We can remove the deprecated methods in new PRs.

PR #873, opened 2023-01-28
feat(blocks): add modal component

Himself65:
```tsx
document.querySelector('affine-default-page').components.modal.open = true
```


doodlewind:
Closing since we are awaiting new UX for database block, and this PR is now marked in the roadmap issue as reference, which should be adequate ATM.

PR #876, opened 2023-01-28
fix: keep native range on cross-frame dragging

fundon:
1. Add new frame to page
2. Select texts on cross-frame

https://user-images.githubusercontent.com/27926/215264597-a8bb0bcb-2dc4-4842-9ad6-bbb2b99e71fa.mov


doodlewind:
Thanks for the swift fix! The only concern is that we have piled a few tech debts here, so I'll try my best to make it maintainable.

fundon:
This is good, the readability has been enhanced.

pionxzh:
@fundon 
Reproduce steps:
1. select any text in the frame
2. select the title
3. error 😢 

https://user-images.githubusercontent.com/9910706/215281825-27afd196-21f1-47b0-b3ab-bce42d743526.mp4

fundon:
@pionxzh should determine if the selection is made in the header by checking the `startRange`.

PR #877, opened 2023-01-28
test: add BROWSER env config to local environment

doodlewind:
This PR brings the `BROWSER` config to non-CI environment, with documentation helping new contributors deal with compatibility issues.

/cc @fi3ework


fi3ework:
better naming than before 😎

PR #887, opened 2023-01-28
fix: should select text in title after selecting text in frame

fundon:
Related: https://github.com/toeverything/blocksuite/pull/876#issuecomment-1407446494


fundon:
@pionxzh I think it should be fixed.

Himself65 (approved):

PR #889, opened 2023-01-29
refactor: cross-frame selection

fundon:


doodlewind (approved):

packages/blocks/src/__internal__/utils/selection.ts
```diff
@@ -452,8 +452,8 @@ function handleCrossFrameDragMove(
 ) {
   let isFrame = container.tagName === 'AFFINE-FRAME';
 
-  if (isBackward && !isFrame) {
```
fundon: @doodlewind It would be better to keep the isFrame here.

doodlewind: Thanks for your feedback!

With looser condition we may have more DOM checks, but (subjectively) more semantically intuitive, since it looks a bit twisted to reassign a value that is `false` in `if` condition.


``` ts
// commonly we would suggest `a` is false inside this statement
if (!a) {
  if(xxx) {
    a = true
  }
}
```

A simpler mindset could be directly "okay, `a` will be true if blablabla".

Well it's opinionated after all. I would keep in mind reserving the original intention afterwards.

PR #892, opened 2023-01-29
chore: add benchmark package

Himself65:


doodlewind:
Opinionated closing this in favor of #1176, since I consider the binary format size and compatibility as our top metrics to keep in mind at the moment.

PR #901, opened 2023-01-30
feat: add dat-gui based CSS debug menu

thorseraq:
https://user-images.githubusercontent.com/20554850/215510465-af4b2116-921e-4d10-93ec-a3b924ac36e4.mov

### DIFF
--- a/packages/blocks/src/page-block/default/default-page-block.ts
+++ b/packages/blocks/src/page-block/default/default-page-block.ts
@@ -79,6 +79,10 @@ export class DefaultPageBlockComponent
       overflow-y: auto;
       height: 100%;
     }
+
+    .dg.ac {
+      z-index: 1001 !important;
+    }
     .affine-default-page-block-container {
       font-family: var(--affine-font-family);
       font-size: var(--affine-font-base);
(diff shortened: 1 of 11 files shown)


packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -79,6 +79,10 @@ export class DefaultPageBlockComponent
       overflow-y: auto;
       height: 100%;
     }
+
+    .dg.ac {
```
doodlewind, line 83: Can we move this style into `DebugMenu`?

thorseraq: The css scope in DebugMenu can't affect on dat-gui element🫠

thorseraq: There seems can be a way to work

doodlewind: You can inject the style globally from JavaScript side in `DebugMenu`, which can be also better than hard-coded here.

doodlewind (approved):
LGTM, thanks for bringing better debug DX!

Himself65 (approved):

PR #907, opened 2023-01-30
ci: test build affine

himself65:


doodlewind:
This may longer time than running the `pnpm test`?

himself65:
if we add cache, it would be ok

doodlewind:
I rethought a bit and this may be out of the scope of BlockSuite. Even if we support this, *building* AFFiNE successfully will not be enough to validate the quality of it, which requires improving CI in AFFiNE. We should work there by the moment.

PR #911, opened 2023-01-31
fix: block hub card list hover state & hovertip style

thorseraq:
1. Make card list disappear when hovering on close button

https://user-images.githubusercontent.com/20554850/215694186-bac5ce5e-7b9f-46be-ae11-4c9cd362bf6f.mov

2.  Make hover tips disappear when after starting dragging from blank line button

https://user-images.githubusercontent.com/20554850/215694590-80796b3e-b50e-4531-90a0-4a05d2bfcdaf.mov

### DIFF
--- a/packages/blocks/src/components/blockhub.ts
+++ b/packages/blocks/src/components/blockhub.ts
@@ -392,7 +392,12 @@ export class BlockHub extends NonShadowLitElement {
           affine-type="text"
         >
           ${RectIcon}
-          <tool-tip inert role="tooltip" tip-position="left" style="top: 5px"
+          <tool-tip
+            inert
+            role="tooltip"
+            tip-position="left"
+            style="top: 5px"
+            ?hidden=${!this._showToolTip}
             >Drag to Insert blank line
           </tool-tip>
         </div>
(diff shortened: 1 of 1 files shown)


packages/blocks/src/components/blockhub.ts
```diff
@@ -586,6 +602,11 @@ export class BlockHub extends NonShadowLitElement {
     this.requestUpdate();
   };
 
+  private _onBlockHubEntryMouseOver = () => {
+    this._isCardListVisible = false;
```
doodlewind, line 606: `_isCardListVisible` should be a `@state`, this avoids manual `requestUpdate`

doodlewind: Well I took a look, in this component we have too many non `@state` variables, so let's keep this as is but here may need further refactoring.

doodlewind (approved):

packages/blocks/src/components/blockhub.ts
thorseraq: There seems to be more variables that can be changed to `@state` variables to reduce the usage of `requestUpdate`👀

PR #915, opened 2023-01-31
feat(blocks): autofocus new block inserted from block hub

thorseraq:
https://user-images.githubusercontent.com/20554850/215693082-7e4db235-2838-41b8-824f-d2b0160a0e21.mov

### DIFF
--- a/packages/store/src/workspace/page.ts
+++ b/packages/store/src/workspace/page.ts
@@ -467,41 +467,27 @@ export class Page extends Space<PageData> {
     });
   }
 
-  @debug('CRUD')
-  insertBlock(
-    blockProps: Partial<BaseBlockModel>,
+  addSiblingBlock(
     targetModel: BaseBlockModel,
-    top = true,
-    autofocus = false
+    blockProps: Partial<BaseBlockModel>,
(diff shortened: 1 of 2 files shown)


packages/store/src/workspace/page.ts
```diff
@@ -471,28 +471,36 @@ export class Page extends Space<PageData> {
...
           type: blockProps.type,
         },
         targetParentModel.id,
         top ? targetIdx : targetIdx + 1
       );
-      // }
+    });
+    if (!autofocus) return;
+    // borrowed from asyncFocusRichText() in @blocksuite/blocks
+    requestAnimationFrame(() => {
```
doodlewind, line 500: We shouldn't do this kind of UI mutation in the `store` package. Can you try to move the `insertBlock` method to the `blocks` package?

doodlewind: Let me try to refactor here a bit.

thorseraq: Thanks！！I will take care next time

doodlewind (approved):
LGTM 👻

PR #918, opened 2023-01-31
feat: sync blob upload state by awareness

alt1o:
There are still some issues. In the previous design, blobStorage supported multiple providers. It's hard to figure out the upload state. 
So I will refactor blobStorage to only support one provider and the developer can custom blobStorage when initing workspace.


packages/blocks/src/embed-block/image/image-block.ts
```diff
@@ -118,15 +119,60 @@ export class ImageBlockComponent extends NonShadowLitElement {
...
     this.model.propsUpdated.on(() => this.requestUpdate());
     this.model.childrenUpdated.on(() => this.requestUpdate());
     // exclude padding and border width
     const { width, height } = this.model;
     const storage = await this.model.page.blobs;
     assertExists(storage);
+
+    // FIXME: We cannot guarantee the synchronization order of ydoc and awareness
+    // If ydoc update first, we cannot get upload state for awareness
+    await sleep(1000);
```
himself65: little weird here. I think we should use something like `event emitter` here

alt1o: Using `event emitter` is also wired. If we use `event emitter`, image block could not figure out when to load the data. Otherwise, image block shows a 404 state, then turns to an upload state.

alt1o:
e2e tests failed wired.

doodlewind (commented):
I tried and reproduced the CI failure locally, the markdown test case that may fail on macOS could be ignored for now.

doodlewind:
> I tried and reproduced the CI failure locally, the markdown test case that may fail on macOS could be ignored for now.

The markdown case is fixed in #929, merged here

packages/blocks/src/embed-block/image/image-block.ts
```diff
@@ -118,15 +121,63 @@ export class ImageBlockComponent extends NonShadowLitElement {
   @state()
   private _source!: string;
 
+  @state()
```
himself65:
```suggestion
```

this doesn't need to be a state as this won't appear in the render

packages/blocks/src/embed-block/image/image-block.ts
```diff
@@ -142,23 +193,43 @@ export class ImageBlockComponent extends NonShadowLitElement {
...
+          <div>${img}</div>
           ${childrenContainer}
         </div>
       </affine-embed>
     `;
   }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    this._imageReady.dispose();
```
himself65:
```suggestion
    this._imageReady.dispose();
    super.disconnectedCallback();
```

himself65 (changes requested):

himself65 (changes requested):
`_onContainerMouseMove` will broken

himself65:
> `_onContainerMouseMove` will broken


https://user-images.githubusercontent.com/14026360/216105602-a0b639a4-8d8c-4750-b621-fbd31a72a870.mp4

himself65 (approved):
LGTM, even though there're still some bugs. But it's ok for this PR to merge


https://user-images.githubusercontent.com/14026360/216402685-03b7a9bc-db9e-4bb4-81df-7bf48140e17f.mp4

himself65:
> LGTM, even though there're still some bugs. But it's ok for this PR to merge
> 
>  20230202-114955.mp4

For this. I think resizing a broken image should be prohibited.

alt1o:
> LGTM, even though there're still some bugs. But it's ok for this PR to merge
> 
>  20230202-114955.mp4

It's really a problem. I will fix it.

PR #921, opened 2023-02-01
fix: wrap update using transact

Flrande:


himself65 (approved):

himself65:
I hope there will have a test case soon

PR #923, opened 2023-02-01
feat: init edgeless toolbar

thorseraq:
A little change from #794, now treating all secondary toolbar as another component, and using popper.js to position it

- [x] change mouse mode and select shape
- [x] unit test
- [x] tooltip


packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
```diff
@@ -196,6 +200,29 @@ export class EdgelessPageBlockComponent
   }
 
   firstUpdated() {
+    if (this.page.awarenessStore.getFlag('enable_toolbar')) {
```
doodlewind, line 203: Could you make this part a method?

packages/playground/src/utils.ts
```diff
@@ -115,6 +115,7 @@ export function getOptions(): Pick<
       enable_drag_handle: true,
       enable_block_hub: true,
       enable_database: true,
+      enable_toolbar: true,
```
doodlewind, line 118: `enable_toolbar` -> `enable_edgeless_toolbar`

packages/blocks/src/page-block/edgeless/toolbar.ts
```diff
@@ -0,0 +1,386 @@
...
+import { createPopper } from '@popperjs/core';
+import {
+  createEvent,
+  MouseMode,
+  ShapeMouseMode,
+} from '../../__internal__/index.js';
+import { assertExists } from '@blocksuite/global/utils';
+
+@customElement('shape-component')
+export class ShapeComponent extends LitElement {
```
doodlewind, line 27: Consider rename to `ShapeMenu`? `ShapeComponent` may indicate shape elements or blocks.

thorseraq:
Naming and extracting method are updated at  https://github.com/toeverything/blocksuite/pull/923/commits/c448c0ce24d763b5a844442580e80c254b220d1b 

/cc @doodlewind

PR #928, opened 2023-02-01
feat: improve debugLog

Himself65:


packages/store/src/yjs/index.ts
```diff
@@ -22,4 +24,9 @@ export class BlockSuiteDoc<
     const map = super.getMap(key);
     return createYMapProxy(map, config);
   }
+
+  @debug('transact')
+  transact(f: (arg0: Transaction) => void, origin?: number) {
+    super.transact(f, origin);
+  }
```
himself65, line 31: image

packages/store/src/yjs/index.ts
```diff
@@ -22,4 +24,9 @@ export class BlockSuiteDoc<
     const map = super.getMap(key);
     return createYMapProxy(map, config);
   }
+
+  @debug('transact')
```
doodlewind, line 28: What's the difference between `transact` and `CRUD`?

himself65: Just a tag name. The reason is crdt functions always have a transact function. That would cause duplicate logs

PR #940, opened 2023-02-02
fix: block hub z-index

thorseraq:


packages/blocks/src/components/blockhub.ts
```diff
@@ -79,6 +79,11 @@ export class BlockHub extends NonShadowLitElement {
   private _delay = 200; // ms
 
   static styles = css`
+    affine-block-hub {
+      position: fixed;
```
himself65, line 83: not absolute?

thorseraq: absolute works as well

doodlewind (approved):

packages/blocks/src/components/blockhub.ts
himself65: please use absolute

PR #945, opened 2023-02-02
refactor(store): move text field into schema

Himself65:
Upstream: https://github.com/toeverything/blocksuite/pull/945


doodlewind:
@Saul-Mirone this is related to our previous discussion about remove hard-coded `text` field. Any idea?

Saul-Mirone:
> @Saul-Mirone this is related to our previous discussion about remove hard-coded `text` field. Any idea?

The solution in this PR looks good to me.

PR #946, opened 2023-02-02
refactor(blocks): use page-level block hub with `@state` pattern

### LINKED ISSUE #884: Should resize frame `xywh` when dragging new block from BlockHub

### LINKED ISSUE #938: Stale selection rect after inserting from BlockHub
After a block was selected by clicking the block handle and the light-blue selection rectangle showed up, this rect's position will not be correctly updated while inserting new block from the block hub by drag-and-dropping. 

Please see the attached screencast. At the 00:00:05 you will find a glitch: the z-index of selection rect was wrong as it appeared above the block hub menu.

https://user-images.githubusercontent.com/132558/216217505-72bf0583-3415-4ef0-8b71-15e3beedc9f2.mp4

However, inserting new block by clicking the edit area will not bring about this problem because the click action will firstly deselect any existed selection.

### LINKED ISSUE #939: Redundant block hub instances exist after editor disposed
https://user-images.githubusercontent.com/14026360/216226409-fce7089d-849c-44e8-bfb9-c92c84dfffa3.mp4

thorseraq:
- [x] extract blockhub init from selection manager to page block component
- [x] replace variables + `requestUpdate()` style with `@state()`
- [x] adapt height of blockhub when resizing
- [x] using popper.js, since there are some bug fixs, I want the merge happens a litter ahead, this will be launched in another PR. The idea is to extract second level menus to independent components

close #884
close #938 
close #939


packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -145,8 +147,10 @@ export class DefaultPageBlockComponent
    */
   components: {
     dragHandle: DragHandle | null;
+    blockHub: BlockHub | null;
```
himself65, line 150: good job

Himself65 (approved):

doodlewind:
Also fixes #1019

PR #947, opened 2023-02-02
fix: isAtLineEdge with format

lawvs:
In `isAtLineEdge`, use the length of the textContent is not accurate,
    since the text may be a part of a text node.

Consider the following case:

```html
    <text>long text
    |<bold>bold</bold> <- when the caret is at the start of bold, it's at the line edge but this function returns false
    </text>
```

And it will break ArrowUp behavior when the caret is at the start of the second line and follow the formatted text.

https://user-images.githubusercontent.com/18554747/216383011-fd868a6c-04b2-4c99-b7e8-a0c8e776cead.mov

### DIFF
--- a/packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
+++ b/packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
@@ -359,6 +359,124 @@ export function handleLineStartBackspace(page: Page, model: ExtendedModel) {
   );
 }
 
+function findTextNode(node: Node): Node | null {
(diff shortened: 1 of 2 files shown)


himself65:
https://user-images.githubusercontent.com/14026360/216664347-3caccffb-cf48-4ac8-b902-fb98f32e8344.mp4

Does `ArrowDown`'s behavior as expected?

lawvs:
> Does `ArrowDown`'s behavior as expected?

Yes, it can be reproduced on master. I'm working on it.

PR #951, opened 2023-02-02
fix: format bar position is unstable

lawvs:
https://user-images.githubusercontent.com/18554747/216400639-fcbc0a76-8b14-4d5e-9fd7-3ff1de6e36de.mov


himself65 (approved):

himself65:
BTW, I prefer to use popper.js in the future

PR #955, opened 2023-02-02
test(blocks): add test case on database block

himself65:


himself65:
I cannot measure the width of the frame block when it has a database block. Because the different system has their own logic on overflow div


/cc @doodlewind

Issue #935, opened 2023-02-02
drag handle regression in database block


Himself65:

doodlewind:
Closing this as not reproduced since drag handle in database block basically works well.

/cc @zqran

PR #957, opened 2023-02-02
fix: drag handle can not escape database block when moving mouse

fundon:
Related: #935


packages/blocks/src/components/drag-handle.ts
```diff
@@ -188,7 +188,7 @@ export class DragHandle extends LitElement {
       return;
     }
     const modelState = this._getBlockEditingStateByPosition(
-      this.getDropAllowedBlocks(this._startModelState?.model ?? null),
+      this.getDropAllowedBlocks(null),
```
fundon, line 191: @Himself65 Currently not in DnD

Himself65 (approved):

PR #980, opened 2023-02-03
fix: format bar position at line edge

lawvs:
Screenshot 2023-02-04 at 3 01 00 AM


lawvs:
The reason for this BUG is at the edge of the line, `range.getBoundingClientRect()` returns a wrong value.

And this problem also occurs when using popper.js, because it also depends on the value return from `getBoundingClientRect`.

doodlewind (approved):
LGTM

PR #985, opened 2023-02-03
feat(playground): use example list

Himself65:


doodlewind (changes requested):
I would suggest keep using React in standalone example pages, rather than the `main` code path.

doodlewind (approved):

Issue #933, opened 2023-02-01
Counter-intuitive select-all behavior


donaldxdonald:
## Expected Behavior
If I press Ctrl/Cmd + A more than three times, I expect all the blocks to still be selected.

## Current Behavior
Pressing Ctrl/Cmd + A three times seems to clear the selection. But when I then press Backspace, all the blocks are deleted.

## Record
![select_all](https://user-images.githubusercontent.com/29999387/216094684-a295b7dd-5230-453c-8a12-0734dc6873ed.gif)

## Solution?
Not sure if this is a feature or bug, but it seems to work for me after removing these codes.
https://github.com/toeverything/blocksuite/blob/892d2f1d6121d401908d44721d45569e20a55c4f/packages/blocks/src/page-block/default/selection-manager.ts#L613-L614

himself65:
I think this is the expected behavior, and I don't feel the counter-intuitive

pingren:
@Himself65 This is unexpected since it unselect all blocks, but the blocks can still be deleted by backspace.

So Either:
1. Press Cmd + A more than two times, all blocks still are selected, and can be deleted by backspace
Or:
2. Fix backspace deletion after all blocks unselected. 

I suggested the first one since it is the most popular editor behavior (Notion, Feishu, etc…)

doodlewind:
> @Himself65 This is unexpected since it unselect all blocks, but the blocks can still be deleted by backspace.
> 
> So Either:
> 
> 1. Press Cmd + A more than two times, all blocks still are selected, and can be deleted by backspace
>    Or:
> 2. Fix backspace deletion after all blocks unselected.
> 
> I suggested the first one since it is the most popular editor behavior (Notion, Feishu, etc…)

See https://github.com/toeverything/blocksuite/issues/839#issuecomment-1411742112 for the thought on product design.

/cc @joebeijing

joebeijing:
@doodlewind Vote for option 1

PR #993, opened 2023-02-04
fix: clear selection when press esc keydown

### LINKED ISSUE #990: Pressing escape doesn't cancel current selection state
Reproduce:

1. Open playground
2. single click anywhere
3. notice that the selection is following you
4. try to press escape key
5. notice it doesn't work as expected (to stop the selection process)

also:
1. make a selection by dragging over text
2. try to press escape key
3. notice it doesn't work as expected (to remove the selection)

Innei:


Innei:
Press ESC down.

https://user-images.githubusercontent.com/41265413/216769841-5e796dbd-b04c-4757-bc45-d20adf3fcd1b.mov

himself65:
```
Error: Unexpected console message: TypeError: Cannot read properties of undefined (reading 'emit')
```

lawvs:
After pressing ESC, pressing Backspace can still delete selected blocks, I think it's related to https://github.com/toeverything/blocksuite/issues/933

Innei:
> After pressing ESC, pressing Backspace can still delete selected blocks, I think it's related to #933

Fixed.

lawvs (approved):

Issue #969, opened 2023-02-03
Unexpected scrollbar when hovering to the bottom of the page


lawvs:
https://user-images.githubusercontent.com/18554747/216530699-20c028f6-7721-4ce2-85ff-5aa26d6cfeba.mov

lawvs:
Fixed in #997

PR #997, opened 2023-02-04
fix: should set overflow to hidden on body

fundon:
Related: #969


lawvs:
But why is it related to mouse movement? 🤔

fundon:
The `affine-drag-handle` element is inserted on the body. When hovering the last block, triggered scrollbar.

Himself65 (approved):

PR #1021, opened 2023-02-06
fix: keydown position in line edge and empty line

lawvs:
Pressing ArrowDown at empty will always focus on the start of the line, it's unexpected when there's a multi-level indent. And it also happens at the end of multiple line blocks.

https://user-images.githubusercontent.com/18554747/216912135-51fe4cd5-e02f-4b10-9307-496b2e11f7f5.mov


himself65:
Coverage: https://app.codecov.io/gh/toeverything/blocksuite/pull/1021

himself65 (changes requested):
Add a test case that full coverage

himself65 (approved):

himself65:
seems that code coverage doesn't change, but it's fine

PR #1025, opened 2023-02-06
fix: remove event listener with options

lawvs:
> Warning: If a listener is registered twice, one with the capture flag set and one without, you must remove each one separately. Removal of a capturing listener does not affect a non-capturing version of the same listener, and vice versa.

See https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener

### DIFF
--- a/packages/global/src/utils/signal.ts
+++ b/packages/global/src/utils/signal.ts
@@ -28,7 +28,11 @@ export class Signal<T = void> implements Disposable {
     (element as HTMLElement).addEventListener(eventName, handler, eventOptions);
     signal._disposables.push({
       dispose: () => {
-        (element as HTMLElement).removeEventListener(eventName, handler);
+        (element as HTMLElement).removeEventListener(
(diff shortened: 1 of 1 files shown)


lawvs:
The existing `Signal.fromEvent` API is confusing. It returns a `signal` and can be disposed by calling `signal.dispose`.

In fact, we use it this way, and it is wrong.

```ts
this._disposables.add(Signal.fromEvent(window, 'keydown').on(handler)`
// ---
// Only dispose the handler, not calling removeEventListener
this._disposables.dispose();

// The right way, but can not dispose multiple events handler at same time
const signal = Signal.fromEvent(window, 'keydown');
signal.on(handler);
// ---
signal.dispose();
```


https://github.com/toeverything/blocksuite/blob/17a62e3ae1f535efaacf08457b8136a943491424/packages/editor/src/components/editor-container.ts#L119-L132

/cc @doodlewind

PR #1027, opened 2023-02-06
refactor: expose blockHub in EditorContainer

thorseraq:
# Why

When blockHub is used as an embedded editor, the consumer of blockHub may need more flexible way of controlling it, as when to create and where to put it. For example, AFFiNE needs to put blockHub in a container created by AFFiNE, to support more flexible layout

# After
Now blockHub can be accessed by `editor.blockhub`, and can set position property `bottom`, `right` relative to  its container


lawvs:
I suggest exporting `createBlockHub` directly from blocksuite, then user can mount it manually.

thorseraq:
> I suggest exporting `createBlockHub` directly from blocksuite, then user can mount it manually.

Updated at: https://github.com/toeverything/blocksuite/pull/1027/commits/cd2fe4a0f4d71673c32aa73c8dce297ee2c5b09e

packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -471,14 +430,13 @@ export class DefaultPageBlockComponent
   override connectedCallback() {
     super.connectedCallback();
     this._initDragHandle();
-    this._initBlockHub();
+    // this._initBlockHub();
```
himself65: remove this line

PR #1029, opened 2023-02-06
test(global): improve test case for debuglog

himself65:


himself65:
https://app.codecov.io/gh/toeverything/blocksuite/pull/1029

PR #1043, opened 2023-02-07
fix: keydown at line edge

lawvs:
Fix caret position when pressing key down at line edge 

https://user-images.githubusercontent.com/18554747/217185250-9f3760fc-8a29-4916-a7c3-3310df4f99ae.mov

### DIFF
--- a/packages/blocks/src/__internal__/rich-text/shortcuts.ts
+++ b/packages/blocks/src/__internal__/rich-text/shortcuts.ts
@@ -24,7 +24,7 @@ export class Shortcuts {
   ): boolean {
     const selection = quill.getSelection();
     if (!selection) {
-      return PREVENT_DEFAULT;
+      return false;
     }
     const [line] = quill.getLine(selection.index);
     if (Shortcuts._isValid(prefix, line.domNode.tagName)) {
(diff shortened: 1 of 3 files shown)


packages/blocks/src/__internal__/rich-text/shortcuts.ts
```diff
@@ -24,7 +24,7 @@ export class Shortcuts {
   ): boolean {
     const selection = quill.getSelection();
     if (!selection) {
-      return PREVENT_DEFAULT;
+      return false;
```
himself65:
```suggestion
      return PREVENT_DEFAULT;
```

packages/blocks/src/__internal__/rich-text/shortcuts.ts
```diff
@@ -35,7 +35,7 @@ export class Shortcuts {
         }
       }
     }
-    return PREVENT_DEFAULT;
+    return false;
```
himself65:
```suggestion
    return PREVENT_DEFAULT;
```

Himself65 (changes requested):

Himself65 (approved):

Issue #1046, opened 2023-02-07
Inaccurate playground source map


doodlewind:

pengx17:
It looks like the issue will be gone after disabling istanbul. 
It seems the Istanbul plugin will always be enabled in the current implementation according to
https://github.com/iFaxity/vite-plugin-istanbul/blob/next/src/index.ts#L141-L145

PR #1047, opened 2023-02-07
fix: local dev sourcemap issue

pengx17:
fix https://github.com/toeverything/blocksuite/issues/1046

### DIFF
--- a/packages/playground/vite.config.ts
+++ b/packages/playground/vite.config.ts
@@ -5,17 +5,20 @@ import { fileURLToPath } from 'url';
 import { hmrPlugin } from './scripts/hmr-plugin';
 import istanbul from 'vite-plugin-istanbul';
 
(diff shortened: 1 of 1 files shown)


packages/playground/vite.config.ts
```diff
@@ -5,17 +5,20 @@ import { fileURLToPath } from 'url';
...
   plugins: [
     react(),
     hmrPlugin,
-    istanbul({
-      cwd: fileURLToPath(new URL('../..', import.meta.url)),
-      include: ['packages/**/src/*'],
-      exclude: ['node_modules', 'tests'],
-      forceBuildInstrument: !!process.env.CI || !!process.env.COVERAGE,
-    }),
+    enableIstanbul &&
```
doodlewind, line 15: Will this pass in a `false` to `plugins` array?

pengx17: yeah, I think vite config can accept false values in `plugins` field.

PR #1050, opened 2023-02-07
build: enable HMR by default

Innei:
refer #1044

Signed-off-by: Innei <tukon479@gmail.com>


himself65:
sometimes I saw the import `@blocksuite/store` package twice when using vite. Not sure what's the reason

Innei:
> sometimes I saw the import `@blocksuite/store` package twice when using vite. Not sure what's the reason

What's mean

PR #1061, opened 2023-02-07
style: sort import

himself65:


himself65:
This would affect all downstream PRs

Himself65:
> We should use eslint import-sort or vscode organize imports settings
> 
> ```
> // .vscode/settings.json
> "editor.codeActionsOnSave": ["source.organizeImports"],
> ```

Both

himself65:
I think enforcing people to sort the imports is better. You cannot assume what IDE people are using

PR #1096, opened 2023-02-09
fix: align content and title

QiShaoXuan:


lawvs (changes requested):
You should update config at here
https://github.com/toeverything/blocksuite/blob/0f742df3cfb8be7172d87f512eedad99f2a474db/packages/global/src/config/consts.ts#L28

PR #1098, opened 2023-02-09
chore: update slash menu

lawvs:


packages/blocks/src/page-block/default/utils.ts
```diff
@@ -312,6 +313,40 @@ export async function copyImage(model: EmbedBlockModel) {
...
+      },
+    ];
+  }
+  return model.text.toDelta();
+}
+
+export async function copyBlock(model: BaseBlockModel) {
+  const copyType = 'blocksuite/x-c+w';
+  const delta = getTextDelta(model);
+  const copyData = {
```
himself65: what is the type of `copyData`

lawvs: Update type to `OpenBlockInfo`

Himself65 (approved):

PR #1104, opened 2023-02-10
feat(blocks): select and drag shape

alt1o:
https://user-images.githubusercontent.com/21084335/217969591-20069250-b873-4088-b850-eddb48aa40e9.mov


doodlewind (approved):
Test works now, thanks!

PR #1129, opened 2023-02-10
refactor: use `@blocksuite/icons`

Himself65:


packages/global/package.json
```diff
@@ -35,6 +35,7 @@
     ]
   },
   "dependencies": {
+    "@blocksuite/icons": "=2.1.0-alpha.7",
```
doodlewind, line 38: Where is the upstream repository? I am concerned if we hold private dependencies.

doodlewind (changes requested):
We should not hold this private dependency, unless we open source the icon repo and make it publicly accessible.

packages/global/package.json
himself65: https://github.com/toeverything/icons

doodlewind:
Opinionated closing this since we don't plan to make to icon workflow public recently.

PR #1138, opened 2023-02-11
fix: should use clientX/clientY to create startRange in edgeless mode

fundon:
Fixes: https://github.com/toeverything/AFFiNE/issues/926


Himself65 (approved):
I test locally. perfect

PR #1142, opened 2023-02-12
fix: await assert

lawvs:
Fix test for https://github.com/toeverything/blocksuite/pull/1122

/cc @Innei


doodlewind:
Fixed in #1104 (https://github.com/toeverything/blocksuite/pull/1104/files#diff-87d5f9f9df663eeb0751d9b7c18771b7683a16a62e77aa1678dd36c81ef2b97cR251-R262)

PR #1144, opened 2023-02-12
fix: copy button of code block menu doesn't work

### LINKED ISSUE #1128: Code block copy does not function as expected
1) Copy paste to the right of the code block I would expect to copy the whole block - but it does not function

2) When copying content from inside the code block it will paste as a code block - I would expect this to only copy the data/string.

Image

### LINKED ISSUE #1145: Copy button of code block menu doesn't work
https://user-images.githubusercontent.com/20554850/218304957-f77a429e-5915-4ae1-b5d3-5ca4dc0a4c4e.mov

qinluhe:
closes https://github.com/toeverything/blocksuite/issues/1145

### DIFF
--- a/packages/blocks/src/page-block/default/utils.ts
+++ b/packages/blocks/src/page-block/default/utils.ts
@@ -12,6 +12,7 @@ import {
   doesInSamePath,
   getBlockById,
   getBlockElementByModel,
+  getCurrentRange,
   getRichTextByModel,
   OpenBlockInfo,
   resetNativeSelection,
@@ -408,8 +409,13 @@ export function copyCode(codeBlockOption: CodeBlockOption) {
   assertExists(richText);
   const quill = richText.quill;
   quill.setSelection(0, quill.getLength());
-  document.dispatchEvent(new ClipboardEvent('copy'));
-  resetNativeSelection(null);
+  document.body.dispatchEvent(new ClipboardEvent('copy', { bubbles: true }));
+
+  const range = getCurrentRange();
+  range.setStart(richText, 0);
+  range.setEnd(richText, 0);
+  resetNativeSelection(range);
+
   toast('Copied to clipboard');
 }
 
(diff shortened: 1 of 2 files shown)


packages/editor/src/managers/clipboard/copy-cut-manager.ts
```diff
@@ -216,6 +214,8 @@ export class CopyCutManager {
     } finally {
       tempElem.removeEventListener('copy', listener);
       document.body.removeChild(tempElem);
+
+      const curRange = getCurrentRange();
```
lawvs, line 218: Please revert this change, it will break copy action in the format bar (selection lost)


https://user-images.githubusercontent.com/18554747/218300412-d4dea242-c77d-4433-97ed-b3a6f93a50b8.mov

thorseraq:
We need further investigation for #1128, key point is `_getClipItems()` during copy, and `_clipboardData2Blocks()` during paste. For example, now hit copy button and paste it at a line end of paragraph, it won't append a new code block.

But it fixes #1145, could you add a unit test in `tests/code.spec.ts`?

qinluhe:
> We need further investigation for #1128, key point is `_getClipItems()` during copy, and `_clipboardData2Blocks()` during paste. For example, now hit copy button of code block and paste it at a line end of paragraph, it won't append a new code block.
> 
> But It fixes #1145, could you add a unit test in `tests/code.spec.ts`?

1. There are two different reasons for the problem that clicking the copy button does not meet expectations, one of which is that the copy event is not passed to handleCopy correctly, this PR fixes the problem for this reason

2. 👌🏻

thorseraq:
@qinluhe I have added the unit test, thanks for contributing! 

we have added the CLA process in BlockSuite. Would you mind signing it by editing on https://github.com/toeverything/blocksuite/edit/master/.github/CLA.md ? Thanks!

himself65 (approved):

himself65:
need to sign up CLA before merging

qinluhe:
> @qinluhe I've added the unit test, thanks for contributing!
> 
> we have added the CLA process in BlockSuite. Would you mind signing it by editing on https://github.com/toeverything/blocksuite/edit/master/.github/CLA.md ? Thanks!
👌🏻

PR #1146, opened 2023-02-12
refactor(store): remove PrelimText

### LINKED ISSUE #1141: Remove `PrelimText` and `page.markTextSplit`
For now, `PrelimText` exists for handling hard split (in `handleBlockSplit`) when user presses enter at the middle of a text block. Example:

``` ts
// current block before split
'123|456'

// after split
'123' // current block
'456' // new block
```

So the mental model of splitting is:

1. Only keep the left half of current block, delete its right part.
2. Add a new block, with the right half of the content.

> See https://github.com/toeverything/blocksuite/discussions/554 for the reason behind this.

But due to the feature of Yjs (https://github.com/yjs/yjs/issues/207), if a `yText.clone()` cloned instance is not inserted to YDoc yet, at this moment, if the original instance got mutated, the cloned instance will also be affected. So `PrelimText` is designed to mock the cloning process.

However, this may be alleviated by copying by the *delta* data in this way:

1. Export the delta representation of current block text, split it to left part and right part.
2. Mutate current block, delete its right part.
3. Insert a new block with the right part as its content.

To make this happen, we need to support the `split` API at delta level.

/cc @Flrande

Flrande:
Closes #1141

### DIFF
--- a/packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
+++ b/packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
@@ -92,10 +92,8 @@ export function handleBlockSplit(
   const parent = page.getParent(model);
   if (!parent) return;
 
-  const [left, right] = model.text.split(splitIndex, splitLength);
   page.captureSync();
-  page.markTextSplit(model.text, left, right);
-  page.updateBlock(model, { text: left });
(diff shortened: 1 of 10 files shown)


doodlewind (approved):
LGTM, a good step closer to fully manage our rich-text editing experience!

PR #1152, opened 2023-02-13
feat(blocks): optimize clipboard support for code block

### LINKED ISSUE #1128: Code block copy does not function as expected

thorseraq:
This PR updates the clipboard support for code block:

* Use menu copy button
  * When paste inside the code block, perceptually copying only the text
  * When paste outside the block, perceptually copying the block and inserting a new block after the current block
* Use cv shortcut keys
  * Behavior inside and outside the code block are the same, only copy the text

close #1128

https://user-images.githubusercontent.com/20554850/218432617-b4683a0f-078d-472b-a610-3531e144cc4c.mov

### DIFF
--- a/packages/blocks/src/__internal__/utils/types.ts
+++ b/packages/blocks/src/__internal__/utils/types.ts
@@ -92,6 +92,11 @@ export type OpenBlockInfo = {
     retain?: number;
     attributes?: TextAttributes;
   }[];
+  rawText?: {
+    insert: string;
+    delete?: number;
+    retain?: number;
+  }[];
   checked?: boolean;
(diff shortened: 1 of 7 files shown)


Himself65 (changes requested):
https://user-images.githubusercontent.com/14026360/218493239-3ea58bb3-5051-439c-9676-678b910e3afa.mp4

thorseraq:
ooooooo

Himself65 (approved):

PR #1157, opened 2023-02-13
feat(playground): support guided tour

Himself65:


doodlewind:
Opinionated closing this since the mask from Shepherd leads to a blink during the initialization of the playground page. Also, the style of the modal box is a bit off. IMO this feature fits better for AFFiNE.

Issue #1159, opened 2023-02-13
deletion through image block


Himself65:
https://user-images.githubusercontent.com/14026360/218600588-390cbabb-b085-4f7c-ae20-afc9023c04fb.mp4

fundon:
Fixed.

PR #1164, opened 2023-02-14
fix: support deleting native-range-selected image block

fundon:
Fixes: #1159

### DIFF
--- a/tests/selection.spec.ts
+++ b/tests/selection.spec.ts
@@ -2086,3 +2086,77 @@ test('should keep native range selection when scrolling forward with the scroll 
   await copyByKeyboard(page);
   await assertClipItems(page, 'text/plain', '123456789');
 });
+
+// #1159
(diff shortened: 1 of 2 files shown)


tests/selection.spec.ts
```diff
@@ -1944,3 +1944,77 @@ test('should not be misaligned when the editor container has padding or margin',
...
+// #1159
+test('should be removed when the embed block is selected in native selection', async ({
+  page,
+}) => {
+  const room = await enterPlaygroundRoom(page);
+  const { pageId, frameId } = await initEmptyParagraphState(page);
+  await initThreeParagraphs(page);
+
+  const MOCK_IMAGE_ID = '_e2e_test_image_id_';
+  await page.route(
```
himself65, line 2099: Somehow we should have a mock server for this. I will try to do this later

fundon: Yes, the embed tests can be refactored.

PR #1168, opened 2023-02-14
refactor(blocks): replace textarea in title with virgo

Flrande:


packages/blocks/package.json
```diff
@@ -12,7 +12,8 @@
   "author": "toeverything",
   "license": "MPL-2.0",
   "peerDependencies": {
-    "@blocksuite/store": "workspace:*"
+    "@blocksuite/store": "workspace:*",
+    "yjs": "^13"
```
himself65: NO. please use Workspace.Y

packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -7,9 +7,8 @@ import {
   DisposableGroup,
   Page,
   Signal,
-  Text,
 } from '@blocksuite/store';
-import autosize from 'autosize';
```
lawvs, line 12: Since `autosize` is not used, you should remove it from package.json

lawvs (commented):
Impressive work, thanks!

lawvs (commented):
There are some console errors that need attention.

Press `cmd+b/e/i` or `ArrowDown` in the title.

doodlewind (commented):
🎉🎉🎉

doodlewind (commented):
I think we need to have compatibility test for this PR. Could you checkout the `migrations.ts` and add a YDoc binary generated with the `string` title type (for example, dumping from current playground)?

For backward compatibility, we need to make sure we can keep editing old data format.

doodlewind (commented):
We have just released AFFiNE Downhills, so next week there may be some hotfix on BlockSuite. In this case I would like hold the landing of this PR for a few days, making sure BlockSuite nightly versions not to have break changes during this period. No worries!

packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -127,6 +127,12 @@ export class DefaultPageBlockComponent
     .affine-default-page-block-title-container {
       margin-top: 78px;
     }
+
+    // This css is used to make text in title line break automatically,
+    // but in Virgo there is still just one line so the link break from this css is just visual.
```
lawvs, line 135: Double slash not work in the CSS

packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -127,6 +127,12 @@ export class DefaultPageBlockComponent
...
+    .affine-default-page-block-title virgo-unit-text > span {
+      white-space: pre-line !important;
```
lawvs, line 137: And I think you can remove the `!important` after fix CSS comment

Flrande: There are inline styles in `span` of `virgo-unit-text` so I use `!important` cover it.

packages/blocks/src/page-block/default/default-page-block.ts
Flrande: I will fix it.

packages/store/src/utils/utils.ts
```diff
@@ -79,6 +79,10 @@ export function syncBlockProps(
       } else if (Array.isArray(value)) {
         yBlock.set(`prop:${key}`, Y.Array.from(value));
       } else {
+        if (key === 'title' && value instanceof Text) {
```
Saul-Mirone, line 82: I don't think we need this since we've already handled `Text` in line 78.

packages/store/src/workspace/page.ts
```diff
@@ -813,9 +808,14 @@ export class Page extends Space<PageData> {
     if (model.flavour === 'affine:page') {
```
Saul-Mirone, line 808: We can change this to a type guard to avoid using `as` below:

```typescript
const isPageModel = (mode: BaseBlockModel): model is BlockSuiteInternal.BlockModels['affine:page'] => mode.flavour === 'affine:page'
```

Flrande: In fact I hava tried but  type guard not work as expected.

https://github.com/toeverything/blocksuite/blob/master/packages/global/src/utils.ts#L32

Saul-Mirone: The type declaration is not correct here, try something like:
```ts
type _K = keyof BlockSuiteInternal.BlockModels
type TupleToUnion<T extends unknown[]> = T[number];

export function matchFlavours<
  K extends _K,
  Key extends _K[] = _K[]
>(
  model: { flavour: K },
  expected: Key
): model is Pick<BlockSuiteInternal.BlockModels, TupleToUnion<Key>> & { flavour: K } {
  return expected.includes(model.flavour);
}
```

Flrande: It seems not work. 😢

image


image

Saul-Mirone: Forget it, I'll try to figure it out.

packages/store/src/__tests__/migration.unit.spec.ts
```diff
@@ -73,4 +73,17 @@ describe('migration', () => {
     expect(hasSurface).toBe(true);
     expect(hasShape).toBe(false);
   });
+
+  test('migrate to new page title', async () => {
+    const doc = await loadBinary('legacy-page-title');
+
+    assert.equal(
```
doodlewind, line 80: This may not reflect the potential breaking change, hope for investigation @Flrande

packages/store/src/workspace/migrations.ts
```diff
@@ -88,6 +88,46 @@ const migrations: Migration[] = [
...
+        .getMap('space:meta')
+        .get('versions') as Y.Map<number>;
+      yVersions.delete('affine:page');
+      yVersions.set('affine:page', 2);
+
+      for (const pageId of pageIds) {
+        const spaceId = `space:${pageId}`;
+        const yBlocks = doc.getMap(spaceId);
+        // @ts-ignore
+        yBlocks.forEach((yBlock: Y.Map<unknown>) => {
```
doodlewind, line 121: Can we early return here?

doodlewind (approved):
LGTM, cheers!

PR #1178, opened 2023-02-15
ci: use vite preview

himself65:
image


doodlewind:
How does this work? Does this mean for each time entering the playground page, the bundle built by `vite preview` will be pre-packaged?

himself65:
> How does this work? Does this mean for each time entering the playground page, the bundle built by `vite preview` will be pre-packaged?

yep, playground would output to a single zip, so that you can download to our local

himself65:
> > How does this work? Does this mean for each time entering the playground page, the bundle built by `vite preview` will be pre-packaged?
> 
> yep, playground would output to a single zip, so that you can download to our local

also this would speed up the e2e test because vite build output is production version

doodlewind:
Got it (https://stackoverflow.com/questions/71703933/what-is-the-difference-between-vite-and-vite-preview), this makes sense.

.github/workflows/cancel.yml
```diff
@@ -13,5 +13,5 @@ jobs:
     steps:
       - uses: styfle/cancel-workflow-action@0.11.0
         with:
-          workflow_id: 46388772, 34612818
```
himself65, line 16: see https://api.github.com/repos/toeverything/blocksuite/actions/workflows

.github/workflows/build-and-test.yaml
```diff
@@ -21,10 +21,51 @@ on:
       - 'packages/react/examples/**'
 
 jobs:
-  main-test:
```
himself65, line 24: move build and test into a single file

doodlewind (approved):

Issue #1186, opened 2023-02-15
Support image upload entry in block hub


doodlewind:
Upstream: https://github.com/toeverything/AFFiNE/issues/837

PR #1188, opened 2023-02-15
feat: support image upload from local file system

thorseraq:
This PR:
- added image support for blockHub
- exposed `async uploadImageFromLocal(editor: EditorContainer)` util for image uploading from local file system 

close https://github.com/toeverything/blocksuite/issues/1186.

### DIFF
--- a/packages/blocks/src/components/blockhub.ts
+++ b/packages/blocks/src/components/blockhub.ts
@@ -1,9 +1,11 @@
 import {
+  BLOCKHUB_FILE_ITEMS,
   BLOCKHUB_LIST_ITEMS,
   BLOCKHUB_TEXT_ITEMS,
   BlockHubIcon,
   CrossIcon,
   DatabaseTableViewIcon,
+  ImageIcon,
   NumberedListIconLarge,
   RectIcon,
   TextIconLarge,
@@ -43,7 +45,7 @@ type BlockHubItem = {
   toolTip: string;
 };
 
-export type CardListType = 'blank' | 'list' | 'text' | 'database';
+export type CardListType = 'blank' | 'list' | 'text' | 'database' | 'file';
 
 @customElement('affine-block-hub')
 export class BlockHub extends NonShadowLitElement {
@@ -101,7 +103,10 @@ export class BlockHub extends NonShadowLitElement {
   @query('[role="menu-entry"]')
   private _blockHubMenuEntry!: HTMLElement;
 
-  private _onDropCallback: (e: DragEvent, lastModelState: EditingState) => void;
+  private readonly _onDropCallback: (
+    e: DragEvent,
+    lastModelState: EditingState
+  ) => Promise<void>;
   private _currentPageX = 0;
   private _currentPageY = 0;
   private _indicator!: DragIndicator;
@@ -180,6 +185,11 @@ export class BlockHub extends NonShadowLitElement {
       align-items: center;
       position: absolute;
       right: 12px;
+    }
+
+    .card-icon-container > svg {
+      width: 20px;
+      height: 20px;
     }
 
     .card-container:hover {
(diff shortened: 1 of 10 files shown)


packages/blocks/src/components/blockhub.ts
```diff
@@ -43,7 +45,7 @@ type BlockHubItem = {
   toolTip: string;
 };
 
-export type CardListType = 'blank' | 'list' | 'text' | 'database';
+export type CardListType = 'blank' | 'list' | 'text' | 'database' | 'file';
```
lawvs, line 48: Useless update

lawvs (commented):
I don't think the motivation behind adding `addBlocksByFlavour` and `addSiblingBlocks` is significant, so I don't believe they need to be included.

packages/store/src/workspace/page.ts
```diff
@@ -461,28 +492,44 @@ export class Page extends Space<PageData> {
     });
   }
 
-  addSiblingBlock(
+  addSiblingBlocks(
     targetModel: BaseBlockModel,
-    blockProps: Partial<BaseBlockModel>,
+    props: Partial<BaseBlockModel> | Array<Partial<BaseBlockModel>>,
```
doodlewind, line 497: The type here should not be mixed IMO.

packages/store/src/workspace/page.ts
```diff
@@ -313,6 +313,37 @@ export class Page extends Space<PageData> {
...
+    const ids: string[] = [];
+    blocks.forEach(block => {
+      const id = this.addBlockByFlavour<ALLProps, Flavour>(
+        block.flavour,
+        block.blockProps ?? {},
+        parent,
+        parentIndex
+      );
+      ids.push(id);
+      typeof parentIndex === 'number' && parentIndex++;
```
doodlewind, line 341: If no `parentIndex` is specified, will the subsequent blocks always being added to the first position?

doodlewind: If you are going to add new API in `Page`, I think the test case in `workspace.unit.spec.ts` should be provided altogether.

packages/blocks/src/components/blockhub.ts
thorseraq: It's used, can you explain?

packages/store/src/workspace/page.ts
```diff
@@ -461,28 +492,44 @@ export class Page extends Space<PageData> {
...
   }
 
-  addSiblingBlock(
+  addSiblingBlocks(
     targetModel: BaseBlockModel,
-    blockProps: Partial<BaseBlockModel>,
+    props: Partial<BaseBlockModel> | Array<Partial<BaseBlockModel>>,
     direction: 'left' | 'right' = 'right'
-  ) {
+  ): string | string[] {
```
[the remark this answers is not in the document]
thorseraq: Updated at https://github.com/toeverything/blocksuite/pull/1188/commits/61967baa696e1136b59eea31dcf7bea499b3a54f

packages/store/src/workspace/page.ts
thorseraq: updated at https://github.com/toeverything/blocksuite/pull/1188/commits/61967baa696e1136b59eea31dcf7bea499b3a54f

packages/store/src/workspace/page.ts
thorseraq: >If you are going to add new API in Page, I think the test case in workspace.unit.spec.ts should be provided altogether.

working on it

Subsequent blocks will be added at the end page of top level (`const parentId = parent === null ? null : parent?.id ?? this.root?.id;`) in order (`const index = parentIndex ?? yChildren.length;`),

thorseraq: Updated at https://github.com/toeverything/blocksuite/pull/1188/commits/111120e449eed470bcb1b70809f491ec55d5a637

doodlewind (approved):

PR #1229, opened 2023-02-16
fix(blocks): use requestAnimationFrame when restoring selection

Flrande:


packages/blocks/src/page-block/utils/container-operations.ts
```diff
@@ -295,7 +297,9 @@ function formatModelsByRange(
       [key]: !isFormatActive,
     });
   }
-  restoreSelection(selectedBlocks);
+  requestAnimationFrame(() => {
+    restoreSelection(selectedBlocks);
+  });
```
lawvs, line 302:
```suggestion
  restoreSelection(selectedBlocks);
```

Flrande:
This pr need to wait for #1168, which changes `asyncFocusRichText()`.

Flrande:
I will change it in #1168, so let us close it.

PR #1230, opened 2023-02-16
feat: add `@blocksuite/vue` package

Innei:
Signed-off-by: Innei <tukon479@gmail.com>


packages/vue/.gitignore
```diff
@@ -0,0 +1,28 @@
+# Logs
```
himself65: move this to the top

doodlewind:
Hi @Innei, sorry for the delay and thanks for the contribution!

I am a big fan of Vue and I really like the idea how you wrap BlockSuite editor using the composition API. The code looks pretty neat! However, according to our recent progress (in refactoring AFFiNE), we are considering deprecate the `react` package, having the Next.js integration example maintained in standalone repo, since this allows better CI checks. So similarly in this case, it's hard to justify adding a `vue` package at the moment. Also, our team may lack the resource to review and a maintain a package that will not be used in AFFiNE in a short term.

As an alternative, how about maintain this vue component in your repo? It would be great fit to be shown in our README.

So I am closing this, but look forward to hearing your thoughts!

PR #1226, opened 2023-02-16
fix: disable shortcut "h" to shown/hidden dat.gui css debug menu

thorseraq:
close: #1227

reference: https://github.com/dataarts/dat.gui/blob/master/API.md#:~:text=If%20true%2C%20GUI,keypress.

PR #1232, opened 2023-02-16
ci: ignore draft pull request

thorseraq:


doodlewind:
This config is in https://github.com/toeverything/blocksuite/blob/master/.github/auto_assign.yml#L10. But since we have fixed the CI issue in bf8def06f2840de9b2796c35b6b94947e0754bbd (see https://github.com/toeverything/blocksuite/pull/1226#issuecomment-1432652525 also), it's expected for us to add draft PR into projects (no matter if it comes from forks) as early as possible. Our goal is to track organic development process using the GitHub project kanban (without disturbing our daily accustomed workflow), so having activities naturally tracked in the AFFiNE Kanban is the expected behavior.

thorseraq:
Got it, Make sense to me as well

PR #1256, opened 2023-02-17
fix: cannot locate insert position through dragging from block hub when page has very many blocks

thorseraq:
# Before

https://user-images.githubusercontent.com/20554850/219600665-4fae13b1-5828-4561-b305-f1d9ff3a43d4.mov

# After

https://user-images.githubusercontent.com/20554850/219600680-ac7a1095-32b1-4529-9c78-89ed4450b771.mov

### DIFF
--- a/packages/blocks/src/components/blockhub.ts
+++ b/packages/blocks/src/components/blockhub.ts
@@ -683,6 +683,24 @@ export class BlockHub extends NonShadowLitElement {
       this._currentPageX = e.pageX;
       this._currentPageY = e.pageY;
     }
+
+    this._refreshCursor(e);
+  };
+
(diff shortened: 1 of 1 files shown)


packages/blocks/src/components/blockhub.ts
```diff
@@ -711,6 +729,7 @@ export class BlockHub extends NonShadowLitElement {
           skipX: true,
         });
 
+    console.log(modelState);
```
doodlewind, line 732: remove here

thorseraq: 👀 updated

doodlewind (approved):

PR #1272, opened 2023-02-18
refactor: enable editable for page

lawvs:
Fix cursor can not move between block


doodlewind:
Closing since we have landed alternative workaround

PR #1281, opened 2023-02-19
refactor: add `text` prop in schema

Saul-Mirone:
Add `text` prop in schema to avoid hardcode validation for `prop:text`.
We can use that to decide whether a block has text content now.

This PR is based on https://github.com/toeverything/blocksuite/pull/945 by @Himself65 ❤️ .


himself65:
CI broken today, ignore it

himself65 (approved):

doodlewind:
@Saul-Mirone thanks for your contribution, welcome!

Issue #1248, opened 2023-02-16
Hovering on shape should not have effect on underlying block


doodlewind:
When hovering at following position (inside a shape), the underlying block is also set to hover state. This is not expected because we assume all the shape elements are located above the blocks.

<img width="878" alt="image" src="https://user-images.githubusercontent.com/7312949/219503919-aa88a5e3-0193-4be9-82fa-c40bcc6a95ff.png">

This is explicitly controlled in our codebase and could be easily fixed by filtering out blocks if the hovering position hits shape (see `onContainerMouseMove` in `default-mode.ts`). We have already supported this similarly in #1104.

Issue #1282, opened 2023-02-19
Support hover state for shape element


doodlewind:
Upstream: #1268

Currently we only support the "active" (selected) state of ShapeElement, but the hover state is expected to be supported altogether, just like what will happen when you hover over FrameBlock.

Related to #1248

doodlewind:
Closed by #1284

PR #1284, opened 2023-02-19
feat: support hover state on shape elements

zqran:
closes: #1248, #1282


doodlewind:
This looks good, I refactored a bit thought, thanks!

PR #1299, opened 2023-02-20
fix(phasor): adapt to resized canvas

### LINKED ISSUE #874: Incorrect page and canvas viewport after resizing the playground page
This is similar with #330, but affects the page container size.

### LINKED ISSUE #1220: Wrong edgeless shape offset on different `devicePixelRatio`
After switching browser `devicePixelRatio` (pressing `cmd` + `+/-` in Chrome), there is an offset between mouse position and the visual shape position when adding shape by drag.

alt1o:
Fixes #1220 
Fixes #874 

https://user-images.githubusercontent.com/21084335/220062787-a8816ba9-3213-4168-b877-b9e9f8d8a154.mov

### DIFF
--- a/packages/phasor/src/renderer.ts
+++ b/packages/phasor/src/renderer.ts
@@ -1,3 +1,6 @@
+import type { Disposable } from '@blocksuite/global/utils';
+import { Signal } from '@blocksuite/global/utils';
+
 import { IBound, MIN_ZOOM } from './consts.js';
 import type { PhasorElement } from './elements/index.js';
 import { GridManager } from './grid.js';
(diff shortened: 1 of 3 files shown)


alt1o:
fix: #1220

packages/phasor/src/renderer.ts
```diff
@@ -8,32 +11,48 @@ export class Renderer {
...
-    this.height = bbox.height;
-    this._centerX = this.width / 2;
-    this._centerY = this.height / 2;
+
+    this._initSize();
+    this.setCenter(this.width / 2, this.height / 2);
 
     this._loop();
+
+    this._disposeResizeListener = Signal.disposableListener(
```
doodlewind, line 33: Is it possible to listen on the resizing of the canvas element itself? This could be more precise if possible.

alt1o: I want to use window resizing event handing devicePixelRatio changing. But It seems also having some issues. I will change it to ResizeObserver.

doodlewind (approved):
LGTM

Issue #1214, opened 2023-02-07
Image block popup doesn't follow scroll


joebeijing:
The expected effect is that the menu is at best aligned top-bottom with the border.

![Image](https://user-images.githubusercontent.com/22443345/217264079-db020266-9558-429d-a2ca-921c705bf5d3.png)

joebeijing:
I can reproduce this issue in the test environment.
![Image](https://user-images.githubusercontent.com/22443345/220814824-a637342b-d752-42ae-b9d6-868892dbe399.png)

doodlewind:
Not reproduced in latest playground.

PR #1300, opened 2023-02-20
fix: popup menu should follow position of image

fundon:
Fixes: #1214


tests/embed.spec.ts
```diff
@@ -281,3 +282,76 @@ test('image get message from awareness', async ({ page, browser }) => {
...
+    if (!menu) {
+      throw new Error();
+    }
+    return [
+      image.getBoundingClientRect(),
+      menu.getBoundingClientRect(),
+    ] as const;
+  });
+
+  // < 1.059180567624251
```
himself65, line 355: what is this

doodlewind (approved):

tests/embed.spec.ts
fundon: Math.abs(a - b) < 1.x

PR #1310, opened 2023-02-21
fix: should switch to native range selection when dbclicking the text

fundon:


doodlewind (approved):

tests/selection.spec.ts
```diff
@@ -2086,3 +2086,38 @@ test('should keep native range selection when scrolling forward with the scroll
...
+  await initEmptyParagraphState(page);
+  await focusRichText(page);
+  await type(page, 'hello block suite');
+  await assertRichTexts(page, ['hello block suite']);
+
+  const helloPosition = await page.evaluate(() => {
+    const paragraph = document.querySelector('[data-block-id="2"] p');
+    const rect = paragraph?.getBoundingClientRect() as DOMRect;
+    return { x: rect.left + 2, y: rect.top + 8 };
+  });
```
lawvs, line 2103: I suggest use this pattern

```ts
  const helloLocator = page.locator('[data-block-id="2"] p');
  const helloRect = await helloLocator.boundingBox();
  if (!helloRect) throw Error();
  const helloPosition = { x: helloRect.x + 2, y: helloRect.y + 8 };
```

Issue #1309, opened 2023-02-21
Undo/Redo should clear selection


fundon:
Clear selection should be the cleanest operation.

doodlewind:
Closed by #1315

PR #1315, opened 2023-02-21
feat: clear selection after undo/redo

Innei:
#1309


packages/blocks/src/page-block/utils/bind-hotkey.ts
```diff
@@ -476,6 +478,16 @@ export function bindHotkeys(
...
 }
 
+function clearSelection(page: Page) {
+  if (!page.root) return;
+  const defaultPageBlock = getDefaultPageBlock(page.root);
+
+  if ('selection' in defaultPageBlock) {
+    // this is not EdgelessPageBlockComponent
+    defaultPageBlock.selection.clear();
+    defaultPageBlock.selection.state.clear();
```
fundon, line 488: This line should not be needed.

lawvs, line 488:
```suggestion
```

Since `defaultPageBlock.selection.clear();` also clear state

lawvs (approved):
Very impressive improvement!

packages/blocks/src/page-block/utils/bind-hotkey.ts
```diff
@@ -476,6 +478,15 @@ export function bindHotkeys(
   // Don't forget to remove hotkeys at `removeHotkeys`
 }
 
+function clearSelection(page: Page) {
+  if (!page.root) return;
+  const defaultPageBlock = getDefaultPageBlock(page.root);
+
+  if ('selection' in defaultPageBlock) {
```
himself65, line 485: use `instanceof`

lawvs: The `getDefaultPageBlock` has an incorrect type, I should we can fix it after

PR #1325, opened 2023-02-21
fix: sub selected blocks should not be drawn when pressing tab key

PerfectPan:
https://user-images.githubusercontent.com/24316656/220421819-ac69eacb-1807-4d46-ac10-903658e4c0e4.mp4

try to fix the above problem, the subtree's blocks are also drawn rects, which is a wrong behavior.


lawvs (commented):
Quite a decent fix, thanks!

fundon (approved):

fundon:
LGTM :)

Issue #1298, opened 2023-02-20
Text overflow in block hub list


Heroboria:
**Section "Insert Blocks" → "Text Block" → Heading 1 - 6**

**Text does not fit in block**

![headings](https://user-images.githubusercontent.com/118163226/220055993-8d09926b-927f-4270-b34a-1960e8453be8.jpg)

thorseraq:
Hi, do you have any browser plugins that modified the font size or page layout? The width of card, text container, and font size if fixed, so it seems unlikely to look like this :)  @Heroboria

Heroboria:
> Hi, do you have any browser plugins that modified the font size or page layout? The width of card, text container, and font size is fixed, it seems unlikely to look like this :) @Heroboria

No, the browser is absolutely clean. This happens in any browser.

thorseraq:
Need more information to reproduce: os, screen resolution, browser version. I tested in some of our machine, it looks normal even you window it to small size
<img width="500" alt="image" src="https://user-images.githubusercontent.com/20554850/220395166-dc6d54c5-f9af-4a53-9544-f65b084de1b3.png">

Heroboria:
Os - Windows 11
Res - Microsoft Edge 110.0.1587.50 and Firefox Developer Browser 111.0b4 and etc

This happens in any interface language, in any browser. OS scaling has no effect.

thorseraq:
Screen resolution?

thorseraq:
I know what causes this

thorseraq:
`Helvetica Neue, Tahoma, PingFang SC, Microsoft Yahei, Arial` those font fallbacks all fails in your browser, with `Hiragino Sans GB`, the line breaking will happen

Heroboria:
> Screen resolution?

2560x1440

PR #1336, opened 2023-02-22
fix: text overflow when font-family fallback fails until "Hiragino Sans GB"

thorseraq:
close https://github.com/toeverything/blocksuite/issues/1298


doodlewind (approved):

doodlewind:
Thanks for the detailed fix!

PR #1337, opened 2023-02-22
feat(phasor): support brush

### LINKED ISSUE #496: Brush element support
Upstream: https://github.com/toeverything/blocksuite/issues/1268

alt1o:
fix #496 
image


packages/playground/examples/canvas/index.html
```diff
@@ -6,19 +6,30 @@
...
       #container {
         position: relative;
         margin: auto;
         margin-top: 10px;
         width: 600px;
         height: 400px;
         border: 1px #ddd solid;
+        overflow: hidden;
+      }
+      #hover-rect {
```
doodlewind, line 22: There is an visible offset the should be addressed. Maybe can try switching to `box-sizing: border-box`?

packages/phasor/src/elements/brush/brush-element.ts
```diff
@@ -0,0 +1,76 @@
+import { Utils } from '../../utils/tl-utils.js';
+import { Vec } from '../../utils/vec.js';
+import { deserializeXYWH, serializeXYWH } from '../../utils/xywh.js';
+import { BaseElement, HitTestOptions } from '../base-element.js';
+
+export class BrushElement extends BaseElement {
+  type = 'brush' as const;
+  color = '#000000';
+  points: number[][] = [];
```
doodlewind, line 9: Is it possible to use `number[]` with the `[x1, y1, x2, y2...]` layout that could reduce the serialized size?

packages/playground/examples/canvas/viewport-state.ts
```diff
@@ -0,0 +1,74 @@
+export class ViewportState {
```
doodlewind, line 1: I think this is a duplicated file, could you justify why doing hard copy-paste?

packages/playground/examples/canvas/main.ts
```diff
@@ -21,20 +24,78 @@ function testClick(surface: SurfaceManager, e: MouseEvent) {
...
 
   // Uncomment to batch load mock data
   // initMockData(surface, 100, 1000, 1000);
 
   surface.initDefaultGestureHandler();
   canvas.addEventListener('click', e => testClick(surface, e));
+  canvas.addEventListener('mousemove', e =>
+    mouseMove(surface, viewportState, e)
+  );
+  canvas.addEventListener('wheel', e => mouseMove(surface, viewportState, e));
```
doodlewind, line 98: I think we already have registered wheel event listener here, we should not add a new one.

packages/phasor/src/elements/brush/brush-element.ts
```diff
@@ -0,0 +1,76 @@
...
+  lineWidth = 4;
+
+  get tiny() {
```
doodlewind, line 12: It doesn't feel right to have an ad-hoc processing for `tiny`. We should expand the hit test area for picking element. For example, there is an `expandStroke` option in `HitTestOptions` that is prepared for this case, not used ATM though.

packages/playground/examples/canvas/main.ts
```diff
@@ -21,20 +24,78 @@ function testClick(surface: SurfaceManager, e: MouseEvent) {
...
+
+  const viewportState = new ViewportState();
+  const bound = canvas.getBoundingClientRect();
+  viewportState.setSize(bound.width, bound.height);
+  viewportState.setCenter(0, 0);
+
   const surface = new SurfaceManager(canvas, yContainer);
 
+  initWheelEventHandlers(surface, viewportState, canvas);
+  surface.setViewport(
```
doodlewind, line 63: Opinionated speaking, I think this part feel hacky and makes the `main` too verbose.

packages/phasor/src/elements/brush/brush-element.ts
```diff
@@ -0,0 +1,76 @@
...
+    const point = Vec.sub([x, y], [this.x, this.y]);
+
+    if (this.tiny) {
+      return Utils.pointInCircle(point, [0, 0], this.lineWidth / 2);
+    }
+    return Utils.pointInPolyline(point, this.points, this.lineWidth);
+  }
+
+  render(ctx: CanvasRenderingContext2D): void {
+    ctx.beginPath();
```
doodlewind, line 26: No matching `closePath`, this will lead to performance degeneration after running for a while.

doodlewind (changes requested):
This PR demonstrates the hover effect by modifying the phasor example. It's good to add several lines adding new elements to render there, but should not bring in duplicated files that doesn't feel right. Instead, having the basic hover support in BlockSuite playground may be more appropriate.

himself65:
how to enable brush mode in playground?

image

packages/phasor/src/elements/brush/brush-element.ts
alt1o: `tiny` is not only for hit testing but also for rendering. When the pattern is too small, only draw a circle instead of whole paths.

packages/phasor/src/elements/brush/brush-element.ts
alt1o: I'm not sure about this change. Because if we want to add brush pressure in the future, it's not easy for upgrading.

packages/phasor/src/elements/brush/brush-element.ts
alt1o: So the better solution is using `Path2D`? I'm not sure and still try to find some information.

alt1o:
> This PR demonstrates the hover effect by modifying the phasor example. It's good to add several lines adding new elements to render there, but should not bring in duplicated files that doesn't feel right. Instead, having the basic hover support in BlockSuite playground may be more appropriate.

The thing begins with this. When I try to draw shapes in phasor, I am not sure the hit test area is correct. I click here and here and here. Then I thought the hover effect maybe fix my problem. I believe there should be an easier method to test hit testing.

alt1o:
> how to enable brush mode in playground?
> 
> image

Not integrated with edgeless mode.

doodlewind:
Just saying that we don't have to add labels to PRs. See https://github.com/toeverything/AFFiNE/pull/1061#issuecomment-1432816825

packages/phasor/src/elements/brush/brush-element.ts
doodlewind: LGTM

packages/phasor/src/elements/brush/brush-element.ts
doodlewind: Okay so let's keep using nested array for `pressure`.

alt1o:
Try to use `perfect-freehand` render brush path. But there are some bad cases. So use `lineTo` currently.

image

doodlewind (approved):

doodlewind:
We should prioritize using perfect-freehand, let's fix its issues in the following PRs.

PR #1338, opened 2023-02-22
fix(blocks): move h1-h6 paragraph children into parent

### LINKED ISSUE #1074: move h1-h6 paragraph children into parent
> In my opinion
> So far we don't allow headings blocks have child-item, the expected effect is that:
> **After changing the paragraph style to headings, its child-items is no longer child-item.**
> And this apply to other block style which is not allowed have child-item

Refs: https://github.com/toeverything/blocksuite/issues/1059#issuecomment-1422040873

Flrande:
Close #1074

### DIFF
--- a/packages/blocks/src/paragraph-block/paragraph-service.ts
+++ b/packages/blocks/src/paragraph-block/paragraph-service.ts
@@ -1,3 +1,4 @@
+import { handleUnindent } from '../__internal__/rich-text/rich-text-operations.js';
 import { BaseService } from '../__internal__/service/index.js';
 import type { ParagraphBlockModel } from './paragraph-model.js';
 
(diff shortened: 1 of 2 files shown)


doodlewind (approved):

packages/blocks/src/paragraph-block/paragraph-service.ts
```diff
@@ -34,6 +35,24 @@ export class ParagraphBlockService extends BaseService {
         return text;
     }
   }
+
+  /**
+   * side effect when update block type
+   */
+  updateTypeEffect(model: ParagraphBlockModel, newType: string) {
+    // if change to header, move all children to parent level
+    if (newType.startsWith('h')) {
```
lawvs, line 44: You should use `isSupportChildren`

https://github.com/toeverything/blocksuite/blob/fb42cf1ce0c0ff6b4ae9fa8486fb366957e639ea/packages/blocks/src/__internal__/utils/std.ts#L6-L20

Flrande: I will update it in the next pr, where I will add test as well.

PR #1339, opened 2023-02-22
feat(phasor): add more shapes

alt1o:
image


Himself65 (approved):

doodlewind:
This looks good generally, I just would like to have some minor refactoring on this branch.

packages/phasor/src/elements/shape/types.ts
```diff
@@ -0,0 +1,43 @@
...
+export interface ShapeRenderConfig {
+  width: number;
+  height: number;
+  strokeWidth: number;
+  fillColor?: string;
+  strokeColor?: string;
+  strokeStyle?: 'solid';
+}
+
+interface RenderSequenceStroke {
```
doodlewind, line 20: I'm not sure if we should switch to command-driven render at the moment. Looks better to revert if this can't bring justifiable performance gain.

alt1o: I'm not sure too. I read the origin code and find `Path2D` cache. So I thought you design this for a purpose. In order to make cache logic consistent, I create `RenderSequence`.

doodlewind:
TODOs:

* Merge the implementation of `RoundedRect` and `Rect`

doodlewind (approved):

PR #1342, opened 2023-02-23
chore: remove usage of "--affine-font-xs"

thorseraq:


doodlewind (approved):

packages/blocks/src/components/blockhub.ts
```diff
@@ -329,6 +329,10 @@ export class BlockHub extends NonShadowLitElement {
       border-radius: 10px 10px 0 10px;
     }
 
+    [role='menu-entry'] tool-tip {
```
lawvs, line 332: > ARIA roles are added to HTML elements using role="role type", where role type is the name of a role in the ARIA specification. Some roles require the inclusion of associated ARIA states or properties; others are only valid in association with other roles.

The `role` is not a string instead of an enum, `menu-entry` is not a valid role value

thorseraq: Hi, get that, keeping the convention is good (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles)

PR #1346, opened 2023-02-23
feat(blocks): drag & drop multiple blocks

pengx17:
- [x] fix [dnd continuous blocks to another block](https://github.com/toeverything/blocksuite/issues/1324) (top/bottom). The dropped blocks will be selected after that
- [x] improve types
- [x] tests
- [x] fix handle offset https://github.com/toeverything/blocksuite/issues/1352
- [x] playground offset switch button to test blocksuite editor with offset 

### The handle offset issue
Root cause: the handle was be attached to the parent of `<editor-container />`. However, unless we wrap `<editor-container />` into a relative container without padding, there will be offset issue for handles.
The solution in this pr moves the handle as a child of `<editor-container />` and make `<editor-container />` a relative container. As a result, there are some event issues this time, mainly `mousedown` & `mousemove` events conflicts defined in `initMouseEventHandlers`. Solution here is to stop event propagation in `dragHandle` event handlers accordingly. 

https://user-images.githubusercontent.com/584378/220875092-640c9f75-cc23-46a9-9778-8c53ef116e47.mp4

### DIFF
--- a/packages/blocks/src/__internal__/utils/query.ts
+++ b/packages/blocks/src/__internal__/utils/query.ts
@@ -10,6 +10,15 @@ import { getCurrentRange } from './selection.js';
 
 type ElementTagName = keyof HTMLElementTagNameMap;
 
+export type BlockComponentElement =
+  HTMLElementTagNameMap[keyof HTMLElementTagNameMap] extends infer U
+    ? U extends { model: infer M }
+      ? M extends BaseBlockModel
+        ? U
+        : never
+      : never
+    : never;
+
 interface ContainerBlock {
   model?: BaseBlockModel;
 }
@@ -18,10 +27,14 @@ export function getBlockById<T extends ElementTagName>(
   id: string,
   container: Element = document.body
 ) {
-  return container.querySelector<T>(`[${ATTR}="${id}"]` as T);
+  return container.querySelector<T>(
+    `[${ATTR}="${id}"]` as T
+  ) as BlockComponentElement | null;
 }
 
-export function getBlockByPoint(point: IPoint): Element | null | undefined {
+export function getBlockByPoint(
+  point: IPoint
+): BlockComponentElement | null | undefined {
   return document.elementFromPoint(point.x, point.y)?.closest(`[${ATTR}]`);
 }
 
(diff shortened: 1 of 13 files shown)


pengx17:
I will add tests tomorrow

packages/blocks/src/components/drag-handle.ts
```diff
@@ -373,18 +386,38 @@ export class DragHandle extends LitElement {
...
+    )
+      ? selectedBlocks
+      : [clickDragState.element];
+
+    // hack: set opacity to 0.9 to remove dragging element's shadow
+    // maybe the dragging element also has opacity?
+    const firstElement = draggingBlockElements[0];
+
+    // fixme: handle multiple blocks case
+    // the drag image right now only renders the first block
```
fundon, line 416: May need this https://stackoverflow.com/questions/34200023/drag-drop-set-custom-html-as-drag-image

pengx17: Let me have a try

pengx17: I tried a solution in [`14c7608` (#1346)](https://github.com/toeverything/blocksuite/pull/1346/commits/14c760883f7e47995e7fd8934deb512c19803ffb#diff-4e6ed695466ea09e7e5564e244dd064445d6379362b0bff473a400155f99ab94R479-R494).

It clones every selected item and then puts them into a new div element as the drag image. The drag image will be placed somewhere not to be seen by the users.
It does work at the beginning but I realize it has some side effects after I cloned the nodes. The nodes we are cloning are web components that will have the same lifecycles of the predefined behaviors and this will cause some unexpected issues.
Now the solution is to copy the selection rects instead when there are multiple selections as a workaround.

I also noticed that the solution does not work well in Safari, which is bad news as this issue will also happen in Tarui on Mac.

Himself65 (approved):

packages/blocks/src/__internal__/utils/query.ts
```diff
@@ -10,6 +10,15 @@ import { getCurrentRange } from './selection.js';
...
 type ElementTagName = keyof HTMLElementTagNameMap;
 
+export type BlockComponentElement =
+  HTMLElementTagNameMap[keyof HTMLElementTagNameMap] extends infer U
+    ? U extends { model: infer M }
+      ? M extends BaseBlockModel
+        ? U
+        : never
+      : never
+    : never;
```
himself65, line 20: nice type

PR #1348, opened 2023-02-23
refactor(edgeless): split toolbar modules

alt1o:
https://user-images.githubusercontent.com/21084335/220945887-c7faedb2-c912-4728-9c69-9dfca2ca6897.mov


packages/blocks/src/page-block/edgeless/toolbar/shape-tool-button/shape-tool-button.ts
```diff
@@ -0,0 +1,95 @@
...
+
+@customElement('edgeless-shape-tool-button')
+export class EdgelessShapeToolButton extends LitElement {
+  @property()
+  mouseMode?: MouseMode;
+
+  @property()
+  edgeless?: EdgelessPageBlockComponent;
+
+  private _shapesMenu?: ShapesMenuInstance;
```
doodlewind, line 50: We prefer explicit `null` as the void value.

doodlewind:
TODOs:

* Try remove the `edgeless-toolbar-with-flag` wrapper and attach edgeless-toolbar without imperative `document.createElement`

Himself65 (commented):
I know what you are doing, but the first left icon is `quare`

doodlewind (approved):
LGTM, cheers!

Issue #1319, opened 2023-02-21
Can not insert text at the end of inline code


yjl9903:
Open the AFFiNE alpha downhills, then move cursor to the end of `in-line code blocks` and follow the following gif.

I can not insert text at the end of inline code block, and the cursor position may also have some bugs?

![bug](https://user-images.githubusercontent.com/30072175/220366717-1e57113e-34e3-416b-8c79-bbc9c98234a5.gif)

System: 
+ OS: Windows 10 10.0.22621
+ CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11700 @ 2.50GHz
+ Memory: 21.86 GB / 31.90 GB

Browser: Edge: Spartan (44.22621.1265.0), Chromium (110.0.1587.50)

lawvs:
Thank you for reporting this bug. We appreciate your feedback and will work on fixing this issue as soon as possible.

Also related to https://github.com/toeverything/blocksuite/issues/415

PR #1353, opened 2023-02-24
fix: inline code break when pressing Enter followed by Backspace twice

zqran:
close: #1319

### DIFF
--- a/packages/blocks/src/__internal__/rich-text/rich-text.ts
+++ b/packages/blocks/src/__internal__/rich-text/rich-text.ts
@@ -160,7 +160,6 @@ export class RichText extends NonShadowLitElement {
         attr &&
         source === 'user'
       ) {
-        // console.log(this.quill);
         const retain = delta.ops[0].retain;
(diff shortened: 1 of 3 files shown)


packages/blocks/src/__internal__/rich-text/rich-text.ts
```diff
@@ -154,7 +154,13 @@ export class RichText extends NonShadowLitElement {
...
       }
       // only length is 2 need to be handled
-      if (delta.ops.length === 2 && delta.ops[1]?.insert && attr) {
+      if (
+        delta.ops.length === 2 &&
+        delta.ops[1]?.insert &&
+        attr &&
+        source === 'user'
+      ) {
+        // console.log(this.quill);
```
himself65: remove

Himself65 (changes requested):

doodlewind (approved):

doodlewind:
Good catch, thanks!

PR #1357, opened 2023-02-24
ci: add build command & OS dependencies options

PerfectPan:
In broswer-compatibilty workflow all tests failed because of not running build command and playwright use vite preview in CI.

And ubuntu webkit environment seems need to add —install-deps options to install OS dep.

But still a lot of test failed.


lawvs (approved):
nice catch!

Himself65 (approved):

Issue #1183, opened 2023-02-15
Support showing format quick bar in block level selection


doodlewind:
Upstream: https://github.com/toeverything/AFFiNE/issues/874

Currently the format quick bar only shows with native range selection, the expected state will be also support it with block level selection.

PR #1358, opened 2023-02-24
fix: format bar should follow selection block

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/1183


fundon:
Thanks, fixed my spelling errors. 🤣

PR #1361, opened 2023-02-25
refactor: eliminate redundant code block button inheritance

thorseraq:
With help of @pengx17, we came to a conclusion that
- `CodeBlockButton` doesn't need Inheriting `IconButton`, as it only uses `IconButton`'s static property `styles`.
- This PR also fixes the apparent delay when clicking the code block button (which is used to show code language selector) in the local dev env when hmr is enabled.

### Before:
image
image

### After:
image

### DIFF
--- a/packages/blocks/src/code-block/components/button.ts
+++ b/packages/blocks/src/code-block/components/button.ts
@@ -1,11 +1,11 @@
-import { css, html } from 'lit';
+import { css, html, LitElement } from 'lit';
 import { customElement, property } from 'lit/decorators.js';
 
 import { IconButton } from '../../components/button.js';
 
 // TODO reuse existing button component
 @customElement('code-block-button')
-export class CodeBlockButton extends IconButton {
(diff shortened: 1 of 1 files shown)


pengx17 (approved):
LGTM

packages/blocks/src/code-block/components/button.ts
```diff
@@ -1,11 +1,11 @@
-import { css, html } from 'lit';
+import { css, html, LitElement } from 'lit';
 import { customElement, property } from 'lit/decorators.js';
 
 import { IconButton } from '../../components/button.js';
 
 // TODO reuse existing button component
```
lawvs, line 6:
```suggestion
```

lawvs (changes requested):
It looks like `code-block-button` is not necessary

lawvs:
Close due to preferring https://github.com/toeverything/blocksuite/pull/1365

PR #1375, opened 2023-02-25
feat: remove redundant `as const`

zqran:


himself65 (changes requested):
It’s necessary. 
const string is different from regular string

zqran:
```ts
const BLOCK_ID_ATTR = 'data-block-id';
```
The variable declared with const is already of type `const BLOCK_ID_ATTR: 'data-block-id'`.
Why use `as const`?

himself65:
<https://stackoverflow.com/questions/66993264/what-does-the-as-const-mean-in-typescript-and-what-is-its-use-case>
***@***.***
What does the "as const" mean in TypeScript and what is its use case?<https://stackoverflow.com/questions/66993264/what-does-the-as-const-mean-in-typescript-and-what-is-its-use-case>
stackoverflow.com<https://stackoverflow.com/questions/66993264/what-does-the-as-const-mean-in-typescript-and-what-is-its-use-case>


Sent from my iPhone

On Feb 25, 2023, at 19:24, zqran ***@***.***> wrote:

﻿

const BLOCK_ID_ATTR = 'data-block-id';

The variable declared with const is already of type const BLOCK_ID_ATTR: 'data-block-id'.
Why use as const?

zqran:
The value mentioned by [what-does-the-as-const-mean-in-typescript-and-what-is-its-use-case](https://stackoverflow.com/questions/66993264/what-does-the-as-const-mean-in-typescript-and-what-is-its-use-case) is a non-primitive value.

```ts
const args = [8, 5];
// const args: number[]

const args = [8, 5] as const;
// const args: readonly [8, 5] // -> readonly tuple
```

But string is a primitive value, 

```ts
const BLOCK_ID_ATTR = 'data-block-id';
// const BLOCK_ID_ATTR: 'data-block-id'

const BLOCK_ID_ATTR = 'data-block-id' as const;
// const BLOCK_ID_ATTR: 'data-block-id'
```

The compiler sees `const BLOCK_ID_ATTR = 'data-block-id';` and infers the type of `'data-block-id'`(literal type). 
So, in this scenario, I think `as const` can be omitted here.

himself65 (approved):

himself65:
good catch

PR #1390, opened 2023-02-26
feat: typed `matchFlavours` utility function

Saul-Mirone:
Before:

```ts
// we have a model here
declare const model: BaseBlockModel;
if (matchFlavours(mode, ['affine:page']) {
  model.title // 💥  Error! The type of model is still BaseBlockModel. 
}
```

After:

```ts
if (matchFlavours(model, ['affine:page'] as const) {
  //                        ^ trade off here, we have to use `as const` to generate the tuple
  model.title // 🚀 Works! The type of model is PageBlockModel
}

// What will happen for multiple flavours?
if (matchFlavours(model, ['affine:code', 'affine:paragraph'] as const) {
  model // 🔮 CodeBlockModel | ParagraphModel
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,8 @@
   "license": "MPL-2.0",
   "devDependencies": {
     "@blocksuite/global": "workspace:*",
+    "@blocksuite/store": "workspace:*",
+    "@blocksuite/blocks": "workspace:*",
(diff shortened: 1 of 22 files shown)


Himself65 (approved):
very helpful!

PR #1392, opened 2023-02-27
refactor: clearly distinguish `state` and `property` in lit

zqran:
1. clearly distinguish `state`(private or protected) and `property`(public)
2. remove unused properties

Trade off:
Errors may occur if Affine uses `state` directly. However, this is more intuitive for blockssuite.


lawvs (approved):
Impressive work

packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -161,9 +161,9 @@ export class DefaultPageBlockComponent
   mouseRoot!: HTMLElement;
 
   @state()
-  frameSelectionRect: DOMRect | null = null;
+  private frameSelectionRect: DOMRect | null = null;
 
-  @state()
+  @property()
```
lawvs, line 166: > This decorator `property` should only be used for public fields. As public fields, properties should be considered as primarily settable by element users, either via attribute or the property itself.

This field `viewportState` seems to be misused. I think it's better to keep it as it is for now.

```suggestion
  @state()
```

zqran: https://github.com/toeverything/blocksuite/blob/956afab69811bad17af70eacb9f24e044442f0d9/packages/blocks/src/page-block/default/selection-manager.ts#L547

The `viewportState` attribute will be used outside the component here

zqran: Alternatively, we can provide a getter to use `viewportState` property outside the component

packages/blocks/src/code-block/code-block.ts
```diff
@@ -199,13 +199,7 @@ export class CodeBlockComponent extends NonShadowLitElement {
...
 
   @state()
-  showLangList = 'hidden';
-
-  @state()
-  disposeTimer = 0;
-
-  @state()
-  filterText = '';
+  private showLangList = 'hidden';
```
doodlewind, line 202: Well I think we can enforce the `_showLangList` convention for private variable here 🤔

zqran:
Why E2E tests always fail?

doodlewind:
> Why E2E tests always fail?

The CI log here doesn't look like flaky, seems that this PR (unexpectedly) breaks some existing test. Let me take a look.

zqran:
When I run the test locally, a single test can succeed(`pnpm run test code`), but all(`pnpm run test`) the tests will fail. 
I am a little confused. So, I don't know how to determine the cause of the problem.

doodlewind:
> When I run the test locally, a single test can succeed(`pnpm run test code`), but all(`pnpm run test`) the tests will fail. I am a little confused. So, I don't know how to determine the cause of the problem.

See the log from CI (https://github.com/toeverything/blocksuite/actions/runs/4293180269/jobs/7480562569), scroll to last part, and then we can try to reproduce the "failed" ones locally:

```
Notice:   14 failed
    clipboard.spec.ts:343:1 › should keep first line format when pasted into a new line ============
    drag.spec.ts:68:1 › move drag handle in list ===================================================
    drag.spec.ts:78:1 › move drag handle in nested block ===========================================
    drag.spec.ts:112:1 › move to the last block of each level in multi-level nesting ===============
    drag.spec.ts:399:1 › should be able to drag & drop multiple blocks to nested block =============
    embed.spec.ts:284:1 › popup menu should follow position of image when scrolling ================
    format-quick-bar.spec.ts:612:1 › should format quick bar follow scroll =========================
    list.spec.ts:28:1 › add new bulleted list ======================================================
    list.spec.ts:43:1 › convert to numbered list block =============================================
    list.spec.ts:280:1 › should indent todo block preserve todo status =============================
    markdown.spec.ts:24:1 › markdown shortcut ======================================================
    selection.spec.ts:524:1 › click the list icon can select and copy ==============================
    selection.spec.ts:726:1 › should not crash when mouse over the left side of the list block prefix 
    slash-menu.spec.ts:168:1 › should slash menu search and keyboard works =========================
  10 flaky
    basic.spec.ts:92:1 › A open and edit, then joins B =============================================
    basic.spec.ts:112:1 › A first open, B first edit ===============================================
    clipboard.spec.ts:77:1 › markdown format parse =================================================
    clipboard.spec.ts:236:1 › copy clipItems format ================================================
    clipboard.spec.ts:464:1 › paste in block-level selection =======================================
    hotkey.spec.ts:478:1 › should hotkey work in paragraph =========================================
    link.spec.ts:250:1 › should keyboard work in link popover ======================================
    list.spec.ts:106:1 › insert new list block by enter ============================================
    list.spec.ts:125:1 › delete at start of list block =============================================
    paragraph.spec.ts:562:1 › delete at start of paragraph immediately following list ==============
  6 skipped
  201 passed (18.4m)
```

zqran:
Due to the large number of changed files this time, it is difficult to troubleshoot the cause of the test failure.
So, I will close this pr, please check the latest implementation: #1428.

Issue #1384, opened 2023-02-26
Replacing an entire selection consisting of multiple blocks with text appends it to the end of the selection instead of replacing it properly


AlessioGr:
How it currently works:

https://user-images.githubusercontent.com/70709113/221417440-1efe04a4-3aaf-4c4b-b8bc-9e89bf9880c1.mp4

How I expect it to work:

https://user-images.githubusercontent.com/70709113/221417462-289a223e-42aa-4b6b-bc4b-ff6abc0fcf6d.mp4

Issue #620, opened 2023-01-09
Unable to paste content in block-level selection state


doodlewind:
When the editor is under block-level selection:

* The pasted content is ignored.
* The selected blocks are not removed.

SteinsHead:
I think the correct logic is that the selected block should be removed first, and then the pasted block should be added to the location of the removed block

SteinsHead:
#718 This is my code for this problem

doodlewind:
Closed by #1687

PR #1396, opened 2023-02-27
fix: paste in block-level selection

lzlme:
fix: #1384 and #620


packages/editor/src/managers/clipboard/paste-manager.ts
```diff
@@ -1,4 +1,9 @@
-import type { OpenBlockInfo } from '@blocksuite/blocks';
+import type {
+  getCurrentBlockRange,
```
lawvs, line 2: > Type error: 'getCurrentBlockRange' cannot be used as a value because it was imported using 'import type'.

lzlme:
My code did not pass the E2E test, but I think there is something wrong with the previous test.
For example `test('drag copy paste')` in `code.spec.ts`: type 'use', then select it, then copy and paste, and expect to get 'useuse', which is a bug in itself.
When performing a paste operation, if there is already selected text, shouldn't it replace the selected content?
bug

tests/clipboard.spec.ts
```diff
@@ -460,3 +460,21 @@ test('cut should work for multi-block selection', async ({ page }) => {
   await page.keyboard.press(`${SHORT_KEY}+v`);
   await assertRichTexts(page, ['a', 'b', 'c']);
 });
+
+test.only('paste in block-level selection', async ({ page }) => {
```
lawvs:
```suggestion
test('paste in block-level selection', async ({ page }) => {
```

Issue #1244, opened 2023-02-16
poor performance with HMR


lawvs:
The HMR feature is incredibly useful. It has saved me a lot of time, especially when adjusting styles.

But I encountered a minor issue during the usage.

If you use the slash menu several times, there is a delay when HMR is activated.

Related to https://github.com/toeverything/blocksuite/pull/621

/cc @fi3ework 

<img width="1073" alt="Screenshot 2023-02-16 at 7 29 06 PM" src="https://user-images.githubusercontent.com/18554747/219353116-51414bff-8298-4bbd-9c52-9984e31ee283.png">

https://user-images.githubusercontent.com/18554747/219352156-348b7989-d189-4f2e-8e31-f20575b95bb7.mov

lawvs:
Also occurs when clicking the language button in the code block.


https://user-images.githubusercontent.com/18554747/220293931-c2a6109d-9645-401c-b839-546fe54aaf3c.mov

lawvs:
I believe it is related to extending `LitElement`

See https://github.com/toeverything/blocksuite/pull/1361

doodlewind:
We've made the HMR command opt-in, so this issue has been mitigated and is not of major concern.

PR #1397, opened 2023-02-27
perf: disable hmr plugin by default

pengx17:
It seems the HMR plugin will cause serious performance issues in development mode, especially for the heavy example.
Need to disable it before upstream is fixed.

In the performance profiling report, there are many tiny tasks within `__web-dev-server__/wc-hmr/runtime.js`, which is in the hmrPlugin:
image

cc @fi3ework


doodlewind (commented):
Perhaps we can make non-hmr mode (instead of `dev:hmr`) by default?

pengx17:
@fi3ework for the time being using HMR mode, I've noticed that sometimes the plugin doesn't re-register event handlers after an update. As a result, I frequently have to perform a full page reload manually. While I believe that HMR has the potential to greatly enhance the developer experience (DX), I think that the current version of the wc-hmr plugin falls short in comparison to similar plugins available for React and Vue.

lawvs:
Also related to https://github.com/toeverything/blocksuite/issues/1244

pengx17:
@lawvs Thanks for the additional insights on this issue!
I would assume it may take quite some effort to fix this issue upstream. I think we can vote on whether or not to enable HMR by default?

doodlewind (approved):

doodlewind:
Let's just land the fallback and I've added #1244 into out roadmap for metrics.

Issue #1395, opened 2023-02-27
The cursor does not disappear after select the code block


JimmFly:
I tried dragging and clicking the drag handle to select the code block, but the cursor still doesn't go away.

https://user-images.githubusercontent.com/102217452/221499187-55b9421a-c6cc-4453-9976-c77400b1cbc3.mov

PR #1400, opened 2023-02-27
fix: unfocus rich-text on starting block selection

fundon:
Fixes: #1395

### DIFF
--- a/packages/blocks/src/page-block/default/selection-manager.ts
+++ b/packages/blocks/src/page-block/default/selection-manager.ts
@@ -385,6 +385,12 @@ export class PageSelectionState {
     }
   }
 
+  blur() {
+    resetNativeSelection(null);
+    // deactivate quill keyboard event handler
+    (document.activeElement as HTMLElement).blur();
+  }
+
   clearRaf() {
     if (this.rafID) {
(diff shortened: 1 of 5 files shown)


packages/blocks/src/page-block/default/selection-manager.ts
```diff
@@ -385,6 +385,12 @@ export class PageSelectionState {
     }
   }
 
+  blur() {
+    resetNativeSelection(null);
+    // deactivate quill keyboard event handler
+    (document.activeElement as HTMLElement).blur();
```
lawvs, line 391: > The [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element) which currently has focus, [<body>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/body) or null if there is no focused element.
>
> https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement


This cast is unsafe, you should check it first
```suggestion
    if (
      document.activeElement &&
      document.activeElement instanceof HTMLElement
    ) {
      document.activeElement.blur();
    }
```

doodlewind (approved):
LGTM, thanks for your maintenance!

PR #1404, opened 2023-02-27
fix(embed): don't show image option menu on selecting

fundon:
https://user-images.githubusercontent.com/27926/227482051-af614b89-e0f4-410b-b679-dcd6bfa30043.mov

### DIFF
--- a/packages/blocks/src/__internal__/utils/gesture.ts
+++ b/packages/blocks/src/__internal__/utils/gesture.ts
@@ -27,6 +27,7 @@ export interface SelectionEvent extends IPoint {
     alt: boolean;
   };
   button?: number;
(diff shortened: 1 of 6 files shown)


fundon:
test later

doodlewind:
@fundon looks like this PR is in stale state. Could you try if the original issue (not sure what it is) can still be reproduced in our current version? Let's see what we can do to help this landed!

fundon:
> @fundon looks like this PR is in stale state. Could you try if the original issue (not sure what it is) can still be reproduced in our current version? Let's see what we can do to help this landed!

I've just added a video

doodlewind (approved):
Let's land this, thanks for your follow up!

PR #1408, opened 2023-02-28
refactor: migrate `BlockSuiteModelProps` to `BlockModelProps`

Himself65:
Use type inference to create the props type instead of rewriting.


Saul-Mirone (approved):
LGTM

PR #1413, opened 2023-02-28
perf(editor): append quill async

pengx17:
The heavy example takes around 5 seconds to load because when `rich-text` is appended to DOM it will also create an Quill instance, which will cause the page to reflow because Quill will access blocking apis like `document.getSelection()`. If there are a lot of rich-text elements being added at once, it will cause the page to freeze for a few seconds.

The proposed fix is to move quill instance instance appending to `requestAnimationFrame`.

The profile charts before & after:

image

image

### DIFF
--- a/packages/blocks/src/__internal__/utils/common-operations.ts
+++ b/packages/blocks/src/__internal__/utils/common-operations.ts
@@ -6,10 +6,14 @@ import type { Quill } from 'quill';
 import type { ExtendedModel } from './types.js';
 
 // XXX: workaround quill lifecycle issue
-export function asyncFocusRichText(page: Page, id: string) {
+export function asyncFocusRichText(page: Page, id: string, retries = 2) {
   requestAnimationFrame(() => {
     const adapter = page.richTextAdapters.get(id);
-    adapter?.quill.focus();
+    if (adapter) {
+      adapter?.quill.focus();
+    } else if (retries > 0) {
+      asyncFocusRichText(page, id, retries - 1);
+    }
   });
 }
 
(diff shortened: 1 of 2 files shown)


packages/blocks/src/__internal__/utils/common-operations.ts
```diff
@@ -6,10 +6,14 @@ import type { Quill } from 'quill';
...
-export function asyncFocusRichText(page: Page, id: string) {
+export function asyncFocusRichText(page: Page, id: string, retries = 2) {
   requestAnimationFrame(() => {
     const adapter = page.richTextAdapters.get(id);
-    adapter?.quill.focus();
+    if (adapter) {
+      adapter?.quill.focus();
+    } else if (retries > 0) {
+      asyncFocusRichText(page, id, retries - 1);
+    }
```
lawvs, line 16: It looks like the DOM is mounted first, then the quill is initialized.

Can we focus on the DOM directly instead of relying on quill API?

pengx17: hmm, I think `focus` here will also mean to place the cursor onto the contenteditable right? When quill is not being appended, the DOM is just an empty placeholder without any contents.

pengx17:
Ops, it looks like the change will cause many regressions. I guess we should do this after we replaced Quill when using Virgo?

doodlewind:
> Ops, it looks like the change will cause many regressions. I guess we should do this after we replaced Quill with Virgo?

Sounds fair, thanks for your investigation!

/cc @Flrande.

Issue #1378, opened 2023-02-26
Pressing Enter when selecting multiple blocks should create new block


lawvs:
https://user-images.githubusercontent.com/18554747/221395933-551ac2d5-97f3-4e8f-a61b-5ee30647b86f.mov

Related code

https://github.com/toeverything/blocksuite/blob/3c6fe84ae2f644febbc604326b04f9bc77fd53a2/packages/blocks/src/page-block/utils/bind-hotkey.ts#L287-L312

PR #1414, opened 2023-02-28
feat: pressing enter with multiple blocks should create block

zqran:
close: #1378

### DIFF
--- a/packages/blocks/src/page-block/utils/bind-hotkey.ts
+++ b/packages/blocks/src/page-block/utils/bind-hotkey.ts
@@ -298,7 +298,35 @@ export function bindHotkeys(
       selection.clear();
       return;
     }
-    // TODO fix native selection enter
+    if (blockRange.type === 'Native') {
+      e.stopPropagation();
+      e.preventDefault();
+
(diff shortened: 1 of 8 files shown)


packages/blocks/src/page-block/utils/bind-hotkey.ts
```diff
@@ -298,7 +298,35 @@ export function bindHotkeys(
       selection.clear();
       return;
     }
-    // TODO fix native selection enter
+    if (blockRange.type === 'Native') {
+      e.stopPropagation();
+      e.preventDefault();
```
lawvs, line 303: Is it necessary?

zqran: i'll fix it

lawvs (commented):
Could you please add some tests for it?

zqran:
Why did the test pass locally and fail here? 🤔

lawvs:
It looks like you break the entering behavior in the code-block, can you check it?

lawvs (approved):
Thank you very much for your contribution. I will handle some edge cases. Please sit back and relax.

PR #1418, opened 2023-02-28
feat(edgeless): support adding brush in edgeless mode

alt1o:
- [x] Shapes
- [x] Brush
- [x] Shortcut for deleting single shape
- [x] Move brush shape in default mouse mode


packages/phasor/src/elements/brush/brush-element.ts
```diff
@@ -33,24 +33,42 @@ export function getBrushBoundFromPoints(
 export class BrushElement extends BaseElement {
   type = 'brush' as const;
   color = '#000000';
+  point: [number, number] = [0, 0];
```
doodlewind, line 36: What is this field used for?

doodlewind: If a brush element is being dragged and moved, shall we update its `point` instead of `xywh`, or both of them? If it's true, this may be an anti-pattern that breaks our convention.

doodlewind (approved):
Good to land now, cheers

PR #1423, opened 2023-02-28
feat(edgeless): view control panel

alt1o:
https://user-images.githubusercontent.com/21084335/221862035-5fdef0c7-6c91-4550-b9b7-b7ad87e03db4.mov


packages/phasor/src/surface.ts
```diff
@@ -32,6 +32,10 @@ export class SurfaceManager {
     this._yElements.observeDeep(this._handleYEvents);
   }
 
+  get elements() {
```
doodlewind, line 35: This should not be exposed, it's concerning to let users fully traverse all elements.

packages/blocks/src/page-block/edgeless/view-control-bar.ts
```diff
@@ -0,0 +1,138 @@
...
+    this.edgeless.viewport.setZoom(zoom);
+    this.edgeless.signals.viewportUpdated.emit();
+  }
+
+  private _zoomToFit() {
+    const { width, height } = this.edgeless.viewport;
+    const frame = this.edgeless.pageModel.children[0] as FrameBlockModel;
+    const frameXYWH = deserializeXYWH(frame.xywh);
+
+    const bound = getCommonBound([
```
doodlewind, line 94: Looks better to compute bound inside the phasor renderer, in which also holds the viewport state.

packages/phasor/src/surface.ts
doodlewind: Marked as `@deprecated` ATM

doodlewind (approved):
LGTM, cheers

PR #1424, opened 2023-02-28
perf: add yDoc binsize benchmark to ci

### LINKED ISSUE #1176: Format size metrics for real-world use case
Due to the nature of CRDT that serializes all historical changes in binaries ([y-protocol](https://github.com/yjs/y-protocols) based) and the extra meta data that are being carried per-block, we need to have a baseline metrics about the real-world binary size, especially when compared with plain-text formats.

pengx17:
fix https://github.com/toeverything/blocksuite/issues/1176
Add basic benchmark to blocksuite.

There are two tests in this PR
- calculating the rendering duration,
- the ratio between yDoc binary and the JSON serialized output

On benchmark job finished, it shall report something like this in [the PR comment](https://github.com/pengx17/blocksuite/pull/1#issuecomment-1448593311):
image

However it seems this repo does not grant write permissions to github actions so commenting on PR will always fail right now.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -143,6 +143,8 @@ jobs:
   #         fail_ci_if_error: true
 
   benchmark:
+    permissions:
+      pull-requests: write
     name: Benchmark
     runs-on: ubuntu-latest
     needs:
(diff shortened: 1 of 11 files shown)


himself65 (commented):
I think the benchmark is good for our editor, but what's the meaning of binary/json size? I thought it should be a render time analysis. I don't get the point

pengx17:
> I think the benchmark is good for our editor, but what's the meaning of binary/json size? I thought it should be a render time analysis. I don't get the point

Please see https://github.com/toeverything/blocksuite/issues/1176. This benchmark is to have metrics around the yDoc binary overhead over plain old JSON format, suggested by @doodlewind

himself65:
> > I think the benchmark is good for our editor, but what's the meaning of binary/json size? I thought it should be a render time analysis. I don't get the point
> 
> Please see #1176. This benchmark is to have metrics around the yDoc binary overhead over plain old JSON format, suggested by @doodlewind

I see; then I think we need some real world instead of machine-generated ydoc

pengx17:
> I see; then I think we need some real world instead of machine-generated ydoc

Yeah, the one I am using was the `preset` example in the playground. Not sure what kind of doc should we use? Maybe we shall simulate some user edits actions and then output the yDoc which would make more sense

pengx17:
Will think about adding benchmark to PR later.

PR #1433, opened 2023-03-01
refactor(blocks): replace quill with virgo

Flrande:


packages/blocks/src/__internal__/rich-text/virgo/affine-text.ts
```diff
@@ -0,0 +1,73 @@
+import {
```
Saul-Mirone, line 1: IMO virgo shouldn't know the context of the client (which is affine here). But let's refactor this later in next PR.

Flrande: Here affine use virgo as a **package**.

Flrande: This something about how user use virgo because virgo should support developers customize their components and schema.

Saul-Mirone: I agree, but the virgo aims to be an infra editor framework like slate and prosemirror. It's not suitable to create customization for every user inside the package. Instead, we should provide an universal way to make virgo can be customized.

Flrande: I'm not sure if we are on the same channel. I will add the basic use document of virgo later. You can read it and see if the evaluation is reasonable. 🙂

Saul-Mirone: Really sorry, I made a mistake here. I thought the file is inside the virgo package. Never mind.

Saul-Mirone:
Just a hint, I also find that the behavior of `onDoubleClick` also need to be take care of. Otherwise, if you double click to select the first word of a paragraph, and apply any style (like bold) by the toolbar, nothing will happen.

packages/blocks/src/__internal__/rich-text/keyboard.ts
```diff
@@ -274,6 +274,7 @@ export function createKeyboardBindings(
   }
 
   function onBackspace(e: KeyboardEvent, vEditor: AffineVEditor) {
+    console.log(1, e, e.composedPath());
```
himself65: remove

tests/slash-menu.spec.ts
```diff
@@ -90,11 +92,12 @@ test.describe('slash menu should show and hide correctly', () => {
     await assertRichTexts(page, ['/_']);
 
     // And pressing backspace immediately should reappear the slash menu
-    await page.keyboard.press('Backspace');
+    await pressBackspace(page);
+    await waitNextFrame(page);
```
lawvs, line 96:
```suggestion
```

I believe that there is no need to wait

Flrande: Just a try forgetted to removed, I will remove it.

packages/blocks/src/__internal__/rich-text/keyboard.ts
```diff
@@ -405,15 +452,145 @@ export function createKeyboardBindings(page: Page, model: BaseBlockModel) {
...
   return keyboardBindings;
 }
+
+export function createKeyDownHandler(
+  vEditor: AffineVEditor,
+  bindings: KeyboardBindings
+): (evt: KeyboardEvent) => void {
+  const bindingStore: Record<string, KeyboardBinding[]> = {};
+
+  const SHORTKEY = /Mac/i.test(navigator.platform) ? 'metaKey' : 'ctrlKey';
```
Saul-Mirone, line 483: Do we need to consider the iPad users?

```ts
export const safari = /Apple Computer/.test(navigator.vendor)
// `true` for both iOS and iPadOS
const ios = safari && (/Mobile\/\w+/.test(navigator.userAgent) || navigator.maxTouchPoints > 2)
```

Flrande: Are you referring to the scenario of using a keyboard on an iPad?

packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
```diff
@@ -149,11 +149,13 @@ export function handleIndent(page: Page, model: ExtendedModel, offset = 0) {
     children: [...previousSibling.children, model, ...children],
   });
 
-  // FIXME: after quill onload
   requestAnimationFrame(() => {
```
Saul-Mirone, line 152: Instead of using `requestAnimationFrame`, can we add an `onLoad` signal for virgo?

packages/blocks/src/__internal__/rich-text/keyboard.ts
Saul-Mirone: > Are you referring to the scenario of using a keyboard on an iPad?

Yep

packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
Flrande: What `onload` means? Are you referring to the synchronization of the selection area or whether the text status has been loaded into the DOM?

packages/blocks/src/components/slash-menu/slash-menu-node.ts
```diff
@@ -76,7 +76,7 @@ export class SlashMenu extends LitElement {
       // Workaround: Use capture to prevent the event from triggering the keyboard bindings action
       capture: true,
     });
-    this._disposables.addFromEvent(richText, 'focusout', this._onClickAway);
+    // this._disposables.addFromEvent(richText, 'focusout', this._onClickAway);
```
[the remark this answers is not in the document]
Flrande: I am not sure. cc @lawvs

packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
Saul-Mirone: > What `onload` means? Are you referring to the synchronization of the selection area or whether the text status has been loaded into the DOM?

I think it means the editor has loaded all the dom it needs and ready for input.

packages/blocks/src/__internal__/rich-text/keyboard.ts
```diff
@@ -405,15 +452,149 @@ export function createKeyboardBindings(page: Page, model: BaseBlockModel) {
...
   return keyboardBindings;
 }
+
+export function createKeyDownHandler(
+  vEditor: AffineVEditor,
+  bindings: KeyboardBindings
+): (evt: KeyboardEvent) => void {
+  const bindingStore: Record<string, KeyboardBinding[]> = {};
+
+  const safari = /Apple Computer/.test(navigator.vendor);
```
Saul-Mirone, line 483: We should still keep the check for mac here.

Saul-Mirone (approved):

tzhangchi:
Amazing moment！

Issue #965, opened 2023-02-03
Unable to scroll the page when selecting


JimmFly:
When I want to select from the content of this page to the content of the next page, I can't scroll down.

https://user-images.githubusercontent.com/102217452/216514485-77754b62-916e-4d06-bd0d-36ae93b6914b.mov

PR #1440, opened 2023-03-01
fix: keep native selection state when upper and lower thresholds are reached

fundon:
Fixes: #965 

https://user-images.githubusercontent.com/27926/222220880-c392e5bc-bc89-48c7-bafb-0cad94175737.mov


Himself65 (approved):

doodlewind (changes requested):
I think the current `DefaultSelectionManager` is overly complex. Before moving on, we should do some refactoring, it doesn't feel right having a module with ~1200 lines of code.

Issue #1393, opened 2023-02-27
Lost line content when code in indented code block is too long


JimmFly:
I indented code blocks, then added overly long sentences in code blocks, and parts of the code were missing and styled wrong.

https://user-images.githubusercontent.com/102217452/221496363-5783f7d7-abdc-45eb-865d-bab879d4004c.mov

PR #1442, opened 2023-03-01
feat: update code block style

zqran:
close: #1393

### DIFF
--- a/packages/blocks/src/code-block/code-block.ts
+++ b/packages/blocks/src/code-block/code-block.ts
@@ -134,7 +134,7 @@ export class CodeBlockComponent extends NonShadowLitElement {
     }
 
     .affine-code-block-container .ql-syntax {
-      width: 620px;
+      width: 89%;
       margin: 0;
       overflow-x: auto;
       /*scrollbar-color: #fff0 #fff0;*/


doodlewind (approved):
LGTM, thanks!

LeuisKen:
I experience this problem too, and the `width: 89%` seems not the best fit for this issue.
This problem is caused by the `left: 40px` style of `.affine-code-block-container .ql-container` at [here](https://github.com/toeverything/blocksuite/blob/7449ab5aa58185852e63e296b7aafee20cd2964d/packages/blocks/src/code-block/code-block.ts#L155).
Due to this, maybe `width: calc(100% - 40px)` would be a better solution, and I think we need a comment to reference the relationship between these two things.
@doodlewind @zqran

zqran:
@LeuisKen Thanks for your feedback. To be precise, we should determine the widths of 1 and 2 based on the UI design. This requires confirmation from @doodlewind .

image

doodlewind:
> I experience this problem too, and the `width: 89%` seems not the best fit for this issue.
> 
> This problem is caused by the `left: 40px` style of `.affine-code-block-container .ql-container` at [here](https://github.com/toeverything/blocksuite/blob/7449ab5aa58185852e63e296b7aafee20cd2964d/packages/blocks/src/code-block/code-block.ts#L155).
> 
> Due to this, maybe `width: calc(100% - 40px)` would be a better solution, and I think we need a comment to reference the relationship between these two things.

Sounds great! I'm open to PR for tweaking here 😀

PR #1445, opened 2023-03-02
refactor(phasor): switch to `getElementsBound`

alt1o:


doodlewind (approved):
LGTM

packages/blocks/src/page-block/edgeless/view-control-bar.ts
```diff
@@ -90,15 +68,20 @@ export class EdgelessViewControlBar extends LitElement {
     const { width, height } = this.edgeless.viewport;
     const frame = this.edgeless.pageModel.children[0] as FrameBlockModel;
```
doodlewind, line 69: I guess we should support `framesBound` in the future

PR #1449, opened 2023-03-02
feat(edgeless): pan tool

alt1o:
https://user-images.githubusercontent.com/21084335/222356822-fc983d3b-73a1-48da-bfec-5994c71425e5.mov

### DIFF
--- a/packages/blocks/src/__internal__/utils/types.ts
+++ b/packages/blocks/src/__internal__/utils/types.ts
@@ -68,6 +68,7 @@ export type BrushMouseMode = {
 
 export type PanMouseMode = {
   type: 'pan';
+  temporary?: boolean;
 };
 
 export type MouseMode =
--- a/packages/blocks/src/__internal__/utils/types.ts
+++ b/packages/blocks/src/__internal__/utils/types.ts
@@ -68,7 +68,6 @@ export type BrushMouseMode = {
 
 export type PanMouseMode = {
   type: 'pan';
-  temporary?: boolean;
 };
 
 export type MouseMode =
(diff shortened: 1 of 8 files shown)


tests/edgeless.spec.ts
```diff
@@ -341,3 +340,93 @@ test('hovering on shape should not have effect on underlying block', async ({
...
+
+  const start = { x: 100, y: 100 };
+  const end = { x: 200, y: 200 };
+  await addBasicRectShapeElement(page, start, end);
+
+  await page.mouse.move(start.x + 5, start.y + 5);
+  await assertEdgelessHoverRect(page, [100, 100, 100, 100]);
+
+  await page.keyboard.down('Space');
+  const panButton = page
```
doodlewind, line 384: Better use an action to query pan button, seems to be reusable in the case below.

packages/blocks/src/__internal__/utils/types.ts
```diff
@@ -66,7 +66,16 @@ export type BrushMouseMode = {
   lineWidth: number;
 };
 
-export type MouseMode = DefaultMouseMode | ShapeMouseMode | BrushMouseMode;
+export type PanMouseMode = {
+  type: 'pan';
+  temporary?: boolean;
```
doodlewind, line 71: When will this be false? And having it non-optional seems to make more sense 🤔

alt1o: Moved this property to `EdgelessPageBlock`. The scene is when user enters pan mode by pressing 'Space', we should roll back to the last mouse mode once user releases the key;

doodlewind (approved):
LGTM

PR #1454, opened 2023-03-02
refactor(virgo): switch to `rootElement`

Flrande:


doodlewind (commented):
Not sure where is the breaking change? If `editor.getRootElement()` was not used outside of BlockSuite repo, this may not require the `breaking` label 🤔

doodlewind (approved):

packages/virgo/src/virgo.ts
```diff
@@ -17,6 +17,10 @@ export type UpdateVRangeProp = [VRange | null, 'native' | 'input' | 'other'];
 
 export type DeltaEntry = [DeltaInsert, VRange];
 
+// corresponding to [anchorNode/focusNode, anchorOffset/focusOffset]
+export type NativePoint = readonly [Node, number];
```
Saul-Mirone, line 21: You can add label for tuple types, for example:
```ts
export type NativePoint = readonly [node: Node, offset: number];
```

PR #1471, opened 2023-03-03
fix(blocks): missing target when mouse fall into between two blocks and Y axis wrongly calculated

thorseraq:
## Y axis
The  `top` value of Rect calculated by `Element.getBoundingClientRect()`  is neither the same as `PageY` nor `ClientY` in `MouseEvent`.  [It represent the distance from the element top to the top of viewport](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect#scrolling), when the element is upside of the viewport, it's negative, and When the element is downside of the viewport, it's bigger than `window.innerHeight`.

**But when the Element is in the viewPort, the `top` can be compared with `clientY`**. This PR used this point, to first filter only the blocks in the viewPort. Then we can always use `clientY` in `MouseEvent` when comparing.

##  When mouse falls into between two blocks

Currently our binary search uses `y >= detectRect.top && y <= detectRect.bottom` to decide current cursor falls into a block. When mouse fall into between two blocks, the search fails. We think returning the nearest block to the cursor maybe better. This is achieved by the fact the when the search fails, the [`start` and `end`](https://github.com/toeverything/blocksuite/blob/b9719d40ed4572c5621f2d40b4c4e7c75717dc78/packages/blocks/src/page-block/default/utils.ts#L149) is already the indexes of two closest candidate blocks. By comparing the distances between the mouse and two blocks we can get the result.

## Before

https://user-images.githubusercontent.com/20554850/222664255-6ecd9653-1ed1-4b17-911b-caf4d6085bbb.mov

https://user-images.githubusercontent.com/20554850/222664265-bec04651-f09e-42de-a76d-267d00a30bf0.mov

## After

https://user-images.githubusercontent.com/20554850/222666123-32f78214-ae0b-4a53-b046-bfbd6b31be6a.mov

### DIFF
--- a/packages/blocks/src/page-block/default/utils.ts
+++ b/packages/blocks/src/page-block/default/utils.ts
@@ -136,12 +136,22 @@ function binarySearchBlockEditingState(
     containerLeft = firstBlock.blockRect.left;
   }
 
+  let inside = false;
   while (start <= end) {
     const mid = start + Math.floor((end - start) / 2);
     const { block, blockRect, detectRect, hoverDom } = getBlockAndRect(
       blocks,
       mid
     );
+
+    // if the detectRect is not in the view port, it's definitely not the block we want
+    if (detectRect.top > window.innerHeight) {
+      end = mid - 1;
+      continue;
+    } else if (detectRect.bottom < 0) {
+      start = mid + 1;
+      continue;
+    }
 
     // code block use async loading
     if (block.flavour === 'affine:code' && !hoverDom) {
@@ -185,7 +195,7 @@ function binarySearchBlockEditingState(
       }
     }
 
-    const inside = y >= detectRect.top && y <= detectRect.bottom;
+    !inside && (inside = y >= detectRect.top && y <= detectRect.bottom);
 
     if (inside) {
       assertExists(blockRect);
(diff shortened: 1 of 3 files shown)


packages/blocks/src/page-block/default/utils.ts
```diff
@@ -242,6 +252,31 @@ function binarySearchBlockEditingState(
...
+      if (start === blocks.length) {
+        targetIndex = end;
+      } else if (end === -1) {
+        targetIndex = start;
+      } else {
+        const { detectRect: prevDetectRect } = getBlockAndRect(blocks, end);
+        const { detectRect: nextDetectRect } = getBlockAndRect(blocks, start);
+        if (
+          Math.abs(prevDetectRect.bottom - y) < Math.abs(y - nextDetectRect.top)
+        ) {
```
fundon, line 269: It can also be replaced with this: `y < prevDetectRect.bottom + (nextDetectRect.top - prevDetectRect.bottom) / 2`.

packages/blocks/src/page-block/default/utils.ts
```diff
@@ -136,13 +136,23 @@ function binarySearchBlockEditingState(
...
   while (start <= end) {
     const mid = start + Math.floor((end - start) / 2);
     const { block, blockRect, detectRect, hoverDom } = getBlockAndRect(
       blocks,
       mid
     );
 
+    // if the detectRect is not in the view port, it's definitely not the block we want
+    if (detectRect.top > window.innerHeight) {
+      end = mid - 1;
```
fundon, line 149: I think we can add a comment here to use the `container.bottom` and `container.top` to determine if further optimization is needed.

fundon (approved):

packages/blocks/src/page-block/default/utils.ts
thorseraq: Thanks! Updated, get rid of `Math.abs()` now (

packages/blocks/src/page-block/default/utils.ts
thorseraq: Does the `container` here mean `detectRect`?

fundon: `window.innerHeight` and `0`

thorseraq: `clientY` and `detectRect.top` are both relative to viewport, I think `window.innerHeight` and `0` should work here? But I'm not sure (

fundon: Sometimes `.affine-default-viewport` does not fill the entire window.

thorseraq: I think viewport here means  **device viewport** used by `clientY` and `detectRect.top` unless it's embedded in an `iframe`

This contributes to the clarity and ease of use of the current method

fundon: It should be the viewport of the scrolling container.
Screenshot 2023-03-04 at 11 24 39 AM

thorseraq: Got it, I only considered the consistency of two coordinates

Issue #1258, opened 2023-02-17
Remove redundant re-render on mouse move


doodlewind:
Currently BlockSuite will re-render the block tree components on mouse move (this is triggered by `requestUpdate`). But due to the setup of `model` property in Lit block components, the `hasChanged` guards are not functioning properly.

We need a more fine-grained mechanism to incrementally update the affected block components.

lawvs:
Setting the `hasChanged` to always return true is not a correct approach, as it leads to recalculating the template every time the parent node renders the current component.

Adding listeners to the model is a better option, but it is not a long-term solution.

Is there a way to adjust the model to an Immutable model, so that Lit can automatically update on property changes?

```ts
  // set `hasChanged` always return true is incorrect!!
  @property({ hasChanged: () => true })
  model!: PageBlockModel;

  @state()
  private _disposableGroup = new DisposableGroup();

  override connectedCallback() {
    super.connectedCallback();
    this._disposableGroup.add(
      this.model.propsUpdated.on(() => this.requestUpdate())
    );
    this._disposableGroup.add(
      this.model.childrenUpdated.on(() => this.requestUpdate())
    );
  }

  override disconnectedCallback() {
    super.disconnectedCallback();
    this._disposableGroup.dispose();
  }
```

pengx17:
> Is there a way to adjust the model to an Immutable model, so that Lit can automatically update on property changes?

I think one solution may be to use smaller lit components, instead of using rendering functions? E.g. for `DefaultPageBlockComponent`, it has several big rendering functions in `render`:

```js
  render() {
    const { page, selection } = this;
    const { viewport } = selection.state;

    const childrenContainer = BlockChildrenContainer(this.model, this, () =>
      this.requestUpdate()
    );
    const draggingArea = DraggingArea(this._draggingArea);
    const selectedRectsContainer = SelectedRectsContainer(
      this._selectedRects,
      viewport
    );
    const selectedEmbedContainer = EmbedSelectedRectsContainer(
      this._selectedEmbedRects,
      viewport
    );
    const embedEditingContainer = EmbedEditingContainer(
      page.readonly ? null : this._embedEditingState,
      this.slots,
      viewport
    );
    const codeBlockOptionContainer = CodeBlockOptionContainer(
      page.readonly ? null : this._codeBlockOption
    );

    return html`
      <div class="affine-default-viewport">
        <div class="affine-default-page-block-container">
          <div class="affine-default-page-block-title-container">
            <div
[...]

PR #1472, opened 2023-03-03
perf: reduce rich text re-render

doodlewind:
Related to #1258

Mouse move before (using http://localhost:5173/?init=preset):

Mouse move after:


pengx17 (approved):
LGTM

PR #1470, opened 2023-03-03
fix(playground): edgeless blank

alt1o:
https://user-images.githubusercontent.com/21084335/222624960-c7865cc6-91d0-45e7-863f-1c531bd35f7e.mov

PR #1478, opened 2023-03-03
fix: should diff before updating

fundon:
On the heavy page, the drag-handle position moves rather slowly, it is mainly caused by the following:
https://github.com/toeverything/blocksuite/blob/75ac3784b82eab62bfb4b34b010afeed36f4b54b/packages/blocks/src/page-block/default/default-page-block.ts#L488-L495

https://user-images.githubusercontent.com/27926/222783312-7c31b9ba-ea14-4e20-8539-67f7f4f67bb6.mov


doodlewind (commented):
I tend to hold this until #1470

PR #1481, opened 2023-03-03
refactor: `Signal` -> `Slot`

doodlewind:
This avoids the ambiguity with the recent `Signal` concept, making room for our future potential reactivity primitives.

Note that this doesn't affect the `AbortController.signal`, which also reduces ambiguity.


doodlewind:
@lawvs still having the status issue, unable to auto merge :(

PR #1482, opened 2023-03-04
feat: hide title placeholder in ime

zqran:
before:
image

after:
image

### DIFF
--- a/packages/blocks/src/page-block/default/default-page-block.ts
+++ b/packages/blocks/src/page-block/default/default-page-block.ts
@@ -220,6 +220,10 @@ export class DefaultPageBlockComponent
 
   private _onTitleKeyDown = (e: KeyboardEvent) => {
     if (e.isComposing) {
+      this.page.workspace.setPageMeta(this.page.id, {
+        title: '',
+      });
+      this.requestUpdate();
       return;
     }
     const hasContent = !this.page.isEmpty;


lawvs (changes requested):
The issue occurred because Lit was not updated in a timely manner. Perhaps a different approach should have been taken.

https://user-images.githubusercontent.com/18554747/222875463-fc2b1825-08d7-4430-b901-a83d490a0947.mov

lawvs:
Also related to https://github.com/toeverything/blocksuite/issues/1258

Flrande:
The placeholder just a pseudo-element from `:before` and you don't need to control meta data for it.

zqran:
> The issue occurred because Lit was not updated in a timely manner. Perhaps a different approach should have been taken.

Yes, I also found out that the placeholder for the title disappears as soon as the mouse is moved

zqran:
> The placeholder just a pseudo-element from `::before` and you don't need to control meta data for it.

I just want to trigger the title change in time and hide the placeholder, so that the UX is better

zqran:
@lawvs Why do we need to re-render the block through mouse move?

lawvs:
> @lawvs Why do we need to re-render the block through mouse move(mentioned in https://github.com/toeverything/blocksuite/issues/1258)?

It's just a bug

zqran:
closed, see #1556

PR #1484, opened 2023-03-04
fix: should cancel when select selected point outside the selected element

PerfectPan:
Fix the following problem(when we select the first elipse, move to the second eplise, and try to drag on the second eplise, the first elipse is dragged):

https://user-images.githubusercontent.com/24316656/222884616-767a90c4-ed39-457f-8567-83ecc00738bd.mp4


doodlewind (approved):
Good catch, thanks for your contribution!

Issue #1363, opened 2023-02-25
Edgeless mode code block can not switch language


PerfectPan:
The language picker does not show up in edgeless mode.

I see the language picker hidden in some cases(https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/code-block/code-block.ts#L242).

PR #1486, opened 2023-03-04
refactor: simplify code block language list logic

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/1363

### DIFF
--- a/packages/blocks/src/__internal__/utils/components.ts
+++ b/packages/blocks/src/__internal__/utils/components.ts
@@ -1,11 +1,14 @@
 /* eslint-disable lit/binding-positions, lit/no-invalid-html */
 import '../../components/loader.js';
 
-import { BLOCK_CHILDREN_CONTAINER_PADDING_LEFT } from '@blocksuite/global/config';
+import {
+  BLOCK_CHILDREN_CONTAINER_PADDING_LEFT,
+  BLOCK_ID_ATTR,
+} from '@blocksuite/global/config';
 import type { BaseBlockModel } from '@blocksuite/store';
 import { matchFlavours } from '@blocksuite/store';
 import { repeat } from 'lit/directives/repeat.js';
-import { html } from 'lit/static-html.js';
+import { html, unsafeStatic } from 'lit/static-html.js';
 
 import type { EmbedBlockModel } from '../../embed-block/index.js';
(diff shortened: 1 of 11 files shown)


packages/blocks/src/__internal__/utils/components.ts
```diff
@@ -1,11 +1,14 @@
...
-import { BLOCK_CHILDREN_CONTAINER_PADDING_LEFT } from '@blocksuite/global/config';
+import {
+  BLOCK_CHILDREN_CONTAINER_PADDING_LEFT,
+  BLOCK_ID_ATTR,
+} from '@blocksuite/global/config';
 import type { BaseBlockModel } from '@blocksuite/store';
 import { matchFlavours } from '@blocksuite/store';
 import { repeat } from 'lit/directives/repeat.js';
-import { html } from 'lit/static-html.js';
+import { html, unsafeStatic } from 'lit/static-html.js';
```
doodlewind, line 11: Is it possible to replace this with the `literal` directive?

doodlewind (commented):
Are there some goals or related issues about the refactoring?

packages/blocks/src/__internal__/utils/components.ts
lawvs: > Error: Value passed to 'literal' function must be a 'literal' result: data-block-id. Use 'unsafeStatic' to pass non-literal values, but
            take care to ensure page security.

It's not work, because tags a string literal so that it behaves like part of the **static template strings** instead of a dynamic value. [ref](https://lit.dev/docs/api/static-html/#literal)

At this location, the unsafe template we're using only covers static strings inside the code, so it's security. We could directly use the string "data-block-id", but this would lose the reference to `BLOCK_ID_ATTR`.
```jsx
      return html`
        <${model.tag}
          .model=${model}
          .host=${host}
          data-block-id=${model.id}
        ></${model.tag}>
      `;
```

lawvs:
> Are there some goals or related issues about the refactoring? It may be hard to tell what happened if we just add `refactor: code block` into our release log.

I intend to extract the `CodeBlockOption` from both the `selection-manager` and `default-page-block`.

However, prior to commencing the actual work, these changes alone will suffice for submitting a comprehensive PR.

Issue #1485, opened 2023-03-04
Hovering the frame selection over a shape will cause the shape to move along with it


PerfectPan:
The problematic code is https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/page-block/edgeless/mode-controllers/default-mode.ts#L196.

https://user-images.githubusercontent.com/24316656/222885279-3b548953-b1b2-4045-aeb7-e9606bc4c273.mp4

The shape element should not move when the selected state is inactive.

However, using frame selection over two elements is not currently supported in this mode. We can only select one element at a time. So, it may be better to solve the problems together.

doodlewind:
Should be closed by #1488, thanks for your feedback!

PR #1488, opened 2023-03-04
feat: shape element should not move when the selected state is inactive

zqran:
Related to #1485

fix:
> The shape element should not move when the selected state is inactive.


doodlewind (approved):
Good catch, thanks!

PR #1490, opened 2023-03-04
fix: remove valueInitializer in workspace meta

himself65:


himself65:
Related: https://github.com/toeverything/AFFiNE/pull/1296

Issue #1480, opened 2023-03-03
Refactor default selection manager


doodlewind:
We should consider re-organize its code structure, since the maintainability is concerning.

Before refactoring lands, recent features, optimizations and non-incremental fixes on selection manager may better be postponed.

/cc @fundon @lawvs @Himself65 

Progress
* #1494 
* #1496 
* #1500 
* #1502
* #1503 
* #1506 
* #1507 
* #1511 
* #1512 
* #1519

fundon:
Here are some of my thoughts:

1. Binary search may not be needed, use `document.elementFromPoint` instead of
2. Block cache may not be needed either
3. Grouping dom manipulations into a single file
4. For view updates, execute `requestUpdate` all at once after batching

lawvs:
I intend to extract the `CodeBlockOption` and `EmbedEditingState` from both the `selection-manager` and `default-page-block`.

These two parts should be controlled internally by the corresponding block.

![Screenshot 2023-03-05 at 2 42 32 AM](https://user-images.githubusercontent.com/18554747/222923322-72599ed0-b60e-4037-8d70-a435ac271623.png)

doodlewind:
> I intend to extract the `CodeBlockOption` and `EmbedEditingState` from both the `selection-manager` and `default-page-block`.
> 
> 
> 
> These two parts should be controlled internally by the corresponding block.
> 
> 
> 
> ![Screenshot 2023-03-05 at 2 42 32 AM](https://user-images.githubusercontent.com/18554747/222923322-72599ed0-b60e-4037-8d70-a435ac271623.png)
> 
> 

Sounds fair, is there ongoing PR?

doodlewind:
Considered completed in 0.5

PR #1512, opened 2023-03-06
refactor: add `selection.state.draggingArea`

fundon:
part of #1480

### DIFF
--- a/packages/blocks/src/__internal__/utils/rect.ts
+++ b/packages/blocks/src/__internal__/utils/rect.ts
@@ -1,3 +1,5 @@
+import { clamp } from '../index.js';
+
 export class Point {
   public x: number;
   public y: number;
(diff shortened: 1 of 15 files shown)


fundon:
It would be better to unify `Point` `Rect`.

e.g. `IPoint`, `IBound`, too many types.

packages/blocks/src/__internal__/utils/rect.ts
```diff
@@ -1,3 +1,5 @@
+import { clamp } from '../index.js';
+
 export class Point {
```
himself65, line 3: I think we need unit test for the basic class

himself65: please use vitest with happy-dom

doodlewind: This class implementation looks trivial, we have added tests in `tl-utils` in phasor. So I suggest not to do that in this PR.

doodlewind (approved):

PR #1550, opened 2023-03-08
perf: remove redundant `hasChanged`

### LINKED ISSUE #1258: Remove redundant re-render on mouse move

doodlewind:
Close #1258

### DIFF
--- a/packages/blocks/src/page-block/default/default-page-block.ts
+++ b/packages/blocks/src/page-block/default/default-page-block.ts
@@ -443,6 +443,8 @@ export class DefaultPageBlockComponent
       if (flag) window.addEventListener('keydown', this._handleNativeKeydown);
       else window.removeEventListener('keydown', this._handleNativeKeydown);
     });
+
+    this.model.childrenUpdated.on(() => this.requestUpdate());
   }
 
   private _initFrameSizeEffect() {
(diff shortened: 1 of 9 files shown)


pengx17:
ops, I pushed some debug code, sorry

packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -443,11 +443,6 @@ export class DefaultPageBlockComponent
       if (flag) window.addEventListener('keydown', this._handleNativeKeydown);
       else window.removeEventListener('keydown', this._handleNativeKeydown);
     });
-
-    this.model.page.slots.blockUpdated.on(() => {
```
pengx17, line 447: I think we still need this to make sure programmatic adding new blocks will work as expected, as stated in PR summary in https://github.com/toeverything/blocksuite/pull/1520

doodlewind: The `page.blockUpdated` triggers at every nested block update, which may be too coarse-grained. So I tried to use `model.childrenUpdated` event instead, which only triggers at shallow children update.

Issue #1551, opened 2023-03-08
Flaky test cases that could pass on CI but fail locally


doodlewind:
Example: The `should keyboard work in link popover` test case.

doodlewind:
Presumably fixed by #1573

PR #1559, opened 2023-03-08
fix: placeholder with IME

lawvs:
https://github.com/toeverything/blocksuite/pull/1482

Revert https://github.com/toeverything/blocksuite/pull/1556

### DIFF
--- a/packages/blocks/src/paragraph-block/paragraph-block.ts
+++ b/packages/blocks/src/paragraph-block/paragraph-block.ts
@@ -157,9 +157,6 @@ export class ParagraphBlockComponent extends NonShadowLitElement {
   @state()
   private _showTipsPlaceholder = false;
 
-  @state()
-  private _isComposing = false;
-
   private _disposables = new DisposableGroup();
 
   firstUpdated() {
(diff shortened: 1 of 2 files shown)


packages/blocks/src/paragraph-block/paragraph-block.ts
```diff
@@ -172,12 +175,25 @@ export class ParagraphBlockComponent extends NonShadowLitElement {
...
-        this.model.text.length === 0 && this.model.type === 'text';
+        this.model.text.length === 0 &&
+        this.model.type === 'text' &&
+        !this._isComposing;
     };
     this.model.text.yText.observe(observer);
     this._disposables = new DisposableGroup();
     this._disposables.add(this.model.propsUpdated.on(observer));
     this._disposables.add(() => this.model.text.yText.unobserve(observer));
+    // Workaround for virgo skips composition event
```
doodlewind, line 186: This may be also required for list block, code block and other Virgo-based components. Any idea to avoid boilerplate?

lawvs: Another block does not have placeholder, so it is okay.

We can handle this scenario in `rich-text` when the list block needs a placeholder.

doodlewind:
The CI issue looks related to #1551

doodlewind (approved):
Verified that this fix works!

PR #1583, opened 2023-03-08
fix(blocks): update paragraph type with new vRange

donaldxdonald:


Flrande:
You can use [setVRange](https://github.com/toeverything/blocksuite/blob/53148c19102026e48d18b1cde14944bd469b65d1/packages/virgo/src/virgo.ts#L506).

Flrande (approved):

PR #1587, opened 2023-03-09
fix: fix the console error when dragging the dividing line

zqran:
error message:
image

The dividing line does not contain text content by default.


lawvs:
In what scenario will this error occur? Should we select this dividing line?

zqran:
> In what scenario will this error occur? Should we select this dividing line?

I uploaded a reproduced gif please see the first comment.

PR #1604, opened 2023-03-09
fix: allow scroll to view at need

lawvs:
Fix https://github.com/toeverything/blocksuite/pull/1582


zqran:
😥 , I didn't consider this scenario. You are awesome.

PR #1613, opened 2023-03-09
refactor: remove the deprecated `addBlock` api

Saul-Mirone:


doodlewind (approved):
I have a bold idea, how about we rename `addBlockByFlavour` to `addBlock` in the next PR 😈

Saul-Mirone:
> I have a bold idea, how about we rename `addBlockByFlavour` to `addBlock` in the next PR 😈

I totally agree 😈.

PR #1621, opened 2023-03-09
refactor: add async query for dom mutation scenario

Saul-Mirone:
This PR adds async version of query functions to avoid the ambiguity `requestAnimationFrame` call when updating dom.

### DIFF
--- a/packages/blocks/src/__internal__/utils/common-operations.ts
+++ b/packages/blocks/src/__internal__/utils/common-operations.ts
@@ -96,6 +96,7 @@ export function convertToList(
 
     model.text?.delete(0, prefix.length + 1);
     page.updateBlock(model, { type: listType });
+    asyncFocusRichText(page, model.id);
   }
   return true;
 }
--- a/packages/blocks/src/__internal__/utils/common-operations.ts
+++ b/packages/blocks/src/__internal__/utils/common-operations.ts
@@ -6,14 +6,9 @@ import type { VEditor, VRange } from '@blocksuite/virgo';
 import { asyncGetRichTextByModel, getRichTextByModel } from './query.js';
 import type { ExtendedModel } from './types.js';
 
-export function asyncSetVRange(model: BaseBlockModel, vRange: VRange) {
-  asyncGetRichTextByModel(model)
-    .then(richText => {
-      richText?.vEditor?.setVRange(vRange);
-    })
-    .catch(e => {
-      throw e;
-    });
+export async function asyncSetVRange(model: BaseBlockModel, vRange: VRange) {
+  const richText = await asyncGetRichTextByModel(model);
+  richText?.vEditor?.setVRange(vRange);
 }
 
 export function asyncFocusRichText(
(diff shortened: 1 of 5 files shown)


lawvs (commented):
I try to deprecate the concept of `richText`. Therefore, I think `asyncSetVRangeForVirgo` or `asyncSetVRange` would be better.

See https://github.com/toeverything/blocksuite/pull/1626

Saul-Mirone:
> I try to deprecate the concept of `richText`. Therefore, I think `asyncSetVRangeForVirgo` or `asyncSetVRange` would be better.
> 
> See #1626

I prefer to do the renaming of `RichText` in another PR because I think they're breaking changes. And it will also make the changes easier to read.

packages/blocks/src/__internal__/utils/common-operations.ts
```diff
@@ -1,32 +1,30 @@
...
 import type { BaseBlockModel } from '@blocksuite/store';
 import type { VEditor, VRange } from '@blocksuite/virgo';
 
-import { getRichTextByModel } from './query.js';
+import { asyncGetRichTextByModel, getRichTextByModel } from './query.js';
 import type { ExtendedModel } from './types.js';
 
+export function asyncSetVRange(model: BaseBlockModel, vRange: VRange) {
+  asyncGetRichTextByModel(model)
+    .then(richText => {
```
zqran, line 11: Can `async/await` syntax be used here, I think the code format will be more uniform

Saul-Mirone: I didn't use the asyn function here is because I want to handle the promise error in the function and make sure when it is called, users won't need to handle it manually.

zqran: I think it is equivalent to write like this:

```ts
async function asyncSetVRange(model: BaseBlockModel, vRange: VRange) {
  try {
    const richText = await asyncGetRichTextByModel(model)
    richText?.vEditor?.setVRange(vRange);
  } catch (e) {
    throw e;
  }
}
```

doodlewind (approved):
Great work!

packages/blocks/src/__internal__/utils/common-operations.ts
doodlewind: This try/catch looks redundant.

Saul-Mirone: > I think it is equivalent to write like this:
> 
> 
> 
> ```ts
> 
> async function asyncSetVRange(model: BaseBlockModel, vRange: VRange) {
> 
>   try {
> 
>     const richText = await asyncGetRichTextByModel(model)
> 
>     richText?.vEditor?.setVRange(vRange);
> 
>   } catch (e) {
> 
>     throw e;
> 
>   }
> 
> }
> 
> ```

IMO they won't be equal. When you're using try catch in an async function, it will reject the error, and if the place where it called doesn't catch it, it will be a floating promise error. But, nevermind, it's not that important in this case.

PR #1624, opened 2023-03-09
fix: press enter twice at end of code block with content can jump out

zqran:


doodlewind:
I think the current implementation matches the product design, closing it for now.

zqran:
> I think the current implementation matches the product design, closing it for now.

The expected behavior is: 
after pressing the enter key **twice**, jump out of the code block(mplemented by this PR).

But the current behavior is: 
after pressing the enter key **three times**, jump out of the code block.

I think it may be that `Quill` and `Virgo` get different text content.

Flrande (approved):

PR #1625, opened 2023-03-09
fix: date & time string can not be inserted using the slash menu

PerfectPan:
The issue is that vEditor does not synchronize its state vRange when deleting search string.

https://user-images.githubusercontent.com/24316656/224102751-1d580025-d50c-450d-b43f-a51b9b71f7e0.mp4


lawvs (approved):
Great job!

Flrande (approved):

PR #1626, opened 2023-03-09
refactor: replace `getRichTextByModel` with `getVirgoByModel`

lawvs:
Replace `getRichTextByModel` with `getVirgoByModel`.

Update the following pattern

```ts
// Before
  const richText = getRichTextByModel(model);
  assertExists(richText);
  const { vEditor } = richText;
  assertExists(vEditor);

// After
  const vEditor = getVirgoByModel(model);
  assertExists(vEditor);
```

### DIFF
--- a/packages/blocks/src/__internal__/utils/query.ts
+++ b/packages/blocks/src/__internal__/utils/query.ts
@@ -313,6 +313,9 @@ export function getStartModelBySelection(range = getCurrentNativeRange()) {
   return startModel;
 }
 
+/**
+ * @deprecated In most cases, you not need RichText, you can use {@link getVirgoEditorByModel} instead.
+ */
 export function getRichTextByModel(model: BaseBlockModel) {
   const blockElement = getBlockElementByModel(model);
   const richText = blockElement?.querySelector<RichText>('rich-text');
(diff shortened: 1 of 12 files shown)


packages/blocks/src/__internal__/utils/query.ts
```diff
@@ -256,13 +256,23 @@ export function getStartModelBySelection(range = getCurrentNativeRange()) {
   return startModel;
 }
 
+/**
+ * @deprecated In most cases, you not need RichText, you can use {@link getVirgoEditorByModel} instead.
```
zqran, line 260:
```suggestion
 * @deprecated In most cases, you not need RichText, you can use {@link getVirgoByModel} instead.
```
Does this name need to be consistent with the corresponding function name?

packages/blocks/src/__internal__/utils/query.ts
```diff
@@ -256,13 +256,23 @@ export function getStartModelBySelection(range = getCurrentNativeRange()) {
...
 export function getRichTextByModel(model: BaseBlockModel) {
   const blockElement = getBlockElementByModel(model);
   const richText = blockElement?.querySelector('rich-text') as RichText;
   if (!richText) return null;
   return richText;
 }
 
+export function getVirgoByModel(model: BaseBlockModel) {
+  const blockElement = getBlockElementByModel(model);
+  const richText = blockElement?.querySelector('rich-text') as RichText;
```
Saul-Mirone, line 271: You can use type parameter to avoid the type cast.

```ts
querySelector<RichText>('rich-text')
```

doodlewind (approved):

Issue #1553, opened 2023-03-08
Text styles should be kept when using CJK IME


lawvs:
https://user-images.githubusercontent.com/18554747/223621007-9354a9fc-2f16-46ef-8e7e-77a41db70cfd.mov

lawvs:
Reopen due to the introduction of a new issue

See https://github.com/toeverything/blocksuite/pull/1636#issuecomment-1463442683

PR #1636, opened 2023-03-10
fix: text styles when using CJK IME

zqran:
close: #1553

### DIFF
--- a/packages/blocks/src/__internal__/rich-text/rich-text.ts
+++ b/packages/blocks/src/__internal__/rich-text/rich-text.ts
@@ -112,13 +112,10 @@ export class RichText extends NonShadowLitElement {
           const delta = vEditor.getDeltaByRangeIndex(index);
           const attributes = delta?.attributes ?? {};
           vEditor.insertText(vRange, data, attributes);
-          vEditor.slots.updateVRange.emit([
-            {
-              index: index + data.length,
-              length: 0,
-            },
-            'input',
-          ]);
+          vEditor.setVRange({
+            index: index + data.length,
+            length: 0,
+          });
           return true;
         }
         return false;
--- a/packages/blocks/src/__internal__/rich-text/rich-text.ts
+++ b/packages/blocks/src/__internal__/rich-text/rich-text.ts
@@ -98,6 +98,31 @@ export class RichText extends NonShadowLitElement {
 
         return false;
       },
+      virgoCompositionEnd: e => {
+        const { data } = e;
+        const vEditor = this._vEditor;
+        assertExists(vEditor);
+        const vRange = vEditor.getVRange();
+        if (!vRange || vRange.length !== 0) {
+          return false;
+        }
+
+        const index = vRange.index;
+        if (index >= 0 && data) {
+          const delta = vEditor.getDeltaByRangeIndex(index);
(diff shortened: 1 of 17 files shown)


Flrande:
Currently, all text is inserted without attributes by default, and if we wish to maintain formatting, we must handle it externally, such as using the [virgoInput handler](https://github.com/toeverything/blocksuite/blob/29f3a9a1cf18f592cb4427b1a78d5e2432abd8a5/packages/blocks/src/__internal__/rich-text/rich-text.ts#L72). (The changes in this pr now will break this design)

To address this issue in a more reasonable manner, I propose adding handlers for IME, such as `virgoCompositionStart` and `virgoCompositionEnd`. By doing so, we can keep the system simple and customizable. :)

Flrande (changes requested):
https://github.com/toeverything/blocksuite/pull/1636#issuecomment-1463205297

packages/blocks/src/__internal__/rich-text/rich-text.ts
```diff
@@ -98,6 +98,31 @@ export class RichText extends NonShadowLitElement {
...
+        if (!vRange || vRange.length !== 0) {
+          return false;
+        }
+
+        const index = vRange.index;
+        if (index >= 0 && data) {
+          const delta = vEditor.getDeltaByRangeIndex(index);
+          const attributes = delta?.attributes ?? {};
+          vEditor.insertText(vRange, data, attributes);
+          vEditor.slots.updateVRange.emit([
```
Flrande, line 115: You can use `setVRange` here,

zqran:
> Currently, all text is inserted without attributes by default, and if we wish to maintain formatting, we must handle it externally, such as using the [virgoInput handler](https://github.com/toeverything/blocksuite/blob/29f3a9a1cf18f592cb4427b1a78d5e2432abd8a5/packages/blocks/src/__internal__/rich-text/rich-text.ts#L72). (The changes in this pr now will break this design)
> 
> To address this issue in a more reasonable manner, I propose adding handlers for IME, such as `virgoCompositionStart` and `virgoCompositionEnd`. By doing so, we can keep the system simple and customizable. :)

Thanks for the detailed explanation, such a design is awesome.

Flrande (changes requested):
I recommend using `setVRange`

packages/blocks/src/__internal__/rich-text/rich-text.ts
zqran: 👌

Flrande (approved):

Flrande:
Thank for your work!

lawvs:
This incorrect fix will result in inconsistent input states between using and not using IME.

https://user-images.githubusercontent.com/18554747/224262789-ceedf814-9a16-4cd9-9559-5a25fa1d974f.mov

Flrande:
It seem that it need to compute conditional branch like `virgoInput`. 🤔

zqran:
I'll try to fix it.

Flrande:
> I'll try to fix it.

https://github.com/toeverything/blocksuite/blob/a86b71fbe4dab0a34d87e62df297163f5949a651/packages/blocks/src/__internal__/rich-text/rich-text.ts#L82-L89
Maybe it can help you.

PR #1644, opened 2023-03-10
fix(virgo): synchronizing selection can potentially result in an infinite loop

Flrande:


packages/virgo/src/virgo.ts
```diff
@@ -1019,16 +1012,23 @@ export class VEditor<
...
 
     if (
-      range.startContainer.nodeType !== Node.TEXT_NODE ||
-      range.endContainer.nodeType !== Node.TEXT_NODE
+      (range.startContainer.nodeType !== Node.TEXT_NODE ||
+        range.endContainer.nodeType !== Node.TEXT_NODE) &&
+      range.startContainer !== this._previousAnchor[0] &&
+      range.endContainer !== this._previousFocus[0] &&
+      range.startOffset !== this._previousAnchor[1] &&
+      range.endOffset !== this._previousFocus[1]
```
lawvs, line 1029: You should add some comments to explain the changes

PR #1661, opened 2023-03-11
fix(virgo): fix dbclick in first word

Flrande:


packages/virgo/src/components/virgo-line.ts
```diff
@@ -25,13 +25,12 @@ export class VirgoLine<
     return result;
   }
 
+  protected firstUpdated(): void {
+    this.style.display = 'block';
```
lawvs, line 29: Can we setting it on `static style`?

Flrande: I don't know how to construct a css in static css to handle `v-line`.

PR #1680, opened 2023-03-13
refactor(code): replace `setTimeout` with `ResizeObserver`

Flrande:


packages/blocks/src/code-block/code-block.ts
```diff
@@ -287,6 +286,11 @@ export class CodeBlockComponent extends NonShadowLitElement {
       assertExists(vEditor);
       vEditor.requestUpdate();
     }
+
+    const richText = this.querySelector('rich-text');
+    assertExists(richText);
+    this._richTextResizeObserver.disconnect();
```
lawvs, line 292: I think you should also call `this._richTextResizeObserver.disconnect();` in `Lit.disconnect`

lawvs (approved):

doodlewind (approved):

PR #1682, opened 2023-03-13
refactor(blocks): improve code-block highlight performance

### LINKED ISSUE #1568: Highlighter in code-block load too slow
The highlight will not be rendered until all the highlight files are loaded in the current code block. This will delay the time when the highlight is first generated. We need to load the selected language first.

image

Flrande:
Close #1568

### DIFF
--- a/packages/blocks/src/code-block/code-block.ts
+++ b/packages/blocks/src/code-block/code-block.ts
@@ -251,6 +251,7 @@ export class CodeBlockComponent extends NonShadowLitElement {
     super.disconnectedCallback();
     this._disposables.dispose();
     this.hoverState.dispose();
+    this._richTextResizeObserver.disconnect();
   }
 
   private _onClickWrapBtn() {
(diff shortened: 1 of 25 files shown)


packages/blocks/src/code-block/code-block.ts
```diff
@@ -261,32 +261,53 @@ export class CodeBlockComponent extends NonShadowLitElement {
...
+                vEditor.requestUpdate();
+              }
+            });
+          }
+        } else {
+          this._startHighlight([lang]);
+        }
+      } else {
+        this._highlighter = null;
+      }
```
lawvs, line 298: Try to optimize this logic

packages/blocks/src/code-block/code-block.ts
```diff
@@ -167,12 +167,11 @@ export class CodeBlockComponent extends NonShadowLitElement {
...
-  private langUpdated = new Slot<{
-    lang: Lang;
-    highlighter: Highlighter;
-  }>();
+  private _richTextResizeObserver: ResizeObserver = new ResizeObserver(() => {
+    this._updateLineNumbers();
+  });
 
-  private _preLang: Lang | null = null;
+  private _preLang: string | null = null;
```
lawvs, line 174: Is this operation safe?

Flrande: I think so. We cannot always assume that this property is legal.

PR #1688, opened 2023-03-13
fix(editor): some safari compatibility issues

pengx17:
This PR fix a few Safari compatibility issues found in Safari test, mainly related to `prop:xywh` inconsistency when comparing yDoc json and make sure the virgo tests can work in Safari.

### DIFF
--- a/packages/blocks/src/__internal__/utils/components.ts
+++ b/packages/blocks/src/__internal__/utils/components.ts
@@ -33,6 +33,7 @@ export function BlockElement(
         <${model.tag}
           .model=${model}
           .host=${host}
+          class="affine-block-element"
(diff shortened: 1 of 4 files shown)


packages/blocks/src/__internal__/utils/components.ts
```diff
@@ -33,6 +33,7 @@ export function BlockElement(
         <${model.tag}
           .model=${model}
           .host=${host}
+          class="affine-block-element"
```
pengx17, line 36: We have `getBoundingClientRect` to get the block element size which will be used for calculating the xywh prop for elements, however this API returns different values between block vs inline elements in Safari.
Making block elements to `display:block` fix this issue.
See
image
image

Flrande (approved):

PR #1698, opened 2023-03-14
fix: code block caret position

zqran:
before:

https://user-images.githubusercontent.com/15389209/224984772-e24b7e90-0a66-4954-881d-5558e67e8c4b.mov

after:

https://user-images.githubusercontent.com/15389209/224984804-85607530-d48b-4fbd-bb8a-64ff3ee4eb5c.mov


doodlewind:
> @QiShaoXuan Could you check if other blocks also have the same issue?

Didn't found regression for paragraph, this looks good to merge.

PR #1703, opened 2023-03-14
feat(edgeless): component toolbar

### LINKED ISSUE #1701: Support whiteboard component toolbar
This locates a toolbar positioned above the selected shape, brush or frame.

### LINKED ISSUE #1733: the scaling label font of the whiteboard global toolbar is not consistent with the editor
I found that the scaling label font of the whiteboard global toolbar is not consistent with the editor. Does this detail need to be unified? @joebeijing @alt1o  If so, is the default font configured in the editor global

before consistent

image

after consistent

CleanShot 2023-03-16 at 18 12 24@2x

alt1o:
fix #1701 
fix #1733

https://user-images.githubusercontent.com/21084335/225603272-1c4bdbf6-b95c-44cc-b6a8-e78ad5cfe6de.mov


packages/phasor/src/utils/std.ts
```diff
@@ -0,0 +1,11 @@
+export function simplePick<T extends object, K extends keyof T>(
```
doodlewind, line 1: Admittedly couldn't find a better place for this

packages/blocks/src/page-block/edgeless/components/component-toolbar/more-button.ts
```diff
@@ -0,0 +1,158 @@
...
+      if (isTopLevelBlock(element)) {
+        const children = this.page.root?.children ?? [];
+        if (children.length > 1) {
+          this.page.deleteBlock(element);
+        }
+      } else {
+        this.surface.removeElement(element.id);
+      }
+    });
+    this.page.captureSync();
```
doodlewind, line 121: I don't think we should do `captureSync()` twice. Having it before operations should be adequate.

doodlewind:
CI only fails at uploading Codecov, good to merge.

PR #1706, opened 2023-03-14
refactor(selection): keep setSelectedBlocks logic simple

fundon:
1. click a image, `selection.state.type` switch to `embed`
2. drag-and-drop image block, `selection.state.type` switch to `block`

The following methods are all for `block`:
https://github.com/toeverything/blocksuite/blob/9df30f433795a314367a271526c66b106b149d00/packages/blocks/src/page-block/default/selection-manager/utils.ts#L144-L148

I think we can set the state after the drag and drop to block first to avoid confusion.


doodlewind (approved):

doodlewind:
Thanks for the fix!

PR #1711, opened 2023-03-15
feat(virgo): support marks

### LINKED ISSUE #1456: Support cursor with format in virgo
Due to the lack of support for quill in cursor marks, we utilized `quill.on` to achieve our need for inline code escaping. 

However, this solution caused several issues, it caused the status of quill and Ytext to be out of sync.

As a result, we hope that the virgo can support cursor marks.

This is similar to the concept of a Mark in prosemirror

Related to https://github.com/toeverything/blocksuite/issues/415 https://github.com/toeverything/blocksuite/issues/1319 https://github.com/toeverything/blocksuite/issues/1184

Flrande:
Close #1456

### DIFF
--- a/packages/virgo/src/virgo.ts
+++ b/packages/virgo/src/virgo.ts
@@ -200,6 +200,7 @@ export class VEditor<
   private _isComposing = false;
   private _isReadonly = false;
   private _yText: Y.Text;
+  private _marks: TextAttributes | null = null;
 
   private _previousAnchor: NativePoint | null = null;
   private _previousFocus: NativePoint | null = null;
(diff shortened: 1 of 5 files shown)


packages/virgo/src/virgo.ts
```diff
@@ -536,6 +541,29 @@ export class VEditor<
...
+        !vRange ||
+        !r ||
+        r.length !== 0 ||
+        (r.index !== vRange.index && r.index !== vRange.index + 1)
+      ) {
+        this.resetMarks();
+        dispose.dispose();
+      } else {
+        vRange = r;
+      }
```
lawvs, line 559:
```suggestion
      }
```

Flrande: Why?

Issue #1705, opened 2023-03-14
Support block hub in edgeless mode


doodlewind:
* Support dragging blocks from block hub into existing frames.
* Support dragging blocks from block hub to blank area.

doodlewind:
Closed by #1712, thanks for the contribution by @fundon!

PR #1712, opened 2023-03-15
feat(block-hub): support in edgeless mode

fundon:
Related to #1705 

* [x] Support dragging blocks from block hub into existing frames.
* [x] Support dragging blocks from block hub to blank area.
* [x] tests

### DIFF
--- a/packages/blocks/src/page-block/edgeless/mode-controllers/default-mode.ts
+++ b/packages/blocks/src/page-block/edgeless/mode-controllers/default-mode.ts
@@ -10,7 +10,6 @@ import {
   getSelectedStateRectByBlockElement,
   handleNativeRangeClick,
   handleNativeRangeDragMove,
-  isEmpty,
   noop,
   Point,
   resetNativeSelection,
@@ -180,7 +179,7 @@ export class DefaultModeController extends MouseModeController<DefaultMouseMode>
   }
 
   private _tryDeleteEmptyBlocks() {
-    const emptyBlocks = this._blocks.filter(b => isEmpty(b));
+    const emptyBlocks = this._blocks.filter(b => !b.children.length);
     // always keep at least one frame block
     if (emptyBlocks.length === this._blocks.length) {
       emptyBlocks.shift();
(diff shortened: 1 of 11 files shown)


fundon:
@doodlewind In which mode is `block-hub` enabled, dragged and dropped?

doodlewind (commented):
Found the only failing case is the regression in the `drag database` case. Can @zqran help taking a look?

zqran:
> Found the only failing case is the regression in the `drag database` case. Can @zqran help taking a look?

ok, let me check

packages/editor/src/utils/editor.ts
```diff
@@ -25,53 +27,104 @@ export const createBlockHub: (
...
-    assertExists(defaultPageBlock);
-    blockHub.slots = defaultPageBlock.slots;
-    blockHub.getAllowedBlocks = () =>
-      getAllowSelectedBlocks(defaultPageBlock.model);
-  } else {
-    const edgelessPageBlock = editor.querySelector('affine-edgeless-page');
-    assertExists(edgelessPageBlock);
-    blockHub.getAllowedBlocks = () =>
-      getAllowSelectedBlocks(edgelessPageBlock.model);
-  }
```
fundon, line 74: This code will only be executed once. 

May need an update subscription when switching the mode.

/cc @doodlewind

doodlewind (approved):
Let's land this, thanks for your hard work!

packages/blocks/src/page-block/edgeless/mode-controllers/default-mode.ts
```diff
@@ -180,7 +179,7 @@ export class DefaultModeController extends MouseModeController<DefaultMouseMode>
   }
 
   private _tryDeleteEmptyBlocks() {
-    const emptyBlocks = this._blocks.filter(b => isEmpty(b));
+    const emptyBlocks = this._blocks.filter(b => !b.children.length);
```
alt1o, line 182: This broken test `add empty text`.  Should fix the skip test `add empty text`.

PR #1716, opened 2023-03-15
feat: add subPageIds in pageMeta & add related events

QiShaoXuan:


packages/store/src/workspace/workspace.ts
```diff
@@ -22,7 +22,11 @@ export interface PageMeta {
   title: string;
   createDate: number;
 
-  [key: string]: string | number | boolean;
+  parentId?: string;
```
doodlewind, line 25: I think this won't go into our CRDT data model

PR #1719, opened 2023-03-15
feat: update `setTagSchema` support update tag

zqran:


packages/store/src/workspace/page.ts
```diff
@@ -236,8 +236,10 @@ export class Page extends Space<PageData> {
     return (this.tagSchema.get(id) ?? null) as TagSchema | null;
   }
 
-  setTagSchema(schema: Omit<TagSchema, 'id'>): string {
-    const id = this._idGenerator();
+  setTagSchema(
```
himself65, line 239: if so, you should add a function called `updateTagSchema`

Himself65 (changes requested):

PR #1727, opened 2023-03-16
fix: zoom in/out shortcut key on mac should be metakey

### LINKED ISSUE #1652: The Zoom in/out shortcut key for Edgeless on Mac should be ⌘ + scroll wheel, not Ctrl + scroll wheel
However, on Windows, it is still Ctrl + scroll wheel.

hanchayi:
Close #1652


doodlewind (approved):
Good catch, thanks!

packages/blocks/src/__internal__/utils/index.ts
```diff
@@ -14,5 +14,6 @@ export * from './lit.js';
 export * from './query.js';
 export * from './rect.js';
 export * from './selection.js';
+export * from './shortcut.js';
```
doodlewind, line 17: According to CI log, this is better not be exposed at top level since it breaks SSR by accessing `navigator`.

doodlewind:
Merged, thanks for your contribution!

PR #1729, opened 2023-03-16
refactor(virgo): extract events into event service

Saul-Mirone:
Extract events in virgo into event service.


Flrande:
Do you  run `pnpm test:virgo` ?

Saul-Mirone:
> Do you run `pnpm test:virgo` ?

Yeah, all case passed.

PR #1735, opened 2023-03-16
fix: update column name

zqran:
update column name and support undo


packages/blocks/src/database-block/components/edit-column-popup.ts
```diff
@@ -35,9 +35,7 @@ export class EditColumnPopup extends LitElement {
               ...this.targetTagSchema,
               name: newName,
             });
-            this.targetModel.page.updateBlock(this.targetModel, {
-              name: newName,
-            });
+            this.targetModel.propsUpdated.emit();
```
zqran: Doing so will make undo not work.

packages/blocks/src/database-block/components/edit-column-popup.ts
```diff
@@ -35,7 +35,9 @@ export class EditColumnPopup extends LitElement {
               ...this.targetTagSchema,
               name: newName,
             });
-            this.targetModel.propsUpdated.emit();
+            this.targetModel.page.updateBlock(this.targetModel, {
+              name: newName,
+            });
```
himself65: This is the wrong implementation that bypasses the real bug.

You update the target model so that every time you redo/undo it, it will call propsUpdate in the event handler.

But actually, the real bug is we didn't listen `tagSchema` update in the handler.

Please see my patch

Himself65 (approved):

packages/blocks/src/database-block/components/edit-column-popup.ts
himself65: see update

Issue #1704, opened 2023-03-14
Support drag handle in whiteboard


doodlewind:
* Support drag handle in edgeless mode inside one frame.
* Support drag handle in edgeless mode across multiple frames (or dragging outside of the original frame).

doodlewind:
Closed by https://github.com/toeverything/blocksuite/pull/1736

PR #1736, opened 2023-03-16
feat(drag-handle): support in edgeless mode

fundon:
Related to: #1704

* [x] Support drag handle in edgeless mode inside one frame.
* [x] Support drag handle in edgeless mode across multiple frames (or dragging outside of the original frame).
* [x] Press `Esc` will cancel the operation.
* [x] tests

https://user-images.githubusercontent.com/27926/226121102-f48a4a98-f4c4-4239-97ba-c764743255cd.mov

### DIFF
--- a/packages/store/src/workspace/page.ts
+++ b/packages/store/src/workspace/page.ts
@@ -513,6 +513,45 @@ export class Page extends Space<FlatBlockMap> {
   }
 
   @debug('CRUD')
+  moveBlocksToFrame(blocks: BaseBlockModel[], frameId: string) {
+    if (this.readonly) {
+      console.error('cannot modify data in readonly mode');
+      return;
+    }
+
+    const firstBlock = blocks[0];
+    const currentParentModel = this.getParent(firstBlock);
+
+    if (currentParentModel === null) {
+      throw new Error('cannot find parent model');
+    }
+
+    // the blocks must have the same parent (siblings)
+    if (blocks.some(block => this.getParent(block) !== currentParentModel)) {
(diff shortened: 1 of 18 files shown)


fundon:
@doodlewind In which mode is `drag-handle` enabled?

doodlewind:
> @doodlewind In which mode is `drag-handle` enabled?

I think it should be the default edgeless mouse mode.

/cc @alt1o

packages/store/src/workspace/page.ts
```diff
@@ -507,6 +507,45 @@ export class Page extends Space<FlatBlockMap> {
     nextParentModel.propsUpdated.emit();
   }
 
+  @debug('CRUD')
+  moveBlocksToParent(blocks: BaseBlockModel[], parentId: string) {
```
doodlewind, line 511: Can we reuse the existing `moveBlocks` method?

fundon: https://github.com/toeverything/blocksuite/blob/75173db962085c38d7e676498dc755b96cfbd3ea/packages/store/src/workspace/page.ts#L483-L486

`targetModel` needs to get its parent model first.

But in this case, we already know the parent model (frameId);

doodlewind (approved):

packages/store/src/workspace/page.ts
doodlewind: I think `moveBlocks` should be refactored based on your usage

fundon: I try to refactor it. :)

PR #1752, opened 2023-03-17
fix(playground): clear selection on clicking undo/redo buttons

PerfectPan:
fix the following problem of the playground debug menu

https://user-images.githubusercontent.com/24316656/225934558-a5714fd3-64fe-4f70-9533-66f6bb707438.mp4

The behavior do not consist with the handler of undo/redo key
https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/page-block/utils/bind-hotkey.ts#L69
```ts
hotkey.addListener(HOTKEYS.UNDO, e => {
    if (page.canUndo) clearSelection(page);
    page.undo();
});

hotkey.addListener(HOTKEYS.REDO, e => {
    if (page.canRedo) clearSelection(page);
    page.redo();
});
```

### DIFF
--- a/packages/blocks/src/__internal__/utils/selection.ts
+++ b/packages/blocks/src/__internal__/utils/selection.ts
@@ -242,6 +242,16 @@ export function resetNativeSelection(range: Range | null) {
   range && selection.addRange(range);
 }
 
+export function clearSelection(page: Page) {
+  if (!page.root) return;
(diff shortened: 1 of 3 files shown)


doodlewind (approved):
Good catch! While just saying that in the future , it may be best to record selection state in history stack, so that we can restore it together with the block state (see https://docs.yjs.dev/api/undo-manager#example-add-additional-information-to-the-stackitems)

PR #1748, opened 2023-03-17
refactor(editor): move content parser out of editor lifecycle

doodlewind:
Fixes #1742 since we can `new ContentParser(page)` when we have a new page, no need to wait for the editor instance

/cc @Himself65 @QiShaoXuan

PR #1753, opened 2023-03-17
fix(playground): import preset after block updated

donaldxdonald:
image


doodlewind:
Thanks for keeping an eye on our state and your contribution! However this should be a regression fixed by #1748, so closing this in favor of it.

PR #1757, opened 2023-03-18
feat(database, selection): improve DnD for database block

fundon:
* [x] drag and drop a block into empty database
* [x] improve DragIndicator position
* [x] improve DragHandle position
* [x] BlockHub 

https://user-images.githubusercontent.com/27926/226604094-2ce7d9f6-eebb-45be-b0b7-d49d59211604.mov

### DIFF
--- a/packages/blocks/src/components/block-hub.ts
+++ b/packages/blocks/src/components/block-hub.ts
@@ -5,8 +5,8 @@ import type {
 } from '@blocksuite/blocks/std';
 import {
   getClosestBlockElementByPoint,
+  getDropZoneByBlockAndPoint,
   getModelByBlockElement,
-  getRectByBlockElement,
   NonShadowLitElement,
   Point,
 } from '@blocksuite/blocks/std';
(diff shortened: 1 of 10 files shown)


fundon:
> One thing I want to confirm is whether the following two visual effects need to be consistent?

If the database is empty, the width of drag handle line is equals database block' width, otherwise it equals to the specific block' width.

fundon:
@zqran 


https://user-images.githubusercontent.com/27926/229326372-21b601e1-3546-4428-a2e4-0fb76ed76dda.mov

packages/blocks/src/components/block-hub.ts
```diff
@@ -727,16 +726,40 @@ export class BlockHub extends NonShadowLitElement {
...
+            ?.getBoundingClientRect();
+          assertExists(headerBounds);
+          targetRect = new DOMRect(
+            headerBounds.left,
+            headerBounds.bottom + 1,
+            targetRect.width,
+            1
+          );
+        }
+      }
```
lawvs, line 751: I suggest adding some comments about drag details here for better understanding

lawvs (changes requested):

zqran:
> @zqran Throws an exception when all rows are moved out of the database

Fixed.

doodlewind (approved):

PR #1759, opened 2023-03-18
feat(database): generate a default structure when creating a database

zqran:


packages/blocks/src/database-block/utils/database-init.ts
```diff
@@ -0,0 +1,43 @@
+import {
+  assertExists,
+  type BaseBlockModel,
+  type Page,
+} from '@blocksuite/store';
+
+export function initDatabaseBlock(
```
doodlewind, line 7: This looks better to be defined as a method in BlockService

zqran: ok,  i will change it.

doodlewind (approved):

Issue #1773, opened 2023-03-19
Override placeholder in database block


Himself65:
I think this should be hidden

![image](https://user-images.githubusercontent.com/14026360/226199690-631aab5e-da55-43be-9cdd-95de39f7529e.png)

zqran:
Can we add an property(e.g. `showTooltip: boolean`) to `ParagraphBlock` to control whether to display tooltip?

doodlewind:
> Can we add an property(e.g. `showTooltip: boolean`) to `ParagraphBlock` to control whether to display tooltip?

Block components only receive `model` and `host` properties, the `host` can be used for similar context-like purpose. If you want to override placeholder in database container, I think on solution is to create a new `host` in database block, override its flavours to `affine:database`, and in this case its children paragraphs can adapt to it based on this host `flavour` property.

PR #1785, opened 2023-03-20
feat(database): hide placeholder of paragraph in database

zqran:
close: #1773

### DIFF
--- a/packages/blocks/src/__internal__/service/components.ts
+++ b/packages/blocks/src/__internal__/service/components.ts
@@ -23,7 +23,21 @@ export function BlockElement(
   edgeless = false
 ) {
   switch (model.flavour) {
-    case 'affine:paragraph':
+    case 'affine:paragraph': {
+      const rowHost = { ...host };
+      const parentModel = model.page.getParent(model);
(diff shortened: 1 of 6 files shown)


packages/blocks/src/__internal__/service/components.ts
```diff
@@ -23,7 +23,21 @@ export function BlockElement(
   edgeless = false
 ) {
   switch (model.flavour) {
-    case 'affine:paragraph':
+    case 'affine:paragraph': {
+      const rowHost = { ...host };
+      const parentModel = model.page.getParent(model);
+      if (parentModel?.flavour === 'affine:database') {
+        rowHost.flavour = 'affine:database';
```
doodlewind, line 30: You should pass in a different host from database block

doodlewind: Guess we can pass a new host here

https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/database-block/database-block.ts#L150

zqran: oh, more clearly here.

doodlewind (approved):

PR #1787, opened 2023-03-20
refactor(store): remove redundant `moveBlocksToParent`

doodlewind:


packages/store/src/workspace/page.ts
```diff
@@ -469,86 +469,49 @@ export class Page extends Space<FlatBlockMap> {
   @debug('CRUD')
   moveBlocks(
     blocks: BaseBlockModel[],
-    targetModel: BaseBlockModel,
-    top = true
```
doodlewind, line 473: The `top` here is confusing. If it's true, the blocks will actually be inserted before the target position.

PR #1790, opened 2023-03-20
fix(drag-handle): height of drag handle line when zooming in/out in edgeless mode

fundon:
### Before
Screenshot 2023-03-20 at 5 50 28 PM

### After
Screenshot 2023-03-20 at 5 52 31 PM


doodlewind (approved):
Good catch!

PR #1795, opened 2023-03-20
feat(database): insert database by slash-menu

zqran:


Himself65 (changes requested):
this should be disabled by `flag.enable_database` since it's still unstable, and we don't want to make it available for the client to create data now in case of stale data in the future

lawvs (approved):

Himself65 (approved):

PR #1804, opened 2023-03-21
fix: do not allow text in button to be selected

fundon:
Screenshot 2023-03-21 at 11 08 23 PM

Click on `zoom out`


doodlewind (approved):

doodlewind:
Seem that we have one test case regression (the `basic link` case). @fundon can you reproduce it locally?

fundon:
Okay, I'll take a look at it and fix it.

PR #1807, opened 2023-03-22
feat(database): database support search

zqran:


Himself65 (changes requested):
focus will lost when you press delete


https://user-images.githubusercontent.com/14026360/227041246-853eccb4-ff7d-4ec4-bec1-54185c0de184.mp4

packages/blocks/src/database-block/database-block.ts
```diff
@@ -276,18 +360,154 @@ export class DatabaseBlockComponent
...
+  }
+
+  private _onSearch = (event: InputEvent) => {
+    const inputValue = (event.target as HTMLInputElement).value.trim();
+    this._searchState = SearchState.Searching;
+    if (inputValue === '') {
+      this._searchState = SearchState.SearchInput;
+    }
+
+    const rowKeyValues = this._getRowKeyValues();
```
himself65: `_getRowKeyValues` is a heavy compted function, consider it to be a memo?

zqran: Yes, I have encountered a problem before, after adding a new row, I can't get the content of the new row header column(first paragraph column).

doodlewind (approved):

PR #1813, opened 2023-03-22
fix(edgeless): the edgeless toolbar should not be able to display two submenus

### LINKED ISSUE #1808: The Edgeless Toolbar should not be able to display two submenus.
CleanShot 2023-03-22 at 5 23 15@2x

alt1o:
fix #1808 
When clicking the shape button, we should set the default shape mode. Using click away to control the shape panel's visibility.

https://user-images.githubusercontent.com/21084335/226867497-be7cda4e-0d33-434c-b9b5-5608eaabbc8d.mov

### DIFF
--- a/packages/blocks/src/page-block/edgeless/toolbar/shape-tool/shape-tool-button.ts
+++ b/packages/blocks/src/page-block/edgeless/toolbar/shape-tool/shape-tool-button.ts
@@ -69,13 +69,6 @@ export class EdgelessShapeToolButton extends LitElement {
     super.firstUpdated(changedProperties);
   }
 
-  updated(changedProperties: Map<string, unknown>) {
-    if (changedProperties.has('mouseMode')) {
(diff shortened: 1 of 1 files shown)


doodlewind (approved):

packages/blocks/src/page-block/edgeless/toolbar/shape-tool/shape-tool-button.ts
```diff
@@ -69,13 +69,6 @@ export class EdgelessShapeToolButton extends LitElement {
     super.firstUpdated(changedProperties);
   }
 
-  updated(changedProperties: Map<string, unknown>) {
```
doodlewind, line 72: I like using less `updated` callback, the `changedProperties.has('prop')` pattern is really bad for refactoring.

doodlewind (approved):

PR #1820, opened 2023-03-22
fix(edgeless): drag handle should only appear in actived frame block

### LINKED ISSUE #1812: When a Text component is not in edit state, the drag handle should not appear when hovering over it with the mouse
CleanShot 2023-03-22 at 5 40 39@2x

alt1o:
fix #1812 

https://user-images.githubusercontent.com/21084335/226877115-0a9dd59e-5f20-4b4f-9aaf-1030f155b5d8.mov

### DIFF
--- a/packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
+++ b/packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
@@ -19,6 +19,7 @@ import {
   SurfaceManager,
 } from '@blocksuite/phasor';
 import {
+  assertExists,
   type BaseBlockModel,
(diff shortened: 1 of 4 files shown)


doodlewind (approved):

alt1o:
> When a block is dragged and dropped into another frame, the selection should be updated because the edit state has been switched to the new frame. Screenshot 2023-03-22 at 7 14 54 PM

Fixed in https://github.com/toeverything/blocksuite/pull/1820/commits/dec32cf33259d61c24ca81edfbeb52e7562fbb19

fundon (approved):

PR #1822, opened 2023-03-22
fix(blocks): inherit behaviour about inline format

### LINKED ISSUE #1818: After using keyboard shortcuts or slash commands to modify the style of a block or text, editing the content after formatted text should inherit the previous formatting
Recorded video is attached

https://user-images.githubusercontent.com/22443345/226873872-8bc9bc26-15ec-4b36-9c1d-37010df61781.mp4

Note:
If I using Markdown syntax such as **xxx**, editing the content after formatted text should not inherit the previous formatting

Flrande:
In addition to link and inline code, for text that has been added with an inline style, the input immediately following it inherits its inline style.

Close #1818


lawvs (changes requested):
Can not work with CJK IME

https://user-images.githubusercontent.com/18554747/226996072-788f204f-1483-4f90-863a-1a7a07d4f226.mov

lawvs (approved):

PR #1825, opened 2023-03-22
fix: lang-button without line break

### LINKED ISSUE #1824: Longer "lang-button" should not have a line break
bug

lzlme:
fix #1824


doodlewind (approved):
Good catch, thanks!

PR #1826, opened 2023-03-22
fix: normal format should be inherited

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/1818


lawvs:
Could you check this implementation? @Flrande

Flrande (approved):

Flrande (changes requested):
We should not inherited code and link style.

Flrande (approved):

Flrande:
I will continue working on this branch.

Flrande:
20230323103813_rec_

I think we should not inherited style here.

PR #1828, opened 2023-03-23
feat: add `shiftPage` api in workspace

QiShaoXuan:


packages/store/src/workspace/workspace.ts
```diff
@@ -99,6 +96,14 @@ class WorkspaceMeta extends Space<WorkspaceMetaState> {
     });
   }
 
+  transformObjectToYMap(obj: Record<string, unknown>) {
```
doodlewind, line 99: Should be private and we may have already got similar utils.

doodlewind (approved):

PR #1829, opened 2023-03-23
fix(edgeless): when focused, Virgo will attempt to automatically scroll into view.

alt1o:
https://user-images.githubusercontent.com/21084335/227096059-69c0b6dd-483d-4f35-a36f-062a7aedb6bc.mov


doodlewind (changes requested):
I think it may be better try disabling the scroll into view behavior in edgeless mode.

/cc @Flrande

Flrande:
> I think it may be better try disabling the scroll into view behavior in edgeless mode.

I will do it later.

PR #1838, opened 2023-03-23
fix(database): undo should work in database title and rich-text

zqran:


packages/blocks/src/components/remote-selection/remote-selection.ts
```diff
@@ -159,6 +163,31 @@ export class RemoteSelection extends LitElement {
...
+          if (
+            models.length === 2 &&
+            matchFlavours(models[1], ['affine:database'] as const)
+          ) {
+            restoreSelection({
+              type: 'DatabaseRichText',
+              startOffset: userRange.startOffset,
+              endOffset: userRange.endOffset,
+              models: [models[0], models[1]],
+              databaseCellPos: userRange.blockPos,
```
doodlewind, line 187: I don't think it's good if adding new block type requires modifying the cursor selection module. @Flrande do we have some solution to avoid this kind of code smell?

In the worst case, I think ignoring cell cursor in database block in the first version can still be acceptable, which is still better than hard-coded here. For the non-title columns in database block, they don't have to support cursor.

Flrande: Considering that there will be a lot of non-text input scenarios (such as dates, statuses, progress bars, etc.) in the database, I think we can ignore the internal selection tracking.

packages/blocks/src/__internal__/utils/block-range.ts
```diff
@@ -41,13 +50,19 @@ export type BlockRange = {
...
+   */
+  databaseCellPos?: DatabaseCellPos;
   /**
    * Only one model, the page model
+   *
+   * Two models for database rich-text
    */
-  models: [BaseBlockModel];
+  models: [BaseBlockModel] | [BaseBlockModel, BaseBlockModel];
 };
```
lawvs, line 66: You should use the unions type to ensure type-safe, for example

```ts
type ExtendBlockRange = {
  type: 'Title';
  startOffset: number;
  endOffset: number;
  models: [BaseBlockModel];
} | 
{
  type: 'Database';
  cellPos: number;
  startOffset: number;
  endOffset: number;
  models: [BaseBlockModel, BaseBlockModel];
}

```

And I think it can merge `DatabaseTitle` and `DatabaseRichText` into a `Databse` type.

lawvs (changes requested):

packages/blocks/src/__internal__/utils/block-range.ts
```diff
@@ -41,13 +50,19 @@ export type BlockRange = {
...
 
 type ExtendBlockRange = {
-  type: 'Title';
+  type: 'Title' | 'DatabaseTitle' | 'DatabaseRichText';
```
doodlewind, line 53: @zqran please revert these changes for `ExtendBlockRange` since its current design is not intended for this usage at the moment. Special treatment for database title and database rich text would make the maintenance of selection harder. We can postpone related cursor requirements in first version of database block.

/cc @lawvs any idea for a more generic range structure design?

zqran: >  Special treatment for database title and database rich text would make the maintenance of selection harder.

Yes, I found it when I implemented it.
It should be supported to achieve undo/redo through shortcuts in the database, right?

lawvs: For the native range, we can include the `virgo` instance within the block range to handle multiple rich-text blocks.

For the block range, we need to wait for the product design.

Flrande: In Virgo, you can convert between `VRange` and browser-native range using `toDomRange` and `toVRange`.

Flrande:
https://github.com/toeverything/blocksuite/blob/d3efce057ecb104b2ba0a2f1b7e2c6b60494e0cb/packages/store/src/awareness.ts#L9-L13

The selection interface for collaborative cursors currently does not support database-type selections. Should we ignore synchronization of collaborative cursors when they are in a database area? @doodlewind

doodlewind:
> Should we ignore synchronization of collaborative cursors when they are in a database area?

LGTM at our next release. Afterwards we will support commenting and annotation on page, which could be the better timing for a more comprehensive design.

packages/blocks/src/__internal__/utils/block-range.ts
lawvs: > In Virgo, you can convert between `VRange` and browser-native range using `toDomRange` and `toVRange`.

In addition, the `block range` is designed to make consistency between the browser-native range and the block range.

It's different from VRange since the user can select multiple blocks at a time, and VRange can't handle range cross blocks.

zqran:
> @zqran please revert these changes for ExtendBlockRange since its current design is not intended for this usage at the moment. 

@doodlewind 
1. removed support for database selection
2. only undo/redo support remains

Flrande (approved):

doodlewind (approved):
Thanks for the precise refactoring!

PR #1861, opened 2023-03-25
feat(database): update cell style

zqran:


packages/global/src/config/icons.ts
doodlewind, line 1: I think we can split this module into multiple files.

zqran: Like this?

```text
/icons
  index.ts
  database.ts
```

Himself65 (commented):
image

the text shouldn't wrap and overflow when larger than its box

Himself65 (approved):
LGTM and I can merge this for now. And should not the PR diff be too large to hard to review

Issue #1850, opened 2023-03-24
Missing drag handle on embedded page


doodlewind:
To reproduce, see https://blocksuite-docs.vercel.app/getting-started.html

<img width="600" alt="image" src="https://user-images.githubusercontent.com/7312949/227490342-f6075b4e-6f42-41e6-9b7d-7932c1d9ada9.png">

Drag handle is missing but slash menu works.

lzlme:
It seems that the &lt;simple-affine-editor&gt; is embedded in the page instead of the full-featured editor (for example, there is also no &lt;debug-menu&gt;), so is it just adding the "drag handle" or everything else?

lzlme:
> @lzlme I think we can reuse the `createDragHandle` util.

Sorry I mistook `drag-handle` for `block-hub`.

I'm trying to fix this, and I found that there may not be enough space for the drag-handle, because even when opening a non-embedded page, there is still no space for the drag-handle as long as the page width is less than 720px.
Although `docs/getting-started.md` has added `padding: 50px` to the container, the `onContainerMouseMove` in the `drag-handle.ts` does not trigger in this padding area.

I've tried adding padding directly to `.affine-default-viewport` or `.affine-default-page-block-container`, which looks fine, but causes many test cases to fail.

Do you have a good solution for this ?

PR #1867, opened 2023-03-27
fix: optimize page padding for drag-handle and mobile

lzlme:
fix: #1850


doodlewind (approved):
Good catch! This also optimizes the layout on mobile.

PR #1879, opened 2023-03-28
feat(database): insert column

zqran:
1. Temporarily skip tests that depend on non `multi-select` columns
2. I will refactor these tests after implementing update column type

### DIFF
--- a/packages/blocks/src/database-block/components/edit-column-popup.ts
+++ b/packages/blocks/src/database-block/components/edit-column-popup.ts
@@ -153,6 +153,11 @@ export const isDivider = (action: ColumnAction): action is Divider => {
   return action.type === 'divider';
 };
 
+export const enum ColumnInsertPosition {
+  Left = 'left',
+  Right = 'right',
+}
+
 @customElement('affine-database-column-type-popup')
 class ColumnTypePopup extends LitElement {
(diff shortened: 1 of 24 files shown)


packages/blocks/src/database-block/components/edit-column-popup.ts
```diff
@@ -251,6 +251,9 @@ export class EditColumnPopup extends LitElement {
   @property()
   setTitleColumnEditId!: (titleId: string) => void;
 
+  @property()
+  insertColumn!: (position: 'right' | 'left') => void;
```
himself65: Use `const enum`

zqran: done

packages/blocks/src/database-block/components/edit-column-popup.ts
```diff
@@ -153,6 +153,11 @@ export const isDivider = (action: ColumnAction): action is Divider => {
   return action.type === 'divider';
 };
 
+export const enum ColumnInsertPosition {
```
himself65: this could be moved to `global/constant`

Himself65 (approved):

doodlewind (approved):

PR #1880, opened 2023-03-28
feat(playground): add new tab menu

lawvs:
- keyed page block in editor
- add tab menu in playground


himself65:
```


../../../blocks/src/__internal__/utils/virgo.ts:4:10
--
10:52:07.201 | Type error: 'getEditorContainer' is declared but its value is never read.


```

doodlewind (approved):

PR #1887, opened 2023-03-29
feat(database): support switching column type

zqran:
https://user-images.githubusercontent.com/15389209/228466586-b7de1511-acd8-42c1-991c-aabe931cd600.mov


Flrande:
When switching column types, the table header does not update correspondingly. Is this part still under development?

zqran:
> When switching column types, the table header does not update correspondingly. Is this part still under development?

Do you mean **column title**? The **column title** should't update when switch column type.

Flrande (approved):

packages/blocks/src/database-block/components/cell-container.ts
```diff
@@ -30,9 +30,14 @@ export class DatabaseCellContainer
...
 
+  private _timerId: number | undefined = undefined;
+
   setValue(value: unknown) {
     if (value !== undefined) {
-      setTimeout(() => {
+      if (this._timerId !== undefined) {
+        window.clearTimeout(this._timerId);
+      }
+      this._timerId = window.setTimeout(() => {
```
doodlewind, line 40: Why is this timeout here? Is this avoiding some anti-pattern?

doodlewind (changes requested):
Fast progress indeed! But I still want to figure out the best practice on a few issues before landing this. I'll see what I can do.

zqran:
> Fast progress indeed! But I still want to figure out the best practice on a few issues before landing this. I'll see what I can do.

Full support, I will refactor these after discussing and clarifying best practices.

doodlewind (approved):

packages/store/src/workspace/page.ts
```diff
@@ -270,6 +279,42 @@ export class Page extends Space<FlatBlockMap> {
     });
   }
 
+  // TODO: remove hard coded types
+  updateColumnValue(columnId: string, newType: 'select' | 'rich-text') {
```
doodlewind, line 283: @zqran this is a generic workaround, we will further improve this

zqran: 👌

PR #1891, opened 2023-03-29
feat(database): support renaming and deletion of `select` tag

zqran:
https://user-images.githubusercontent.com/15389209/228783801-dedc617a-9b38-4d25-9d3a-bb3c420bc51d.mov


packages/blocks/src/database-block/components/column-type/select.ts
```diff
@@ -137,7 +228,7 @@ class SelectAction extends LitElement {
           return html`
             <div
               class="action ${action.type}"
-              @click=${() => this.onAction(action.type, this.index)}
+              @mousedown=${() => this.onAction(action.type, this.index)}
```
doodlewind, line 231: `action.type` should be strongly typed

packages/store/src/workspace/page.ts
```diff
@@ -279,6 +279,36 @@ export class Page extends Space<FlatBlockMap> {
     });
   }
 
+  updateSelectedColumnValue(
```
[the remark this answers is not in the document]
doodlewind: `updateColumnValue` is renamed to `convertColumn` in #1913

packages/store/src/workspace/page.ts
```diff
@@ -279,6 +279,36 @@ export class Page extends Space<FlatBlockMap> {
...
+    this.transact(() => {
+      const yColumns = this.yColumns.get(rowId);
+      assertExists(yColumns);
+      const cell = yColumns.get(columnId) as Y.Map<string[]> | undefined;
+      if (!cell) return;
+
+      const selected = cell.get('value') as string[];
+      let newSelected = [...selected];
+      if (value !== undefined) {
+        // rename tag
```
doodlewind, line 297: I believe mixing rename and deleting actions based on `value` param is an anti-pattern

doodlewind (changes requested):
I feel current column updating API design still worth reconsidering

doodlewind (approved):

PR #1904, opened 2023-03-30
feat: support using component based on `WithDisposable`

lawvs:
usage

```ts
class SomeElement extends WithDisposable(ShadowlessElement) {
    super.connectedCallback();
    this._disposables.add(...);
  }
}
```
see https://lit.dev/docs/composition/mixins/


doodlewind (commented):
I would choose not to introduce the concept of mixin, I'll try to handle this.

Issue #1881, opened 2023-03-28
Incorrect indentation when deleting list item


lzlme:
https://user-images.githubusercontent.com/117659326/228162013-c7b77b6a-9bfe-45a0-a71b-7100a97b3576.mov

https://user-images.githubusercontent.com/117659326/228162061-0e1d5167-5bd9-45ab-9947-028826465b8c.mov

zuoxiaodong0815:
@doodlewind I think the behaviour is normal, so I'm going to roll back the relevant code

PR #1907, opened 2023-03-31
fix: indent correctly when deleting list item

lzlme:
fix: #1881


doodlewind (commented):
The `handleLineStartBackspace` function is too verbose, I'll try refactoring it.

doodlewind (approved):

PR #1914, opened 2023-03-31
refactor: remove `y-indexeddb`

himself65:


himself65:
See https://github.com/toeverything/blocksuite/pull/2207

Issue #1845, opened 2023-03-24
Hovertips should disappear once the submenu list is opened


joebeijing:
https://user-images.githubusercontent.com/22443345/227420288-6658843e-5937-4ab2-b631-b340cdedfe89.mp4

PR #1924, opened 2023-04-02
fix: tooltip should be hidden when the submenu list is opened

PerfectPan:
Close https://github.com/toeverything/blocksuite/issues/1845

I had two considerations in fixing this issue:
1. It seems that I can either create a new prop to control whether a tooltip should be displayed or just clear tooltip. I choose the latter because I think the value of tooltip can potentially represent whether it should be displayed. Just like the code written in file:
```ts
${tooltip
  ? html`<tool-tip inert role="tooltip" tip-position="top" arrow
      >${tooltip}</tool-tip
    >`
  : nothing}
```
2. I create a state to track the state of the popper so that I can request a render when the state changes reactively.

### DIFF
--- a/packages/blocks/src/page-block/edgeless/toolbar/brush-tool/brush-tool-button.ts
+++ b/packages/blocks/src/page-block/edgeless/toolbar/brush-tool/brush-tool-button.ts
@@ -4,7 +4,7 @@ import './brush-menu.js';
 import { PenIcon } from '@blocksuite/global/config';
 import { createPopper } from '@popperjs/core';
 import { css, html, LitElement } from 'lit';
-import { customElement, property } from 'lit/decorators.js';
+import { customElement, property, state } from 'lit/decorators.js';
 
 import type { MouseMode } from '../../../../__internal__/index.js';
 import { getTooltipWithShortcut } from '../../components/utils.js';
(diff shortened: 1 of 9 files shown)


packages/blocks/src/page-block/edgeless/toolbar/brush-tool/brush-tool-button.ts
```diff
@@ -101,8 +106,11 @@ export class EdgelessBrushToolButton extends LitElement {
 
     return html`
       <edgeless-tool-icon-button
-        .tooltip=${getTooltipWithShortcut('Pen', 'P')}
+        .tooltip=${this._popperShow
+          ? undefined
+          : getTooltipWithShortcut('Pen', 'P')}
```
alt1o, line 111: Prefer to use ```null``` or ```empty string```

packages/blocks/src/page-block/edgeless/toolbar/brush-tool/brush-tool-button.ts
```diff
@@ -101,8 +106,11 @@ export class EdgelessBrushToolButton extends LitElement {
...
         .active=${type === 'brush'}
+        .testId=${'brush'}
```
alt1o, line 113: It is recommended to use locatorEdgelessToolButton to locate the tool button, without the need to add a new testId.

alt1o (changes requested):

doodlewind (approved):

doodlewind:
Landed, thanks!

PR #1925, opened 2023-04-02
refactor(database): `columns` -> `cells`

doodlewind:
Lacking the concept of `cell` leads to serious ambiguity and incorrect variable naming in current database implementation. So the decision is to rename the `ext:columns` nested map to `ext:cells`, which is indeed a two-dimensional map (`rowId` + `columnId`) that contains database cells.

Before (nuanced "column" concept):

``` ts
  getColumn(model: BaseBlockModel, schema: ColumnSchema): BlockColumn | null {
    const yColumns = this.yColumns.get(model.id);
    const yColumnMap = (yColumns?.get(schema.id) as Y.Map<unknown>) ?? null;
    if (!yColumnMap) return null;

    return {
      columnId: yColumnMap.get('columnId') as string,
      value: yColumnMap.get('value') as unknown,
    };
  }
```

After (`yCells` -> `yRow` -> `yCell`):

``` ts
  getCell(model: BaseBlockModel, schema: ColumnSchema): Cell | null {
    const yRow = this.yCells.get(model.id);
    const yCell = (yRow?.get(schema.id) as Y.Map<unknown>) ?? null;
    if (!yCell) return null;

    return {
      columnId: yCell.get('columnId') as string,
      value: yCell.get('value') as unknown,
    };
  }
```

For future kanban view, we can still group them based on different values on the same `column` (columnSchema).

After this is landed, I would continue renaming `columnSchema` to `column`, which reflects the table structure better.

### DIFF
--- a/packages/store/src/workspace/page.ts
+++ b/packages/store/src/workspace/page.ts
@@ -721,7 +721,7 @@ export class Page extends Space<FlatBlockMap> {
     });
 
     if (matchFlavours(model, ['affine:page'] as const)) {
-      model.columns = yBlock.get('ext:columns') as Y.Map<Y.Map<unknown>>;
+      model.cells = yBlock.get('ext:cells') as Y.Map<Y.Map<unknown>>;
       model.columnSchema = yBlock.get('ext:columnSchema') as Y.Map<unknown>;
 
       const titleText = yBlock.get('prop:title') as Y.Text;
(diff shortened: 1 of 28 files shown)


packages/store/src/workspace/page.ts
```diff
@@ -721,7 +721,7 @@ export class Page extends Space<FlatBlockMap> {
     });
 
     if (matchFlavours(model, ['affine:page'] as const)) {
-      model.columns = yBlock.get('ext:columns') as Y.Map<Y.Map<unknown>>;
+      model.cells = yBlock.get('ext:cells') as Y.Map<Y.Map<unknown>>;
```
doodlewind, line 724: The semantics here is wrong, will be fixed

Issue #1926, opened 2023-04-02
Cannot enter content below the divider


lzlme:
If there is no content below the divider when it is inserted, no new content can be entered.

https://user-images.githubusercontent.com/117659326/229369727-80cbe9a8-5b8e-4f9c-b688-ce379541f6cc.mov

PR #1927, opened 2023-04-02
fix: add paragraph when insert divider

lzlme:
fix: #1926

### DIFF
--- a/packages/blocks/src/divider-block/divider-block.ts
+++ b/packages/blocks/src/divider-block/divider-block.ts
@@ -2,7 +2,11 @@
 import { css, html } from 'lit';
 import { customElement, property } from 'lit/decorators.js';
 
-import { type BlockHost, NonShadowLitElement } from '../__internal__/index.js';
+import {
+  asyncFocusRichText,
+  type BlockHost,
+  NonShadowLitElement,
+} from '../__internal__/index.js';
 import { BlockChildrenContainer } from '../__internal__/service/components.js';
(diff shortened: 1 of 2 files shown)


packages/blocks/src/divider-block/divider-block.ts
```diff
@@ -32,6 +36,13 @@ export class DividerBlockComponent extends NonShadowLitElement {
   firstUpdated() {
     this.model.propsUpdated.on(() => this.requestUpdate());
     this.model.childrenUpdated.on(() => this.requestUpdate());
+    const { page } = this.model;
+    const nextSibling = page.getNextSibling(this.model);
+    const parent = page.getParent(this.model);
+    if (!nextSibling) {
+      const id = page.addBlock('affine:paragraph', {}, parent);
+      asyncFocusRichText(page, id);
+    }
```
lawvs, line 45: Consider moving the side effects to the slash menu config instead of performing them when mounting the UI.

https://github.com/toeverything/blocksuite/blob/7c5010cdbc33a315bf434fb601b54d3c56d289f4/packages/blocks/src/components/slash-menu/config.ts#L70-L77

lawvs (changes requested):
Good improvement

lzlme:
> Good improvement!

@lawvs OK, I've done.

doodlewind (approved):

doodlewind:
Thanks for the feedback and the fix!

Issue #1888, opened 2023-03-29
Support dragging blocks under block selection


doodlewind:
Currently we've supported dragging blocks under block selection using drag handle:

https://user-images.githubusercontent.com/7312949/228499162-91fca053-78a0-4923-b19e-e0f5f997dd4b.mov

But for the optimal user experience, it would be best if the cursor could be the grip style when hovering over these selected blocks, and support direct dragging that starts from these selected block content.

@fundon could you evaluate if it's feasible for you to implement? Thanks!

/cc @joebeijing

doodlewind:
Implemented in #1946

PR #1946, opened 2023-04-04
feat(selection): support dragging blocks under block-level selection

fundon:
Related to: #1888

https://user-images.githubusercontent.com/27926/230043752-e9085930-bab7-4e0b-9a46-5c437baadb81.mov


fundon:
One problem, when developing locally, the `affine-page-selected-rects` doesn't render correctly when developed locally in AFFiNE.

doodlewind (approved):

PR #1950, opened 2023-04-04
test(store): add unit tests for `DatabaseManager`

Flrande:


packages/store/src/__tests__/database.unit.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+
+    db.deleteColumn(columnId);
+    expect(db.getColumn(columnId)).toBeNull();
+  });
+
+  test('getCell', () => {
+    const modelId = page.addBlock('affine:paragraph', {
+      text: new page.Text('paragraph'),
+    });
+    const schema: Column = {
```
doodlewind, line 172: Name this as `column` will be better. I plan to remove the concept of `ColumnSchema` and make it `Column`.

PR #1953, opened 2023-04-05
feat(playground): add sentry browser tracing

himself65:


doodlewind:
Why adding sentry in BlockSuite playground? It makes sense for AFFiNE but I don't see the playground as an online service, here is used for editor bug report and local testing. I'm reverting this.

himself65:
> Why adding sentry in BlockSuite playground? It makes sense for AFFiNE but I don't see the playground as an online service, here is used for editor bug report and local testing. I'm reverting this.

It's also could help when user try playground

doodlewind:
We use playground to reproduce issues, to my knowledge remote sentry logs may not helpful if some issue is only recorded by sentry but not reported on GitHub. I'm also not sure if this will bring unnecessary noises during local development.

PR #1959, opened 2023-04-06
refactor(store): use id from `idGenerator` as page id

lawvs:
Use don't need to pass the `id` parameter when creating a new page.

### DIFF
--- a/packages/store/src/workspace/workspace.ts
+++ b/packages/store/src/workspace/workspace.ts
@@ -444,7 +444,8 @@ export class Workspace {
     });
   }
 
-  createPage(pageId: string, parentId?: string) {
+  createPage(parentId?: string) {
+    const pageId = this.idGenerator();
     if (this._hasPage(pageId)) {
       throw new Error('page already exists');
     }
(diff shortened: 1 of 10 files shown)


packages/store/src/workspace/workspace.ts
```diff
@@ -444,7 +444,8 @@ export class Workspace {
     });
   }
 
-  createPage(pageId: string, parentId?: string) {
+  createPage(parentId?: string) {
```
doodlewind, line 447: I think we can support `createPage(id? string)` that makes it possible to override the generated id (so that we can construct the `page0` without breaking ids), and having a new API for setting `parentId`.

Himself65 (approved):
I don't mind this change

packages/store/src/workspace/workspace.ts
lawvs: I plan to modify the API to the following form:

`createPage(options: { id?: sting, parentId?: string } = {})`

doodlewind: LGTM since we have `new Workspace({ id })` API

doodlewind:
Closing in favor of #2272

PR #1993, opened 2023-04-08
feat: add up and down keys to select language in code block

### LINKED ISSUE #1991: Code block language selection should support up and down keys
When choosing a language using code block, it should support using up and down keys to select a different language.

yuusheng:
close #1991

### DIFF
--- a/packages/blocks/src/code-block/components/lang-list.ts
+++ b/packages/blocks/src/code-block/components/lang-list.ts
@@ -98,6 +98,9 @@ export class LangList extends ShadowlessElement {
   @state()
   private _filterText = '';
 
+  @state()
+  private _currentSelectedIndex = -1;
+
   @property()
   selectedLanguage = '';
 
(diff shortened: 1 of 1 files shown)


packages/blocks/src/code-block/components/lang-list.ts
```diff
@@ -98,6 +98,9 @@ export class LangList extends ShadowlessElement {
   @state()
   private _filterText = '';
 
+  @state()
+  private _currentSelectedIndex = -1;
```
doodlewind, line 102: Would be better if this index can start from the hover position!

https://user-images.githubusercontent.com/7312949/230734987-a7558641-7ec1-4a4e-852c-328c9acd3beb.mov

yuusheng: In fact, I don't think we need to do this because hovering and up and down keys are not the same behavior, and combining these two behaviors together would be a little bit strange. But if you want the editor acts like this, I will make some modification.

zqran: Here's an example of a google search, which is a common behavior.

https://user-images.githubusercontent.com/15389209/230771543-bc67d373-2d58-46af-b41b-f55cfb1488e1.mov

doodlewind (approved):
Well I think we can land this and leave the space for future optimizations, thanks @yuusheng !

yuusheng:
> Well I think we can land this and leave the space for future optimizations, thanks @yuusheng !

OK! Have something to deal with. I will complete it tomorrow.

PR #1995, opened 2023-04-08
fix: reverse dark mode icon and add tooltip

lzlme:
The icon display in dark/light mode is reversed, so fix it.
And add a tooltip like other buttons.


doodlewind (approved):
Good catch!

PR #1996, opened 2023-04-08
fix: remove redundant button attribute

lzlme:
&lt;el-button&gt; has no `content` attribute


lzlme:
However, the additional `content` attribute is used in the test case.

Issue #629, opened 2023-01-09
Support database block


Himself65:
```mermaid
---
title: Database
---
classDiagram
    DatabaseBlock <|-- DatabaseHeader
    DatabaseBlock <|-- DatabaseRows
    DatabaseBlock <|-- DatabaseFooter
    class DatabaseBlock {
        +String title
    }
    
    class DatabaseHeader {
        +Column[] columns
    }

    class DatabaseFooter {
        
    }

    DatabaseHeader <|-- DatabaseColumnEditingPopup
    class DatabaseColumnEditingPopup {
        +Column column
        +editTitle()
        +editProperty()
    }

    DatabaseRows <|-- DatabaseCellContainer
    class DatabaseRows {
        +BaseBlock[] children
    }

    DatabaseCellContainer <|-- DatabaseCell
    DatabaseCellContainer <|-- DatabaseCellEditing
    class DatabaseCellContainer {
        +BaseBlock rowBlock
    }

    class DatabaseCell {
        +DatbaseBlock database
        +BaseBlock rowBlock
        +Column column
        +Tag tag
    }

    class DatabaseCellEditing {
        +DatbaseBlock database
        +BaseBlock rowBlock
        +Column column
        +Tag tag
    }
```

```[tasklist]
# Features
- [ ] https://github.com/toeverything/blocksuite/pull/960
- [ ] https://github.com/toeverything/blocksuite/pull/959
- [ ] https://github.com/toeverything/blocksuite/pull/831
- [ ] https://github.com/toeverything/blocksuite/pull/909
- [ ] https://github.com/toeverything/blocksuite/pull/920
```

```[tasklist]
## Fixes
- [ ] https://github.com/toeverything/blocksuite/pull/954
- [ ] https://github.com/toeverything/blocksuite/pull/953
- [ ] https://github.com/toeverything/blocksuite/pull/950
```

```[tasklist]
## Generate Data
- [ ] Multi-selection and click the button to create a database
- [x] Slash manu to create a database
- [x] #888
```

```[tasklist]
## Modify Data
- [x] Tag System https://github.com/toeverything/blocksuite/pull/734
- [x] model part initialization https://github.com/toeverything/blocksuite/pull/770
- [x] block part initialization https://github.com/toeverything/blocksuite/pull/770
- [ ] resort column
```

```[tasklist]
# Regression
- [x] Drag Handle https://github.com/toeverything/blocksuite/pull/854 https://github.com/toeverything/blocksuite/pull/957
[...]

PR #1997, opened 2023-04-09
refactor(database): separate title

zqran:
part of: #629


doodlewind (approved):
LGTM

packages/blocks/src/database-block/components/database-title.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    // for title placeholder
+    this.targetModel.title.yText.observe(() => {
+      this.requestUpdate();
+    });
+
+    // after the database structure is created
+    requestAnimationFrame(() => {
+      this._vEditor?.focusEnd();
+    });
```
lawvs, line 97: I believe we should not perform effects during the render process.

zqran: > It should be moved after the operation of creating the databse.

This is more intuitive. I will change it.

PR #2000, opened 2023-04-09
feat(database): add length limit to database title and select tag name

zqran:
https://user-images.githubusercontent.com/15389209/230917689-b2c66e46-daac-4aa4-9805-1558f5a4e4e4.mov

---
par of: #629


packages/blocks/src/database-block/utils.ts
```diff
@@ -61,3 +70,95 @@ export const getTagColor = tagColorHelper();
...
+type InitLimitedLengthVEditorConfig = {
+  yText: Y.Text | string;
+  container: HTMLElement;
+  targetModel: BaseBlockModel;
+  maxLength: number;
+  readonly?: boolean;
+  handlers?: VirgoEventService<BaseTextAttributes>['_handlers'];
+  options?: VEditorOptions;
+};
+export function initLimitedLengthVEditor({
```
doodlewind, line 83: It doesn't feel right requiring so much setup and handling events like `compositionend` to implement a length limit. How about supporting this as virgo configuration @Flrande ?

doodlewind (changes requested):
I'd consider it good to wait for better text API support, current implementation involves too much complexity.

Flrande:
> It doesn't feel right requiring so much setup and handling events like compositionend to implement a length limit. How about supporting this as virgo configuration @Flrande ?

I will do some refactor in virgo for it.

packages/blocks/src/database-block/utils.ts
```diff
@@ -61,3 +70,80 @@ export const getTagColor = tagColorHelper();
...
+    virgoInput: ctx => {
+      const length = vEditor.yText.length;
+      if (length >= maxLength && ctx.event.data) {
+        // prevent input
+        ctx.skipDefault = true;
+      }
+      return ctx;
+    },
+    paste: handlePaste,
+    virgoCompositionEnd: ctx => {
```
doodlewind, line 126: @Flrande this sill looks a bit leaky since we still require handling yText on upper layer for adding length limit, let's land this for now, and hope for future improvements to further simplify this.

/cc @Saul-Mirone

doodlewind (approved):

PR #2023, opened 2023-04-11
feat(store): support role in block schema

Saul-Mirone:


packages/store/src/workspace/database.ts
```diff
@@ -21,12 +21,16 @@ export class DatabaseManager {
   }
 
   protected get yCells() {
```
doodlewind, line 23: The migration to block level props looks great, while seems that we are still actually storing the database props on the root page block? We can move this in following PRs.

doodlewind (approved):

PR #2025, opened 2023-04-11
refactor(store): init internal props should respect ext in schema

Saul-Mirone:


doodlewind (commented):
I guess we don't actually need `ext` for now, having `prop:columns` and `prop:cells` in database block makes sense.

Saul-Mirone:
> I guess we don't actually need `ext` for now, having `prop:columns` and `prop:cells` in database block makes sense.

Will name it back to prop in future PRs, and remove codes about `ext`.

Issue #2036, opened 2023-04-12
Feature Request: support select all in text-block using double(triple) click


promer94:
Currently, double click text-block would only select the current word.

https://user-images.githubusercontent.com/15878786/231360591-b6983ef4-5380-47c6-a104-542adee2091c.mp4

It might be better if it could select the whole text content like notion

https://user-images.githubusercontent.com/15878786/231360887-a47698e9-7090-44aa-8d81-6fb61b733c8b.mp4

doodlewind:
Thanks for your feedback! Alternatively for now you can click the drag handle to select single block :)

In BlockSuite, word-level selection is done by overriding browser events, recognizing double click and partially select text by `Intl.Segmenter`. So this is highly controllable but not documented and public yet. From engineering perspective, it may make sense to have this behavior pluggable together with our SelectionManager. I hope to document the related patten behind this.

/cc @Saul-Mirone @Flrande

PR #2037, opened 2023-04-12
feat(selection): support triple-click to select all text of block

fundon:
Related to: #2036

https://user-images.githubusercontent.com/27926/232003284-b8571615-c897-4be5-a966-1c83e8175d80.mov


packages/blocks/src/page-block/default/selection-manager/default-selection-manager.ts
```diff
@@ -309,6 +310,14 @@ export class DefaultSelectionManager {
     showFormatQuickBarByDoubleClick(e, this.page, this.container, this.state);
   };
 
+  private _onContainerTripleClick = (e: SelectionEvent) => {
```
fundon, line 313: @lawvs You can add the logic of all the text  of the block in this method.

fundon: It mainly shows the logic of format-quick-bar and whether the content is correct when copying.

doodlewind (approved):
This is handled pretty well!

PR #2041, opened 2023-04-12
fix: unable to scroll to bottom

luke358:
After the content height changes, selecting an element results in the viewportElement height not being updated, which prevents the page from scrolling to the bottom. Perhaps listening to pageBlockContainer can fix this issue.

https://user-images.githubusercontent.com/48149577/231428127-2a901fa2-a1c6-425a-a637-fcb63286ca58.mov


luke358:
signed #2060

PR #2045, opened 2023-04-12
feat: selective cmd or ctrl

### LINKED ISSUE #2035: Support selective `cmd`/`ctrl` key binding based on OS platform
Currently we have taken both `cmd-Z` and `ctrl-Z` as hotkey:

https://github.com/toeverything/blocksuite/blob/master/packages/global/src/config/consts.ts#L13-L16

But the better behavior would be having `cmd-Z` on Mac and `ctrl-Z` on Windows or Linux.

This also applies for configurations on [block-hub](https://github.com/toeverything/blocksuite/blob/master/packages/global/src/config/consts/block-hub.ts) and [page format](https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/page-block/utils/const.ts) hotkeys.

lzlme:
resolve #2035

### DIFF
--- a/packages/blocks/src/page-block/utils/const.ts
+++ b/packages/blocks/src/page-block/utils/const.ts
@@ -5,6 +5,7 @@ import {
   InlineCodeIcon,
   ItalicIcon,
   LinkIcon,
+  SHORT_KEY,
   StrikethroughIcon,
   UnderlineIcon,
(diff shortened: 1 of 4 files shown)


doodlewind (approved):

doodlewind:
Quite a decent fix, thanks!

PR #2047, opened 2023-04-12
feat(edgeless): optimize background grid

alt1o:
https://user-images.githubusercontent.com/21084335/231506153-71ad5765-2f4a-4a30-95a1-e57bdd36f0d5.mov

### DIFF
--- a/packages/editor/src/components/editor-container.ts
+++ b/packages/editor/src/components/editor-container.ts
@@ -14,7 +14,7 @@ import {
 } from '@blocksuite/blocks';
 import { isFirefox, type Page, Slot } from '@blocksuite/store';
 import { html } from 'lit';
-import { customElement, property, query, state } from 'lit/decorators.js';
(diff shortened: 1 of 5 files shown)


packages/editor/src/components/editor-container.ts
```diff
@@ -45,8 +45,8 @@ export class EditorContainer extends WithDisposable(ShadowlessElement) {
     type: 'default',
   };
 
-  @state()
-  private showGrid = false;
+  @property()
+  showGrid = true;
```
doodlewind, line 49: We can land this for now, but I think this can better be defined a property of config on the edgeless page block

doodlewind (approved):

packages/editor/src/components/editor-container.ts
alt1o: `showGrid` could be a configuration option for the user's workspace. Each user has their own configuration.

PR #2053, opened 2023-04-13
refactor(store): move cell and column to database props and add database model

Saul-Mirone:


packages/blocks/src/database-block/database-model.ts
```diff
@@ -1,24 +1,185 @@
...
-import { defineBlockSchema, type SchemaToModel } from '@blocksuite/store';
+import type { Cell, Column, SelectTag } from '@blocksuite/global/database';
+import type { Text } from '@blocksuite/store';
+import { BaseBlockModel, defineBlockSchema, Y } from '@blocksuite/store';
 import { literal } from 'lit/static-html.js';
 
+export type Props = {
+  title: Text;
+  columns: string[];
+  yCells: Y.Map<Y.Map<unknown>>;
```
doodlewind, line 9: I hope we can hide the concept of `ySomething` in props (just renaming them would be fine), previously the `yCells` and `yColumns` are not serialized and are protected props on page.

And I think the previous `columns` field is not needed, since it was used to filter out the block-level columns from the global map.

We haven't landed the best practice handling nesting `YMap`, so having the `Y.Map<Y.Map<unknown>>` typing here is just fine, but hope we can make it better afterwards!

doodlewind (approved):
After all, this is a great step in having self-contained database block!

packages/blocks/src/database-block/database-model.ts
Saul-Mirone: > I hope we can hide the concept of `ySomething` in props (just renaming them would be fine), previously the `yCells` and `yColumns` are not serialized and are protected props on page.
> 
> And I think the previous `columns` field is not needed, since it was used to filter out the block-level columns from the global map.
> 
> We haven't landed the best practice handling nesting `YMap`, so having the `Y.Map<Y.Map<unknown>>` typing here is just fine, but hope we can make it better afterwards!

I agree that we need to improve our data structure transformation between `Y.*` and user friendly data structures. I'll do that in future PRs.

PR #2066, opened 2023-04-13
fix: avoid clicking on the icon to close

luke358:
https://user-images.githubusercontent.com/48149577/231768109-41f0934e-62b7-4cb1-bb57-9b14e814bf23.mov


doodlewind (approved):
Thanks for the fix!

PR #2078, opened 2023-04-14
fix: callback control for event removal

luke358:
When in the `SearchState.Searching state`, clicking outside clears the `onClickOutside event`, which should not be removed as intended. Is it possible to have more control over the removal of the event in the `callback`, with removal being the default behavior?

https://user-images.githubusercontent.com/48149577/231923509-5b402292-60ec-4a96-aca4-bc2cf87d579a.mov


doodlewind (approved):
LGTM

packages/blocks/src/database-block/utils.ts
```diff
@@ -19,14 +19,14 @@ function isVisible(elem: HTMLElement) {
 
 export function onClickOutside(
   element: HTMLElement,
-  callback: (element: HTMLElement) => void,
+  callback: (element: HTMLElement) => boolean | void,
```
doodlewind, line 22: `boolean | void` looks a bit weird, hope for improvement :)

zqran:
@luke358 

Thanks for your contribution! Currently, `Database` is still under development, and some details need to be confirmed and optimized. I will review in time after confirmation.

packages/blocks/src/database-block/utils.ts
luke358: yeah，but currently, the onClickOutside function is being used in other places where it doesn't return a value, which may cause conflicts. I haven't thought of a particularly good solution yet👀

zqran: @luke358 
I refactored `onClickOutside` and added a `reusable` parameter instead of the return `boolean` value of callback.

luke358: @zqran Yes, using parameters for control is also feasible, good idea!

zqran: Can you add some tests for this pr?

doodlewind (commented):
Well this seems to be good to land

packages/blocks/src/database-block/utils.ts
zqran: I've added it, you can check it out: #2089

PR #2086, opened 2023-04-14
fix(edgeless): using color codes instead of using actual color values

alt1o:


packages/blocks/src/page-block/edgeless/components/component-toolbar/change-brush-button.ts
```diff
@@ -19,10 +19,10 @@ import { createButtonPopper } from '../utils.js';
 
 function getMostCommonColor(
   elements: BrushElement[]
-): RawCssVariablesName | undefined {
```
doodlewind: @alt1o a minor style hint: our (subjective) convention prefer not to use the `undefined` literal explicitly. Generally `null` is more preferable.

doodlewind (approved):

doodlewind (approved):

PR #2088, opened 2023-04-14
feat(store): add array proxy

Saul-Mirone:


packages/blocks/src/database-block/database-model.ts
```diff
@@ -180,9 +228,8 @@ export const DatabaseBlockSchema = defineBlockSchema({
   flavour: 'affine:database',
   props: (internal): Props => ({
     title: internal.Text(),
-    columns: [],
     yCells: internal.Map<Y.Map<unknown>>(),
```
doodlewind, line 231: Is it possible renaming `yCells` -> `cells`?

Saul-Mirone: Yeah, and we'll have an object proxy for it.

Himself65 (commented):
unit test case?

Saul-Mirone:
> Member
> ###

Done

Saul-Mirone:
> unit test case?

Done

doodlewind (approved):

PR #2090, opened 2023-04-14
fix(database): should the indicator display correctly when resize the window

zqran:
before:

https://user-images.githubusercontent.com/15389209/232023197-bfe1fd0c-1d77-47bd-b540-7825993f2d2a.mov

after:

https://user-images.githubusercontent.com/15389209/232023254-f0be6104-9e17-489f-a0f1-012487b99da0.mov

--- 

part of: #629


doodlewind (approved):
I think it's also fine not to add test case for minor tweak like this, unless it's somewhere easy to have regression.

zqran:
> I think it's also fine not to add test case for minor tweak like this, unless it's somewhere easy to have regression.

Get!

PR #2098, opened 2023-04-15
fix: onShowSearch alter the searchState

luke358:
https://user-images.githubusercontent.com/48149577/232205825-93383109-a44e-44bb-965a-1d95eacd328c.mov


luke358:
The main cause of this issue is likely the modification of the searchState. When in the Searching or SearchInput state, the searchInput is open and should not be triggered repeatedly. Furthermore, repeated triggering can result in the duplication of the onClickOutside event registration.

_searchInput to  _searchContainer: related: #2066. It seems that some conflicts have been overwritten.

zqran:
> The main cause of this issue is likely the modification of the searchState. When in the Searching or SearchInput state, the searchInput is open and should not be triggered repeatedly. Furthermore, repeated triggering can result in the duplication of the onClickOutside event registration.
> 
> _searchInput to _searchContainer: related: #2066. It seems that some conflicts have been overwritten.

Thanks for your feedback, I'm reviewing your changes. Now, after showing the search box, clicking the search icon again should do noting. I will change it.

zqran:
Since `Database` is still under development, some interaction details still need to be confirmed, so we may modify your code implementation. Alternatively, you can open an issue with your suggestions.

Anyway, thanks a lot for your contribution!

zqran (approved):

luke358:
> Since `Database` is still under development, some interaction details still need to be confirmed, so we may modify your code implementation. Alternatively, you can open an issue with your suggestions.
> 
> Anyway, thanks a lot for your contribution!

yeah，it's ok.

doodlewind (approved):
Thanks for your help improving database block! FYI, feel free to join our discussion group at https://t.me/AffineDevCN if you are a Chinese speaker 😀

Issue #2005, opened 2023-04-10
Balloon toolbar doesn't update on keyboard selection change


lawvs:
The status of the balloon toolbar does not update when the selection is changed using the keyboard.

https://user-images.githubusercontent.com/18554747/230827547-379e71b0-f27e-4e30-ae22-b07dd6ef3999.mov

PR #2100, opened 2023-04-15
fix(format-bar): update state on change in keyboard selection

PerfectPan:
Close https://github.com/toeverything/blocksuite/issues/2005

And fix the incorrect test. 
In the following picture, you can see the correct left position for quick format bar should be 377 instead of 395. It is wrong before I fix this problem.

CleanShot 2023-04-15 at 20 19 33@2x

The key point is "convert-to-database" button is shown when range type is block(https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/page-block/utils/const.ts#L143). 

But the range type is changed asynchronous after the render, and before the next render which is caused by update position there is no another render. The following picture shows that the quick bar has already been changed but the database button is not shown.

CleanShot 2023-04-15 at 20 07 04@2x

So when the update position function is called, it used the incorret dom to calculate. https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/components/format-quick-bar/index.ts#L86
The bounding client rect's width only includes two buttons.

However, after I fix the problem, it will rerender before to call update position function because I change the reactive state this._format. This time the range type is block and "convert-to-database" button is shown so that we can calculate the correct position.

The inner logic seems depend on the correct timing, which is a little tricky. Maybe there is a better design to avoid that happens.

### DIFF
--- a/packages/blocks/src/components/format-quick-bar/format-bar-node.ts
+++ b/packages/blocks/src/components/format-quick-bar/format-bar-node.ts
@@ -12,7 +12,10 @@ import { customElement, property, query, state } from 'lit/decorators.js';
 import { styleMap } from 'lit/directives/style-map.js';
 
 import type { AffineTextAttributes } from '../../__internal__/rich-text/virgo/types.js';
-import { restoreSelection } from '../../__internal__/utils/block-range.js';
+import {
+  getCurrentBlockRange,
+  restoreSelection,
+} from '../../__internal__/utils/block-range.js';
 import {
   getRichTextByModel,
   WithDisposable,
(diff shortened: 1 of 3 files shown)


lawvs (approved):
Quite a decent fix, thanks!

lawvs:
> However, after I fix the problem, it will rerender before to call update position function because I change the reactive state this._format. This time the range type is block and "convert-to-database" button is shown so that we can calculate the correct position.


The rect of the format bar is mostly used to prevent it from overflowing outside of the screen and for centering purposes. I will be careful with this scenario in the follow-up development. Thank you very much for your detailed explanation.

PR #2102, opened 2023-04-15
docs: remove line continuation characters in installation command

pionxzh:
This PR removes the line continuation characters(`\`) in documents as it will fail on other platforms/command lines.

| System | Line continuation character |
| --- | --- |
| Windows command prompt | `^` |
| PowerShell | `` ` ``
| Bash | `\` |


doodlewind:
Thanks for helping out! However, this style is used on purpose for better readability (subjectively). Instead of having a single long line of bash command, splitting it into separate lines can make it more clear what it's actually doing.

PR #2107, opened 2023-04-16
feat(virgo): add `setText` api

Flrande:
related to https://github.com/toeverything/blocksuite/pull/2093


packages/virgo/src/virgo.ts
```diff
@@ -359,6 +359,16 @@ export class VEditor<
     });
   }
 
+  coverText(
```
doodlewind, line 362: `coverText` -> `setText`

doodlewind (approved):

PR #2124, opened 2023-04-17
feat(edgeless): mobile compatible

alt1o:
Using pointer events instead of mouse events for mobile development to achieve better compatibility

https://user-images.githubusercontent.com/21084335/232424988-7829db99-393e-4464-a3ef-802d541b616f.mov


alt1o:
fixed by #2269

PR #2128, opened 2023-04-17
refactor(store): use map proxy for object props

Saul-Mirone:


packages/blocks/src/database-block/database-model.ts
```diff
@@ -1,11 +1,11 @@
...
-import type { Text } from '@blocksuite/store';
-import { BaseBlockModel, defineBlockSchema, Y } from '@blocksuite/store';
+import { Text } from '@blocksuite/store';
+import { BaseBlockModel, defineBlockSchema } from '@blocksuite/store';
 import { literal } from 'lit/static-html.js';
 
 export type Props = {
   title: Text;
-  yCells: Y.Map<Y.Map<unknown>>;
+  cells: SerializedCells;
```
himself65, line 8: breaking changes

doodlewind (approved):

PR #2130, opened 2023-04-17
chore: simplify array return and fix typo

pionxzh:


doodlewind (approved):
Thanks for improving our code quality!

PR #2135, opened 2023-04-17
fix(edgeless): resize-Manager not consider zoom

blurname:
before:

https://user-images.githubusercontent.com/50226652/232549652-f7042c49-2f97-4e66-8269-e346405af5d4.mp4

after:

https://user-images.githubusercontent.com/50226652/232549677-5d5769fe-0697-4813-be39-e830703de06d.mp4


doodlewind (approved):
Thanks for the fix!

PR #2153, opened 2023-04-19
fix: delete emoji completely

### LINKED ISSUE #2138: Emoji gets partially deleted when pressing backspace
Pressing backspace will partially delete some emoji characters.
For example: 
🙅‍♂️ (no_good_men) will become 🙅 (no_good_woman)
🏳️‍🌈 will become 🏳️‍

This is a common problem for editors as the length of the string characters and the visual letters are not aligned.
```js
'🌷'.length === 2
'🙅‍♂️'.length === 5
'🏳️‍🌈'.length === 6
```

I noticed this by deleting the gear emoji in the playground.
```
⚙️ Serve as E2E test entry.
 ^
```

lzlme:
Fix #2138 , but there is a small problem, `Intl.Segmenter` is not compatible with Firefox, one solution to use [polyfill](https://github.com/surferseo/intl-segmenter-polyfill), another solution is to use lodash directly, which one do you prefer?

### DIFF
--- a/packages/virgo/src/utils/transform-input.ts
+++ b/packages/virgo/src/utils/transform-input.ts
@@ -57,13 +57,13 @@ function handleDelete(vRange: VRange, editor: VEditor) {
     }
 
     if (vRange.index > 0) {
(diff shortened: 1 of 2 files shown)


doodlewind (approved):

doodlewind:
Thanks for handling such a detailed improvement!

Issue #2147, opened 2023-04-18
blobOptionsGetter shouldn't have a default to API


pionxzh:
When I import markdown by:

```js
const contentParser = new ContentParser(page)
const pageBlockId = page.addBlock('affine:page')
const frameId = page.addBlock('affine:frame', {}, pageBlockId)
contentParser.importMarkdown(markdown, frameId)
//             ^^^^^^^^^^^^
```

If there are any image element, it will automatically call `workspace.blobs` (the `BlobStorage`) to store the blob.

But the default value of `_blobOptionsGetter` caused the parser to send a request to `/api/workspace/{workspace_id}/blob` to persistent the blob on the server, which shouldn't be enabled by default.

See 

https://github.com/toeverything/blocksuite/blob/c673f6637d9d0fc32147b931b37c1e4932cbc027/packages/blocks/src/__internal__/content-parser/parse-html.ts#L468

https://github.com/toeverything/blocksuite/blob/c673f6637d9d0fc32147b931b37c1e4932cbc027/packages/store/src/workspace/workspace.ts#L35-L37

## Current workaround
```diff
const workspace = new Workspace({
    id: workspaceId,
    providers: [],
+   blobOptionsGetter: () => undefined,
})
```

doodlewind:
Closed by #2169

PR #2169, opened 2023-04-19
feat: init new blob storage

### LINKED ISSUE #1940: refactor blob storage
```[tasklist]
### Tasks
- [ ] https://github.com/toeverything/blocksuite/pull/1166
```

himself65:
Fixes: https://github.com/toeverything/blocksuite/issues/1940, https://github.com/toeverything/blocksuite/issues/2147

Usage

```ts
const blobStorages = [createMemoryStorage(id), createIndexeddbStorage(id)]
```

Known issue:
- Blob syncing state has been lost, need to be rewritten in AFFiNE side
- Indexeddb Test has been lost, need to be rewritten in new logic


packages/blocks/src/embed-block/image/image-block.ts
```diff
@@ -136,73 +135,26 @@ export class ImageBlockComponent extends ShadowlessElement {
...
-      url = await storage.get(this.model.sourceId);
-    }
-    if (url) {
-      this._source = url;
-      this._imageState = 'ready';
-    } else {
+    try {
+      const blob = await storage.get(this.model.sourceId);
+      if (blob) {
+        this._source = URL.createObjectURL(blob);
```
pengx17, line 151: we may also need to revoke the blob url when the image is no longer being used.

packages/store/src/workspace/workspace.ts
```diff
@@ -58,19 +56,33 @@ export class Workspace {
...
+      get: async id => {
+        const result = await Promise.allSettled(
+          this._storages.map(s => s.crud.get(id))
+        );
+        const answer = result.find(
+          (result): result is PromiseFulfilledResult<Blob | null> =>
+            result.status === 'fulfilled'
+        );
+        return answer?.value ?? null;
+      },
```
pengx17, line 74: May be we can race it here so that the first one settled will be used?

packages/blocks/src/embed-block/image/image-block.ts
himself65: good catch

packages/store/src/workspace/workspace.ts
himself65: No. Because some providers will return `null`. You have to wait for all response and find the first one that is not `null`

Oh, I see my logic problem here

pengx17: something like this. Resolve the result when the first one resolved with a value.

        let found = false;
        let count = 0
        return new Promise(res => {
          this._storages.forEach(s => s.crud.get(id).then(result => {
            if (result.value && !found) {
              found = true;
              res(blob);
            }
            if (++count === this._storages.length && !found) {
              res(null)
            }
          }))
        })

himself65: thanks

pengx17 (approved):

tests/embed.spec.ts
```diff
@@ -137,126 +137,6 @@ async function initMockImage(page: Page) {
   });
 }
 
-/**
```
himself65, line 140: These tests have removed since we move the logic to the affine side

packages/playground/examples/blob/index.html
```diff
@@ -33,7 +33,8 @@
     </div>
     <br />
     <script type="module">
-      import '@blocksuite/store/persistence/blob/__tests__/test-entry';
+      // todo: move this back later
```
himself65: move this back

doodlewind (approved):

doodlewind:
It looks good to deprecate and remove the IndexedDB blob storage test, they are staled and not checked in CI for now.

himself65:
> It looks good to deprecate and remove the IndexedDB blob storage test, they are staled and not checked in CI for now.

I see we check the image status in existing playwright, so it's okay to remove it

PR #2178, opened 2023-04-20
test: import from node environment

### LINKED ISSUE #2176: Add Node environment compatibility test
For now some imports from `@blocksuite/blocks` are fragile to become incompatible with Node (e.g., `@blocksuite/blocks/content-parser` and `@blocksuite/blocks/models`). Since BlockSuite is published in a bundle-less way (running `tsc` only), we can't track such case in our build process.

So, it will be good to add CI checks in `playground` that imports these package entries in Node:

* `@blocksuite/store`
* `@blocksuite/blocks/content-parser`
* `@blocksuite/blocks/models`

And check no error is thrown. This helps ensuring our Node compatibility. And we can remove the `@blocksuite/react` after this is shipped, because this package was mainly used for validating this.

Himself65:
Closes #2176


doodlewind (approved):

.github/workflows/test.yaml
```diff
@@ -23,6 +23,41 @@ on:
       - 'packages/docs/**'
 
 jobs:
+  test-node:
+    name: Node.js Environment Compatibility
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
lawvs, line 32: I think this step is unnecessary.
```suggestion
```

.github/workflows/test.yaml
```diff
@@ -23,6 +23,40 @@ on:
       - 'packages/docs/**'
 
 jobs:
+  test-node:
+    name: Node.js Environment Compatibility
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
```
lawvs, line 31:
```suggestion
```

Issue #2143, opened 2023-04-18
Incorrect behavior when paste code from outside


lawvs:
When pasting the following code, there are some issue

- The language of code is incorrect.
- The pasted content is incomplete.

```js
'🌷'.length === 2
'🙅‍♂️'.length === 5
'🏳️‍🌈'.length === 6
```

<img width="963" alt="Screenshot 2023-04-18 at 11 48 21 AM" src="https://user-images.githubusercontent.com/18554747/232666219-fa6c3a96-b279-4d94-b777-d83bcbcce200.png">

Related to https://github.com/toeverything/blocksuite/issues/1529

Related code https://github.com/toeverything/blocksuite/blob/23e41d035bf6970369c33849465eb0b46e142cc2/packages/blocks/src/__internal__/clipboard/utils.ts#L152-L155

zuozijian3720:
It seems that the current code logic cannot handle GitHub's code block format
<img width="808" alt="image" src="https://user-images.githubusercontent.com/17165520/233144611-d4c4edbb-91c8-4dd6-9794-61538ef9ae2f.png">
<img width="683" alt="image" src="https://user-images.githubusercontent.com/17165520/233144761-1e5eb8c6-4d66-4511-8097-1a9c2847b3cd.png">
I can submit a PR to fix this bug if support GitHub's code block format is necessary.
There are two class names, such as "highlight highlight-source-js", in the parent element of the "pre" element that can be used to detect the language.

lawvs:
> I can submit a PR to fix this bug if support GitHub's code block format is necessary.
There are two class names, such as "highlight highlight-source-js", in the parent element of the "pre" element that can be used to detect the language.

Adapting HTML for all websites that users may copy from is impractical. Therefore, I believe that parsing the code block and treating all content inside of code as plain text is the best option in this case.

cc @QiShaoXuan

PR #2185, opened 2023-04-20
fix: incorrect behavior when paste code from GitHub

zuozijian3720:
Fixes: https://github.com/toeverything/blocksuite/issues/2143

Nodes that do not match will be treated as plain text.


doodlewind (approved):
Thanks for the fix!

PR #2190, opened 2023-04-20
refactor: subpage meta sync

lawvs:
Part of https://github.com/toeverything/AFFiNE/issues/1332

Redesign `subpageIds` about https://github.com/toeverything/blocksuite/pull/1716


lawvs:
@Himself65 This pull request is not related to sub doc, but instead concerns subpage.

the Subpage is a concept in a product. It has the following feature:

- Each subpage can only have one parent page.
- When deleting a subpage, the subpage node in the parent page should be deleted synchronously.
- When deleting the subpage node in the page, the corresponding subpage will be put into the trash.

PR #2197, opened 2023-04-20
fix: small cleanup related to clipboard

thorseraq:


packages/blocks/src/__internal__/content-parser/index.ts
```diff
@@ -77,8 +78,8 @@ export class ContentParser {
     this.slots.beforeHtml2Block.emit(htmlEl);
     return this._convertHtml2Blocks(htmlEl);
   }
+
   async file2Blocks(clipboardData: DataTransfer) {
-    const { getFileFromClipboard } = await import('../clipboard/utils.js');
```
thorseraq, line 81: async importing will cause losing `clipboardData`, referencing https://stackoverflow.com/questions/66621194/access-drag-n-drop-object-inside-promise-then by @alt1o 


https://user-images.githubusercontent.com/20554850/233381211-4ef9a273-f1ce-4022-b16f-fd446bce1e9c.mov

packages/blocks/src/__internal__/content-parser/index.ts
```diff
@@ -4,6 +4,7 @@ import { Slot } from '@blocksuite/store';
 import { marked } from 'marked';
 
 import type { PageBlockModel } from '../../models.js';
+import { getFileFromClipboard } from '../clipboard/index.js';
```
doodlewind, line 7: This breaks Node compatibility. Any idea to workaround?

thorseraq: It may be caused by side effects importing in `packages/blocks/src/__internal__/clipboard/utils.ts`.

I'm trying to extract a new util file that keeps no dependencies on others. Make it can be relied on by other file such as `packages/blocks/src/__internal__/content-parser/index.ts`

doodlewind (approved):

PR #2206, opened 2023-04-20
fix(drag-handle): optimize position of indicator in database

fundon:
https://user-images.githubusercontent.com/27926/233495956-7e36dfd3-7578-4414-8428-f0f76fa01a95.mov

### DIFF
--- a/packages/blocks/src/__internal__/utils/query.ts
+++ b/packages/blocks/src/__internal__/utils/query.ts
@@ -930,6 +930,12 @@ export function getDatabaseBlockRowsElement(element: Element) {
   return element.querySelector('.affine-database-block-rows');
 }
 
+export enum DropFlags {
+  Normal,
+  Database,
+  EmptyDatabase,
+}
+
 /**
  * Gets the drop rect by block and point.
  */
@@ -939,12 +945,11 @@ export function getDropRectByPoint(
   element: Element
 ): {
   rect: DOMRect;
-  flag: number;
+  flag: DropFlags;
 } {
   const result = {
     rect: getRectByBlockElement(element),
-    // 0: others, 1: empty database, 2: database
-    flag: 0,
+    flag: DropFlags.Normal,
   };
 
   const isDatabase = matchFlavours(model, ['affine:database'] as const);
@@ -960,7 +965,7 @@ export function getDropRectByPoint(
   if (tempElement) {
     // If the database is empty
     if (isDatabase && model.isEmpty()) {
-      result.flag = 1;
+      result.flag = DropFlags.EmptyDatabase;
       const table = getDatabaseBlockTableElement(element);
       assertExists(table);
       const bounds = table.getBoundingClientRect();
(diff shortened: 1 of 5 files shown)


packages/blocks/src/__internal__/utils/query.ts
```diff
@@ -931,29 +931,47 @@ export function getDropRectByPoint(
...
   element: Element
-) {
+): {
+  rect: DOMRect;
+  flag: number;
+} {
   const result = {
     rect: getRectByBlockElement(element),
-    // 0: others, 1: empty database
+    // 0: others, 1: empty database, 2: database
```
doodlewind, line 940: I think it's better to use a typed `type` field (or enums) instead of hard-coded integers.

zqran:
image

@fundon 
This breaks the indicator for changing column width, let me check why.

zqran (approved):

doodlewind (approved):

zqran:
@fundon 
I resolved some conflicts (some related to database), please confirm if there is any problem.

packages/blocks/src/database-block/components/row-container.ts
```diff
@@ -56,7 +56,7 @@ export function DataBaseRowContainer(
         border-right: 1px solid var(--affine-border-color);
         transform: translateX(0);
       }
-      .affine-database-block-row-cell-content > .affine-block-element {
+      .affine-database-block-row-cell-content > affine-paragraph {
         width: 100%;
```
fundon, line 60: @zqran I think it would be better to use the `.affine-block-element` class here for better compatibility.
Currently, users can drag and drop any block (except database) into the database.

zqran: Get! 

But one thing needs to be mentioned, the title column should only be a block of type `text`.

Others, like code, image, quote, database, etc., should not appear in the title column.

fundon: > Others, like code, image, quote, database, etc., should not appear in the title column.

Need to be clear here. 
/cc @doodlewind

fundon: We can add a filter to only allow text block to be added to the database.

zqran: Now there is a schema validator for blocks. See #2291.

fundon: So do we need to filter dragging blocks?

doodlewind (approved):

packages/blocks/src/components/drag-handle.ts
```diff
@@ -494,36 +494,54 @@ export class DragHandle extends WithDisposable(LitElement) {
     );
   }
 
-  static calcTarget(
+  /**
+   * Calculates the drop target.
+   */
+  static calcDropTarget(
```
doodlewind, line 500: I hope this method can be split into smaller pieces, current implementation is a bit too heavy.

doodlewind (approved):

packages/blocks/src/components/drag-handle.ts
fundon: Ok, I will refactor it.

PR #2207, opened 2023-04-20
refactor: use `@toeverything/y-indexeddb`

Himself65:


doodlewind (approved):

packages/store/src/persistence/doc/index.ts
```diff
@@ -2,7 +2,6 @@ import type { Awareness } from 'y-protocols/awareness';
 import type * as Y from 'yjs';
 
 export { DebugDocProvider } from './debug-provider.js';
```
doodlewind, line 4: Perhaps we can also move the WebRTC-based `DebugDocProvider` to playground

doodlewind (approved):

Issue #2123, opened 2023-04-17
Cursor behaves unexpectedly in quote block when Enter is pressed and undo


SkyeSun:
# Demo

https://user-images.githubusercontent.com/16323946/232418624-bbc253a9-eb41-440f-9e8f-ca31a5c17b63.mov

# Step to reproduce
1. Add a `Quote` block.
2. Type any character.(unnecessary)
3. Press Enter.(x2)
4. Type any character.
5. Press Enter.
6. Undo.

# Result
The cursor will fallback one character after undo. If we put more empty lines before current editing line, the number of fallback characters will be the same as the total number of empty lines before current line in this quote block.

# Expected
The cursor should always be at the position when we pressed Enter after undo.

# Environment
- OS: Mac OS Monterey (Version 12.3)
- Browser: Chrome (Version 112.0.5615.49)

SkyeSun:
It seems that this is not a `quote` block problem, more like a `rich-text` problem. Because it can also reproduced in plain `rich-text`.


https://user-images.githubusercontent.com/16323946/232425344-1ada5571-7d6d-4bee-88ba-e0fdfd6b6057.mov

## Step to reproduce
1. Press `Shift` + `Enter`.
2. Type any characters.
3. Press `Enter`.
4. Undo.

zuozijian3720:
Seems this bug is because ZERO_WIDTH_SPACE has only one cursor position, but is counted as two.

https://github.com/toeverything/blocksuite/blob/64520c8a7a824c28dbb26b00333e6e135fd9ba29/packages/virgo/src/virgo.ts#L204-L224

PR #2223, opened 2023-04-21
fix(virgo): getTextPoint offset calculation

zuozijian3720:
Currently, ZERO_WIDTH_SPACE will count as two cursor position, but it should only count as one.
Fixes: https://github.com/toeverything/blocksuite/issues/2123


lawvs (commented):
Could you add a test for it?

zuozijian3720:
> Could you add a test for it?

Yes, I'll try it

Flrande:
If you find that the local test passes but the pipeline fails, it may be due to abnormal logs that are ignored by the local environment.

tests/rich-text.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+  await page.waitForTimeout(100);
+  await page.keyboard.down('Shift');
+  await press(page, 'Enter');
+  await page.keyboard.up('Shift');
+  await page.waitForTimeout(100);
+  await type(page, 'asdfg');
+  await page.waitForTimeout(100);
+  await press(page, 'Enter');
+  await page.waitForTimeout(100);
+  await page.keyboard.down('Meta');
```
lawvs, line 26: The testing platform is Linux, so there is no Meta key.

You can use `SHORT_KEY` instead, Or you can use the following pattern

`await page.keyboard.press(`${SHORT_KEY}+Enter`, { delay: 50 });`

lawvs (approved):

doodlewind (approved):

zuozijian3720:
@lawvs Thanks very much

PR #2224, opened 2023-04-21
fix(selection): responsive image block

fundon:
https://user-images.githubusercontent.com/27926/233637464-ca608eb2-3809-4e0e-9b4d-7f918bf73c7a.mov

In the video you will see that the embed-resize-manager(or block-level selection) may appear misaligned, this is caused by transition in AFFiNE.

Screenshot 2023-04-21 at 8 37 34 PM


doodlewind (approved):
Thanks!

Issue #2136, opened 2023-04-17
The same linked page node may be accidentally merged


lawvs:
https://user-images.githubusercontent.com/18554747/232550754-6674ed7e-8a68-4945-9998-5244be5dbafd.mov

PR #2231, opened 2023-04-22
fix(link): consecutive identical reference nodes should not be merged

PerfectPan:
Close https://github.com/toeverything/blocksuite/issues/2136

Problem is that same attribute text nodes will be merged in yjs in delta mode(https://github.com/toeverything/blocksuite/blob/master/packages/virgo/src/services/delta.ts#L159). We should handle this case.
```js
yText.insert(0, ' ', { type: 'LinkedPage', pageId: 'page0' });
yText.insert(0, ' ', { type: 'LinkedPage', pageId: 'page0' });

console.log(yText.toDelta())
/**
[
  { insert: '  ', attributes: { type: 'LinkedPage', pageId: 'page0' } }
]
*/
```


lawvs (approved):
Great solution, thank you very much.

PR #2232, opened 2023-04-22
fix: reference node ime compatibility

PerfectPan:
before:

https://user-images.githubusercontent.com/24316656/233796234-94b3da3b-4ce5-4294-a2a4-15a7234d3e21.mp4

after:

https://user-images.githubusercontent.com/24316656/233796036-ba466119-3bfe-4878-b36a-e6093a919ca9.mp4

### DIFF
--- a/packages/blocks/src/__internal__/rich-text/reference-node.ts
+++ b/packages/blocks/src/__internal__/rich-text/reference-node.ts
@@ -69,8 +69,12 @@ export class AffineReference extends WithDisposable(ShadowlessElement) {
       white-space: break-spaces;
     }
 
+    .affine-reference-title {
+      color: var(--affine-text-primary-color);
+    }
     .affine-reference-title::before {
       content: attr(data-title);
+      color: var(--affine-link-color);
     }
   `;
 
--- a/packages/blocks/src/__internal__/rich-text/rich-text.ts
+++ b/packages/blocks/src/__internal__/rich-text/rich-text.ts
@@ -1,5 +1,5 @@
 import { assertExists, type BaseBlockModel } from '@blocksuite/store';
-import { VEditor } from '@blocksuite/virgo';
+import { VEditor, ZERO_WIDTH_NON_JOINER } from '@blocksuite/virgo';
 import { css, html } from 'lit';
 import { customElement, property, query } from 'lit/decorators.js';
 
@@ -135,6 +135,29 @@ export class RichText extends ShadowlessElement {
           }
 
           ctx.attributes = attributes ?? null;
+        }
+
+        if (vRange.index >= 0) {
+          const selection = window.getSelection();
+          if (selection && selection.rangeCount !== 0) {
+            const range = selection.getRangeAt(0);
+            const container = range.startContainer;
+
+            if (
+              container instanceof Text &&
+              container.parentElement?.dataset.virgoText === 'true'
+            ) {
+              const [text] = vEditor.getTextPoint(vRange.index);
+              const affineReference =
+                text.parentElement?.closest('affine-reference');
+              if (
+                affineReference &&
+                text.textContent !== ZERO_WIDTH_NON_JOINER
+              ) {
+                text.textContent = ZERO_WIDTH_NON_JOINER;
+              }
+            }
+          }
         }
         return ctx;
       },
--- a/packages/blocks/src/__internal__/rich-text/rich-text.ts
+++ b/packages/blocks/src/__internal__/rich-text/rich-text.ts
@@ -3,7 +3,7 @@ import { VEditor, ZERO_WIDTH_NON_JOINER } from '@blocksuite/virgo';
 import { css, html } from 'lit';
 import { customElement, property, query } from 'lit/decorators.js';
 
-import { type BlockHost } from '../utils/index.js';
+import { type BlockHost, getCurrentNativeRange } from '../utils/index.js';
 import { ShadowlessElement } from '../utils/lit.js';
 import { setupVirgoScroll } from '../utils/virgo.js';
 import { InlineSuggestionController } from './inline-suggestion.js';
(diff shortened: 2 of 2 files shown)


Flrande:
Thank you very much for your fix! However, we would like the virgo package and the blocks package to be decoupled, which means that there should be no blocks-related code in virgo. We can choose to override the default handler of virgo to make it meet our requirements. For example, in this use case, we can use `bindHandlers` and `virgoCompositionEnd` to override the behavior of virgo in the blocks package.

PerfectPan:
> Thank you very much for your fix! However, we would like the virgo package and the blocks package to be decoupled, which means that there should be no blocks-related code in virgo. We can choose to override the default handler of virgo to make it meet our requirements. For example, in this use case, we can use `bindHandlers` and `virgoCompositionEnd` to override the behavior of virgo in the blocks package.

Thanks for your advise! I will reconsider how to solve this issue following the override pattern.

PerfectPan:
> > Thank you very much for your fix! However, we would like the virgo package and the blocks package to be decoupled, which means that there should be no blocks-related code in virgo. We can choose to override the default handler of virgo to make it meet our requirements. For example, in this use case, we can use `bindHandlers` and `virgoCompositionEnd` to override the behavior of virgo in the blocks package.
> 
> Thanks for your advise! I will reconsider how to solve this issue following the override pattern.

After I read the code and think a lot, I am not sure whether we need to open a new custom way for such a ime solution. Because I found that the text attribute rendering is customized in Virgo, so ime solution must be bound to the attribute rendering we define. However, virgoCompositionEnd handler may be more coarse-grained, something like code block may do not need this logic because the textSchema is different between code block and paragraph code.

Since I have no godd idea for this situation to abstract a proper handler to bind, I just move the logic into the rich-text element.

packages/blocks/src/__internal__/rich-text/rich-text.ts
```diff
@@ -136,6 +136,29 @@ export class RichText extends ShadowlessElement {
...
+              const [text] = vEditor.getTextPoint(vRange.index);
+              const affineReference =
+                text.parentElement?.closest('affine-reference');
+              if (
+                affineReference &&
+                text.textContent !== ZERO_WIDTH_NON_JOINER
+              ) {
+                text.textContent = ZERO_WIDTH_NON_JOINER;
+              }
+            }
```
lawvs, line 159: I do not advocate special treatment for use cases, as this does not completely solve the problem and makes the `rich-text` difficult to maintain.

@Flrande Do you have any other effective approaches for this situation?

lawvs (changes requested):

packages/blocks/src/__internal__/rich-text/reference-node.ts
```diff
@@ -69,8 +69,12 @@ export class AffineReference extends WithDisposable(ShadowlessElement) {
       white-space: break-spaces;
     }
 
+    .affine-reference-title {
+      color: var(--affine-text-primary-color);
+    }
     .affine-reference-title::before {
       content: attr(data-title);
+      color: var(--affine-link-color);
```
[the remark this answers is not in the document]
PerfectPan: :before pesudo element is render the reference title, which is blue. Originally it's color is inherit from .affine-reference class, but it will affect character's color that ime insert. I want it to be consistent with text style. So to solve this problem, I change the css.
CleanShot 2023-04-24 at 09 53 13@2x

CleanShot 2023-04-24 at 09 53 32@2x

packages/blocks/src/__internal__/rich-text/rich-text.ts
```diff
@@ -136,6 +136,29 @@ export class RichText extends ShadowlessElement {
 
           ctx.attributes = attributes ?? null;
         }
+
+        if (vRange.index >= 0) {
+          const selection = window.getSelection();
+          if (selection && selection.rangeCount !== 0) {
+            const range = selection.getRangeAt(0);
```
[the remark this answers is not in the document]
PerfectPan: Done, thanks.

packages/blocks/src/__internal__/rich-text/rich-text.ts
PerfectPan: I understand your concerns just as I comment. So maybe we could find a more universal approach.
I consider some way last night:
1. We add the custom ime handler to override in handler context https://github.com/toeverything/blocksuite/blob/master/packages/virgo/src/services/event.ts#L11
```ts
export interface VHandlerContext<
  T extends BaseTextAttributes,
  E extends Event = Event
> {
  event: E;
  data: string | null;
  vRange: VRange;
  skipDefault: boolean;
  attributes: T | null;
  imeHandler: () => boolean; // boolean represent whether we need to use the default way to solve
}
```
But it may pollute other handler since it may only occured in compositionend.

2. We add the custom ime handler to as part of the virgo handler we bind. https://github.com/toeverything/blocksuite/blob/master/packages/virgo/src/services/event.ts#L30
```ts
private _handlers: {
    keydown?: (event: KeyboardEvent) => void;
    paste?: (event: ClipboardEvent) => void;
    // corresponding to native input event and used to take over default behavior in virgo
    virgoInput?: (
      ctx: VHandlerContext<TextAttributes, InputEvent>
    ) => VHandlerContext<TextAttributes, InputEvent>;
    // corresponding to native compositionend event and used to take over default behavior in virgo
    virgoCompositionEnd?: (
      ctx: VHandlerContext<TextAttributes, CompositionEvent>
    ) => VHandlerContext<TextAttributes, CompositionEvent>;
    imeHandler: () => boolean;
  } = {};
```
But I think imeHandler is contained by virgoCompositionEnd, and it is not the same level with other handler.
3. Maybe we can find a way to make reference node implementation use v-text as well https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/__internal__/rich-text/reference-node.ts#L176. And it may also recommend any other custom render use v-text to make the logic more consistent.

Flrande: > I do not advocate special treatment for use cases, as this does not completely solve the problem and makes the rich-text difficult to maintain.
> 
> @Flrande Do you have any other effective approaches for this situation?

Before we implement a better abstraction of Virgo's exposed interface, I think we can make this fix take effect.

Flrande (approved):

lawvs (approved):

PR #2198, opened 2023-04-20
fix: the vRange is wrong in some cases

zuozijian3720:
Fixes: https://github.com/toeverything/blocksuite/issues/1901

This PR recompute vRange before each 'beforeinput' trigger

If the problem can be resolved. Then we can consider how to optimize the performance of this code.

PR #2238, opened 2023-04-23
fix: `selection` wrong when type many times in one frame

zuozijian3720:
Fix the bug caused by https://github.com/toeverything/blocksuite/pull/2198
Add test

This bug will cause the wrong cursor position to appear when you type many times in one moment.

Kapture 2023-04-23 at 16 07 51


Flrande (approved):

doodlewind (approved):
Good catch, thanks!

PR #2253, opened 2023-04-24
feat(playground): example for store usage

### LINKED ISSUE #2071: Example demonstrating `@blocksuite/store` usage
Based on the [docs](https://block-suite.com/workspaces-and-pages.html), it's now easy to reuse only `@blocksuite/store` to build a block-based app. Steps:

* Define some custom block flavours.
* Listen to the `page.rootAdded` event slot, bind the `page.root` to framework component.
* Listen to the `model.propsUpdated` and `model.childrenUpdated` event slot, update the UI framework component.
* Call APIs like `page.addBlock` and `page.updateBlock` to manage state.

This example can be as simple as a todo list (no rich text) with no styling support. For this usage, we can define `ContainerBlock` and `TodoBlock` with the components you are familiar with. Using any popular framework can be acceptable (React, Vue...)

We have `packages/playground/examples` Vite environment that can be used as new example entry. The newly added example can also be added into our documentation for reference.

devlzl:
resolve #2071 

There is a small problem, in steps in #2071, it mentions listening to the `page.rootAdded` event, but I found that `rootAdded` is coupled with `affine:page` flavour, If I completely use only custom schema, I won't be able to listen to the `pageAdded` event. So I only listened to `model.propsUpdated` and `model.childrenUpdated` to update the UI.


himself65 (changes requested):
Please also update 
https://github.com/toeverything/blocksuite/blob/ff9cc9dd731efc2ac600c334b25474c961916c0c/packages/playground/vite.config.ts#L32-L46

lzlme:
> Please also update
> 
> https://github.com/toeverything/blocksuite/blob/ff9cc9dd731efc2ac600c334b25474c961916c0c/packages/playground/vite.config.ts#L32-L46

done

himself65 (commented):
I see; this is a cool example. But it should be a full to-do list app that supports deleting and updating.

Also, we are a CDRT library, so it must include redo/undo features.

For now it's too basic, I think it's not ready for a real show case

lzlme:
> 

Yes, as mentioned in #2071 , this is as simple an example as possible, to demonstrate the use of `@blocksuite/store` (defining custom block flavours, listening to model.propsUpdated and model.childrenUpdated, calling page. addBlock and page.updateBlock). That's why I only used VanillaJS and not a framework like React/Vue.

The reason for not adding the delete feature is that `_handleYBlockDelete` does not dispatch model delete event (I'm trying to fix this).
https://github.com/toeverything/blocksuite/blob/33c2000afbabad8933ec75b4dd42f29489217778/packages/store/src/workspace/page.ts#L743-L751

If necessary, I can change it to a full-featured TodoApp. Do I have to do this ?

himself65:
> 

If doodlewind said that, it's ok

himself65 (approved):

doodlewind (approved):
Looks good! I will update docs to reflect this example.

It will be great if the features like updating items and undo/redo can be added in further PRs.

PR #2254, opened 2023-04-24
refactor(virgo): support forged v-text node

Flrande:
Related to https://github.com/toeverything/blocksuite/pull/2232


lawvs (commented):
What does `forged` mean?

Flrande:
> What does forged mean?

It means that we do not directly use `v-text`, but instead simulate a `v-text` through data-* and usually used in scenarios where the characters in Y.Text do not match the actual rendered characters.

PR #2272, opened 2023-04-25
refactor(store)!: simplify `createPage` API

lawvs:
Rework of https://github.com/toeverything/blocksuite/pull/1959

- new API
```ts
  createPage(
    options: { pageId?: string; init?: true | { title: string } } = {}
  )
```

### DIFF
--- a/packages/store/src/workspace/workspace.ts
+++ b/packages/store/src/workspace/workspace.ts
@@ -212,19 +212,17 @@ export class Workspace {
     });
   }
 
-  createPage(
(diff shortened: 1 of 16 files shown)


packages/store/src/workspace/workspace.ts
```diff
@@ -212,7 +212,19 @@ export class Workspace {
     });
   }
 
-  createPage(pageId: string) {
+  createPage(
+    options: { pageId?: string; init?: true | { title: string } } = {}
```
doodlewind, line 216: `pageId` -> `id`

Issue #1453, opened 2023-03-02
Support selecting multiple blocks by holding down the Shift key


joebeijing:
This allows selecting related blocks in one go, based on the starting block where the cursor is located and the final block where the click occurs.

fundon:
The native behavior is to select the text. Is this text(native) or block.

doodlewind:
@fundon if we are already under block level selection, it will be great if shift + click can select more blocks.

fundon:
* native selection
* block selection

doodlewind:
Closed by #2280

PR #2280, opened 2023-04-25
feat(selection): support shift-click

fundon:
Related to: https://github.com/toeverything/blocksuite/issues/1453
* [x] native: select texts
* [x] block: select blocks
* [x] tests


Flrande:
It looks like we need to select at least one character before we can use shift + click to select multiple lines at once, and if we just have a collapsed selection, shift + click won't work.

https://user-images.githubusercontent.com/50035259/234451486-baef4952-ee09-4e4a-b41f-c9d02286ce7b.mp4

fundon:
@Flrande fixed

doodlewind (approved):

PR #2291, opened 2023-04-26
feat(store): add schema validator for blocks

Saul-Mirone:
Currently the schema validator will run for API:

* addBlock
* updateBlock
* moveBlocks


packages/store/src/__tests__/workspace.unit.spec.ts
```diff
@@ -217,15 +234,21 @@ describe('addBlock', () => {
   it('can add block to root', async () => {
     const page = createTestPage();
 
-    queueMicrotask(() => page.addBlock('affine:page'));
+    let frameId: string;
+
+    queueMicrotask(() => {
+      const pageId = page.addBlock('affine:page', {});
```
doodlewind, line 240: Can omit the `{}` here

Saul-Mirone: Removed

PR #2292, opened 2023-04-26
fix: store user can't define own root flavor

zuozijian3720:
I'm using the @blocksuite/store separately and stuck with this bug.


doodlewind (approved):

Saul-Mirone (approved):

Saul-Mirone:
@doodlewind we need to revert this commit because this will break current database block. For now we have two root blocks: `affine:surface` and `affine:page`. The code will mark `isRoot` as `true` for both of them. The final solution will be done by https://github.com/toeverything/blocksuite/issues/2320.

Issue #2284, opened 2023-04-25
After dragging the database, the number/text column data is lost


zqran:
After dragging the database, the number/text column data is lost:

https://user-images.githubusercontent.com/15389209/234295283-bbe524ba-705d-42fd-b76c-ff87e6ac0688.mov

zqran:
@fundon What happens when dragging and droping the `database` block?

fundon:
@zqran In the video, the children of the model in the database have two blocks.

zqran:
> @zqran In the video, the children of the model in the database have two blocks.

@fundon Thanks! This problem is related to virgo, not caused by dragging.

PR #2319, opened 2023-04-26
fix(database): data loss after dragging database

zqran:
close: #2284


packages/store/src/workspace/page.ts
```diff
@@ -452,8 +452,12 @@ export class Page extends Space<FlatBlockMap> {
         yChildrenB.insert(nextIdx + 1, ids);
       }
     });
-    currentParent.propsUpdated.emit();
-    newParent.propsUpdated.emit();
+
+    currentParent.childrenUpdated.emit();
+    newParent.childrenUpdated.emit();
+    setTimeout(() => {
```
doodlewind: The timeout here doesn't look right

zqran: Yes, this is a temporary operation and has been fixed.

doodlewind (approved):

PR #2347, opened 2023-04-27
refactor(phasor): new element operation APIs

### LINKED ISSUE #2318: Change `updateXXXElementProps` to `updateElementProps`
- [ ] Resolve element bound and other props by element self
- [ ] Utility function `updateXXXElementProps` provided in edgeless-page-block

alt1o:
fix #2318 

Remove the `getProps` & `getBoundProps` functions because they are semantically unclear and difficult to reuse.
Add `applySerializedProps` for applying props to the element.
Add `getUpdatedSerializedProps` for processing props.


doodlewind (commented):
This generally looks good, I'm just thinking about finding some better naming for the update pattern we are using.

doodlewind (approved):
LGTM, some housekeeping on the deprecated APIs may be required.

Issue #2336, opened 2023-04-27
Losing quote style when pasting at the start of empty quote block


doodlewind:
https://user-images.githubusercontent.com/7312949/234749389-56704b6a-b5e7-4c8a-a504-751fd9329980.mov

PR #2350, opened 2023-04-27
fix: keep paragraph block's type when pasting excerpt text type (quote block)

PerfectPan:
Close https://github.com/toeverything/blocksuite/issues/2336


doodlewind (approved):

doodlewind:
Thanks for the fix!

PR #2371, opened 2023-04-29
fix: linked page keyboard shortcut not working on German keyboards

AlessioGr:
On German keyboards, you have to press "Option" + "L" in order to get "@". This PR fixes it, by setting "altKey" (this seems to be the option key on Mac) to null there


lawvs (approved):
Good catch! Thanks!

PR #2383, opened 2023-05-03
feat(database): support column type `progress`

zqran:
Headless version of `progress`.

close: #2384

https://user-images.githubusercontent.com/15389209/236100315-2e73a259-ede5-4650-979c-dee0bc17b234.mov

### DIFF
--- a/packages/blocks/src/database-block/components/column-type/progress.ts
+++ b/packages/blocks/src/database-block/components/column-type/progress.ts
@@ -195,7 +195,7 @@ class ProgressColumnPropertyEditing extends DatabaseCellElement<number> {
 export const ProgressColumnRenderer = defineColumnRenderer(
   'progress',
   () => ({}),
-  () => 0,
+  () => null,
   {
     Cell: ProgressCellEditing,
     CellEditing: false,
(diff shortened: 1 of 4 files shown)


doodlewind (approved):

packages/blocks/src/database-block/components/column-type/progress.ts
```diff
@@ -0,0 +1,207 @@
...
+  static override tag = literal`affine-database-progress-column-property-editing`;
+}
+
+export const ProgressColumnRenderer = defineColumnRenderer(
+  'progress',
+  () => ({}),
+  () => 0,
+  {
+    Cell: ProgressCellEditing,
+    CellEditing: false,
```
doodlewind, line 201: I think `null` here makes more sense

zqran: Currently, all column types depend on this field.

I will open a new pr to modify it, and optimize some redundant fields in the original design.

doodlewind: That would be great!

Tips: you can reference this issue comment in your refactoring PR :)

PR #2391, opened 2023-05-04
fix: support dragging block to blank

rottenpen:
fixed: https://github.com/toeverything/AFFiNE/issues/2125

Although my code runs without errors, I don't believe it follows best practices. I would appreciate any suggestions you may have for improvement.


doodlewind (approved):

zqran:
@rottenpen  When press ArrowUp there is an error. See the reproduce video:

https://user-images.githubusercontent.com/15389209/236096129-cd55c4ad-38c7-40bd-810e-7746f531ac9b.mov

packages/editor/src/utils/editor.ts
```diff
@@ -81,7 +84,7 @@ export const createBlockHub: (
       }
 
       // In edgeless mode.
-      const pageBlock = getEdgelessPage(page);
+      const pageBlock = getEdgelessPage(page) || getDefaultPage(page);
```
doodlewind, line 87: @lawvs is this still deprecated currently?

lawvs, line 87: Thank you for your contribution, but in this position, we should only handle the logic of Edgeless. I will make adjustments to your pull request to ensure it fits with the existing logic.

lawvs (changes requested):

packages/editor/src/utils/editor.ts
rottenpen: Yes, I think so. However, I have no idea what "Edgeless" refers to. So I can't distinguish when it is Edgeless.

rottenpen:
> @rottenpen When press ArrowUp there is an error. See the reproduce video:
> 
>  2023-05-04.10.10.39.mov

fixed

lawvs:
Closing your PR in favor of https://github.com/toeverything/blocksuite/pull/2469, but thank you for your contribution and effort.

PR #2398, opened 2023-05-04
feat(store): add custom formatter for BaseBlockModel

lawvs:
https://user-images.githubusercontent.com/18554747/236152767-80dcf544-5957-410c-98f2-68ab2266867f.mov


doodlewind:
Just saying that we don't have to add label for PRs.

doodlewind (commented):
image

I'm unable to see the formatter with the `$0.model` expression.

lawvs:
> I'm unable to see the formatter with the `$0.model` expression.

Now it is only enabled in a development environment, so it does not work on the preview site.

Screenshot 2023-05-05 at 3 32 01 PM

```ts
  if (
    !(process.env.NODE_ENV === 'development') ||
    typeof window === 'undefined'
  ) {
    return;
  }
```

PR #2400, opened 2023-05-04
feat(database): separate files to support kanban view

zqran:
Main directory structure:

```html
/database-block
  /kanban
    kanban-view.ts
    ...
  /table
    /components
    table-view.ts
    ...
  database-block.ts
  database-model.ts
  ...
```

part of: #2401


packages/blocks/src/database-block/kanban/kanban-block.ts
```diff
@@ -0,0 +1,59 @@
+// related component
```
doodlewind, line 1: Should be `kanban-view.ts` and `table-view.ts`

doodlewind (approved):

doodlewind:
Any link for quick preview?

doodlewind:
Got it, it's still under headless state, which is fine. Just remember to add feature flag for the slash menu!

zqran:
> Got it, it's still under headless state, which is fine. Just remember to add feature flag for the slash menu!

Get!

PR #2404, opened 2023-05-04
style: fixed z-index for toolbar 

karti16:
Toolbar was under the side bar. Changed z-index of the toolbar to the modal  [Issue link](https://github.com/toeverything/AFFiNE/issues/2205)

Before : 
before

After : 
after


zqran:
> I think this is not a good approach, this file is exported from design side I think.

As https://github.com/toeverything/AFFiNE/pull/2215#pullrequestreview-1411649259 mentioned, these CSS variables are defaults by design.

/cc @hwangdev97

Himself65 (approved):

karti16:
@Himself65 do i have to look into the test that has been failed?

PR #2462, opened 2023-05-09
chore: add shadow for block hub when editor mode is edgeless

JimmFly:


packages/blocks/src/components/block-hub.ts
```diff
@@ -855,7 +861,11 @@ export class BlockHub extends WithDisposable(ShadowlessElement) {
       >
         ${blockHubMenu}
         <div
-          class="has-tool-tip new-icon ${this._expanded ? 'icon-expanded' : ''}"
+          class="has-tool-tip new-icon ${this._inEdgelessMode
+            ? this._expanded
+              ? 'icon-expanded'
+              : 'new-icon-in-edgeless'
+            : ''}"
```
lawvs, line 868: To simplify complex CSS classes, you should consider using the `classMap` function from Lit.

Issue #2447, opened 2023-05-08
Cant select text when clicking on blank area in edgeless


fundon:
https://user-images.githubusercontent.com/27926/236778893-b4eff20e-8648-4f3a-b466-22bc156d98f5.mov

ShortCipher5:
Not very consistently reproduced, but still seems to be an issues in the latest version (v0.5.x-staging-36871e1d9667667718c6bee7891d507142cc5f19)

![Select](https://github.com/toeverything/blocksuite/assets/4605025/4d25299e-fe4b-465c-b406-4ff614dc4470)

joebeijing:
Can not reproduce it in the latest beta.

PR #2474, opened 2023-05-09
fix: select texts when clicking on blank area

fundon:
Fixes: https://github.com/toeverything/blocksuite/issues/2447


Flrande:
I was unable to reproduce the problem in https://github.com/toeverything/blocksuite/issues/2447 and need some time to confirm whether this fix is correct.

packages/virgo/src/services/event.ts
```diff
@@ -158,7 +158,10 @@ export class VirgoEventService<TextAttributes extends BaseTextAttributes> {
           node => node instanceof HTMLElement
         ).length === 1
       ) {
-        this._editor.focusEnd();
+        this._editor.rangeService.setVRange({
+          index: this._editor.yText.length,
+          length: range.endOffset,
+        });
```
fundon, line 164: @Flrande  When backward selection, the cursor will jump, start <-> end, very unstable, so changed to the current

Flrande: What is the value of native range in this case? It seems that this change will make vRange overflow because the index has been reached max value (this._editor.yText.length).

fundon: Screenshot 2023-05-10 at 9 12 09 AM

Flrande (approved):

PR #2485, opened 2023-05-10
feat: restore css debug menu

thorseraq:


Flrande (approved):

packages/editor/themes/affine.css
```diff
@@ -10,3 +10,7 @@
 :root {
   --affine-editor-width: 720px;
 }
+
+.dg > ul {
+  overflow: scroll;
+}
```
lawvs, line 16: What is the function of this style?

thorseraq: It’s css selector of dat.gui menu , to make container scrollable, rather than hide overflowed content by default

Issue #2441, opened 2023-05-08
Support adding text above the shape


doodlewind:
In whiteboard, when shape overlaps with current text (frame actually), the shape will always be put on top of the text. This is by design since we put all the canvas content inside one top `SurfaceBlock` layer, but makes user confused about our text feature. Since **it's not supported to put a piece on top of the shape**. It was understood as a bug (from Azure).

![image](https://user-images.githubusercontent.com/7312949/236724991-4619bd8b-7a17-4d37-b652-e4e7f2cf06b5.png)

/cc @joebeijing for related product design (UX to differentiate adding text block and adding canvas handwriting style text element)

joebeijing:
Please refer to [Internal link](https://toeverything.feishu.cn/wiki/wikcnP7za5C8FCOvC1msgPqncQe#MMkoduY0IoY4aSxsFRVcNFw6nRc) for Product details，and this part also requires work from the design team, mainly because new component menu options need to be added. CC @Svaney-ssman

PR #2486, opened 2023-05-10
refactor(phasor): make yjs as the single of truth

Flrande:
Preceding steps for #2441

An element in phasor corresponds to a `SurfaceElement` and a `YElement`, the renderer is essentially controlled by the `SurfaceElement`, and `surface.ts` implements a two-way binding between the `SurfaceElement` and the `YElement`. This creates a lot of code for mutual transformation and synchronization between them which is confusing for readers and difficult for maintenance, and also prone to collaborative bugs, because yjs is not the single of truth in essence.

For this reason, the following refactoring was done, removing `YElement` from the original code and adding `Y.Map` field to the `SurfaceElement`.


doodlewind (approved):

doodlewind:
Great progress, hope to see more tractions in Phasor!

Issue #2483, opened 2023-05-10
Unable to adjust layer z-index on edgeless frames


doodlewind:
"Bring to front" and "Send to back" currently only works for phasor elements

<img width="729" alt="image" src="https://github.com/toeverything/blocksuite/assets/7312949/75ce5e73-fef2-4868-964d-d62bf75ce30f">

doodlewind:
Closed by #2527

PR #2527, opened 2023-05-11
feat(edgeless): support bring forward and send backward elements

### LINKED ISSUE #2439: Unable to directly bring item to top or bottom in whiteboard
Currently, user has to adjust layer index by one step each time.

(from Azure)

fundon:
Close #2439, #2483

* [x] shapes
* [x] frames
* [x] shapes + frames
* [x] ~~opt codes~~
* [x] tests


packages/blocks/src/frame-block/frame-model.ts
```diff
@@ -7,6 +7,7 @@ export const FrameBlockSchema = defineBlockSchema({
   props: () => ({
     xywh: '[0,0,720,480]',
     background: FRAME_BACKGROUND_COLORS[0],
+    zIndex: 0,
```
doodlewind, line 10: We need to use `string` type for this, so that we can use the `generateKeyBetween` util to handle [fractional indexing](https://github.com/rocicorp/fractional-indexing). This can help making less index updates on switching layer sequence, see [this figma blog post](https://www.figma.com/blog/realtime-editing-of-ordered-sequences/#fractional-indexing)

doodlewind (approved):

doodlewind:
Great work, thanks!

PR #2530, opened 2023-05-11
fix: wrong format when paste a specify text

### LINKED ISSUE #2173: Wrong format when paste a specify text
### What happened?

When I try to paste the text. the format is worry
```
The number of pictures within paper mode is not always corresponded with the number of picture within edgeless mode toeverything/AFFiNE#1639
```
but the text is not reproduced. Need to copy text from the browser again.  for exmaple https://github.com/toeverything/AFFiNE/issues/1639

video:
local:
CleanShot 2023-04-19 at 17 21 10

app.affine.pro
CleanShot 2023-04-19 at 17 21 39

screenshot:
CleanShot 2023-04-19 at 17 28 53@2x

I think the bug was caused by didn't correctly handling the format of copy text. 
I'd like to fix the bug by submitting a PR!  Although it might be too difficult to find this one at the end. But I want to try.👀

### Version

app.affine.pro

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

_No response_

### Anything else?

_No response_

zuoxiaodong0815:
close #2173


doodlewind (approved):
Good catch!

PR #2546, opened 2023-05-11
feat(phasor): support `TextElement`

Flrande:
https://github.com/toeverything/blocksuite/assets/50035259/a3944ba6-2fb7-4950-abaf-9ece613ec923


tests/edgeless/frame.spec.ts
```diff
@@ -122,28 +122,28 @@ test('add Text', async ({ page }) => {
   await assertEdgelessSelectedRect(page, [0, 0, 448, 80]);
 });
 
-test('add empty Text', async ({ page }) => {
+test('add empty Note', async ({ page }) => {
```
doodlewind, line 125: I think we'll make the migration to `affine:note` in later PRs?

packages/phasor/src/surface.ts
```diff
@@ -202,7 +202,7 @@ export class SurfaceManager {
 
   updateElement<T extends keyof IPhasorElementType>(
```
doodlewind, line 203: Can we always call `doc.transact` in `updateElement`?

fundon: https://github.com/toeverything/blocksuite/pull/2677#issuecomment-1557612615

doodlewind (approved):

doodlewind (approved):

PR #2551, opened 2023-05-12
fix: selection status with redo/undo

### LINKED ISSUE #1250: Edgeless selection rect doesn’t follow undo/redo
This is similar with #1249 but just triggered by performing undo/redo.

https://user-images.githubusercontent.com/7312949/225053721-d0741d49-bcc8-4088-b9a7-98c07b59f760.mov

donglixiaoche:
close #1250 

add selection state to history stack meta info, for recreate selection rect when redo/undo


donglixiaoche:
https://github.com/toeverything/blocksuite/assets/19243114/59347277-f424-4f01-8594-cc3ecbc41843

@alt1o for undo/redo, why do we need to keep current cursor active after applying undo? i think it also makes sense if redo/undo change cursor status, what do you think?

PR #2554, opened 2023-05-12
refactor(store): change editor structure to single root block

Saul-Mirone:
This PR moves surface block into children of page block and keeps the editor has single root.


doodlewind (commented):
This would require migration (e.g., trying to export YDoc from https://blocksuite-toeverything.vercel.app/?init and import locally).

packages/store/src/workspace/migrations.ts
```diff
@@ -174,6 +174,56 @@ const migrations: Migration[] = [
       }
     },
   },
+  {
```
Saul-Mirone, line 177: @doodlewind Does this migration look good to you?

Issue #2433, opened 2023-05-08
Unable to input under block-level selection


doodlewind:
Currently pressing ctrl-A will go into block-level selection for single block, but user can't type in new character in this state, which is a major blocker. Input under range selection was supported (manually), so this one should be a low-hanging fruit.

(from Azure)

/cc @QiShaoXuan

xell:
Please refer to #2514 .

xell:
I reopen this issue because #2558 solution can not have a certain effect which can be achieved by "Input under range selection".

When I type / (slash) in a range selection, the slash menu will successfully pop up as following:

https://github.com/toeverything/blocksuite/assets/132558/ed050562-c586-40f0-b7c1-b7969a503660



However, when typing / in a block selection (pressing Ctrl/Cmd + A one or two times), the / is literally input and there's no slash menu.

doodlewind:
@xell I think this could be another issue.

xell:
> @xell I think this could be another issue.

Ok, please see https://github.com/toeverything/blocksuite/issues/2611 .

PR #2558, opened 2023-05-12
feat: support input under block-level selection

QiShaoXuan:
Close https://github.com/toeverything/blocksuite/issues/2433


lawvs (commented):
Could you add some tests for it?

PR #2565, opened 2023-05-14
feat(phasor): simulate pressure in brush

Flrande:
before:

https://github.com/toeverything/blocksuite/assets/50035259/c213b5a4-48e3-4e5d-987a-f670bfe627a8

after:

https://github.com/toeverything/blocksuite/assets/50035259/d494c8e1-38d7-4a2a-bdb9-ab471f5d8606


doodlewind (approved):
It feels great to see this happen!

doodlewind:
I think this also need to be applied in shape elements

Flrande:
> I think this also need to be applied in shape elements

I am working on it.

PR #2596, opened 2023-05-16
fix(edgeless): zIndex on tooltip

### LINKED ISSUE #2586: Edgeless toolbar tooltip is incorrectly masked
Perhaps the z-index of the tooltip should be corrected.

Image

fundon:
Close #2586


doodlewind (approved):
Please use `close #2586` next time :), Since `related to #2586` won't perform auto close

PR #2626, opened 2023-05-17
fix(edgeless): connector's start/end handler not consider zoom

blurname:
before

https://github.com/toeverything/blocksuite/assets/50226652/cd889b83-c381-4cd6-b0d7-202c383ca655

after

https://github.com/toeverything/blocksuite/assets/50226652/659feda8-4626-495c-8998-51f318a2a624


doodlewind (approved):
Thanks!

PR #2655, opened 2023-05-18
fix(block-hub): fix the block-hub toggle

### LINKED ISSUE #2644: If blockHub is open, clicking the blockHub icon in the row does not respond
If blockHub is open, clicking the blockHub icon in the row does not respond
I not sure this rule is a feature or bug, but I feel confused about this rule.
I think we can toggle blockHub when we clicking the blockHub icon in the row, not just open.
If this proposal is accepted, I can provide a PR.
image

imyuanx:
Fixed #2644 

https://github.com/toeverything/blocksuite/assets/50652358/9cafa6f0-61fe-442d-9b60-bf4227568c03

### DIFF
--- a/packages/blocks/src/components/block-hub.ts
+++ b/packages/blocks/src/components/block-hub.ts
@@ -673,13 +673,9 @@ export class BlockHub extends WithDisposable(ShadowlessElement) {
     this._onClickCardCallback(data);
   };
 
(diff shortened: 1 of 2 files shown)


doodlewind (approved):
Thanks for the improvement!

Issue #2641, opened 2023-05-18
When a single line of text exceeds the visible range in Code Blocks under Windows, the horizontal scrollbar is missing


joebeijing:
<img width="681" alt="CleanShot 2023-05-18 at 3 36 40@2x" src="https://github.com/toeverything/blocksuite/assets/22443345/fd7049f5-2b66-4207-9763-5ab6adeb188e">

joebeijing:
On Windows, it is necessary to ensure that there is a horizontal scrollbar that appears when hovering the mouse, so that scrolling horizontally is possible.

hritikchaudhary:
I just setup blocksuite on my locall.. looking into it can you please assign it to me..

hritikchaudhary:
One question? Do we want horizontal scrollbar to appear when user hover and try to scroll or just by hovering on rich-text it should be visible at all time.. cause i analysed other components and scrollbar is only visible when we try to scroll.. so I'm moving forward with this assumption only to maintain consistency in behaviour.
Let me know if you want it visible at all time when cursor is on the rich-text class.

PR #2659, opened 2023-05-18
fix: scrollbar not visible for codeblocks issue

hritikchaudhary:
close: #2641 

https://github.com/toeverything/blocksuite/assets/33057454/33feae63-5cd5-471a-8fe2-96e217b1f71e


doodlewind (approved):
Thanks for the fix!

PR #2675, opened 2023-05-19
fix(edgeless): color of icon on toolbar when it is activated

### LINKED ISSUE #2650: The Edgeless toolbar buttons lack an active state

fundon:
Close #2650


doodlewind (approved):
Good catch!

PR #2692, opened 2023-05-22
feat(playground): add `heavy-whiteboard` preset example

fundon:
Screenshot 2023-05-22 at 3 26 37 PM


packages/playground/src/data/heavy-whiteboard.ts
```diff
@@ -0,0 +1,72 @@
+import {
+  DEFAULT_SHAPE_FILL_COLOR,
+  DEFAULT_SHAPE_STROKE_COLOR,
+} from '@blocksuite/blocks';
+import { serializeXYWH } from '@blocksuite/phasor';
+import { nanoid, Text, type Workspace } from '@blocksuite/store';
+
+import { addShapeElement, type InitFn } from './utils';
+
+const COUNT = 500;
```
doodlewind, line 10: How about making this configurable using URL param like `?init=heavy-whiteboard&count=500`, with `100` as the default value?

doodlewind (approved):

Issue #2576, opened 2023-05-15
After selecting an individual block by clicking the drag handle, the balloon toolbar did not appear


joebeijing:
https://github.com/toeverything/blocksuite/assets/22443345/3eaf53fb-a6d6-4b3e-9f4a-cade77ac9459

lawvs:
Related to https://github.com/toeverything/blocksuite/issues/1677

PR #2694, opened 2023-05-22
feat: show format-bar when click drag handler

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/2576


doodlewind:
Seems that there are some remaining test cases to be passed, any blocker here @lawvs?

doodlewind:
@lawvs any update on this one? I believe the issue is also expected to be solved in 0.7

doodlewind (approved):
Good to see this finally lands!

PR #2699, opened 2023-05-22
fix(edgeless): switch to default mode after using shape

### LINKED ISSUE #1341: Fix `getDefaultPageBlock` typing issue and support edgeless mode
The `getDefaultPageBlock` function has an unsafe type. It returns `DefaultPageBlockComponent | EdgelessPageBlockComponent | null`, but is typed as `DefaultPageBlockComponent`.

This will break edgeless mode, so we need to design a new API to get `DefaultPageBlockComponent` and distinguish between page and edgeless mode.

https://github.com/toeverything/blocksuite/blob/dd801b5d28d82ba3f5786892e5b4538a903cd9c8/packages/blocks/src/__internal__/utils/query.ts#L132-L144

### LINKED ISSUE #2652: Optimize the switching logic of Edgeless toolbar functions after use
It is recommended to follow the common processing logic (reference to Miro, FigJam, and Tldraw) where after using Shape, Text, Connector, and Note functions, the state should automatically switch to Select. Only Pen can be used continuously.

fundon:
Closes #2652
Closes #1341

* [x] tests


packages/blocks/src/page-block/edgeless/toolbar/edgeless-toolbar.ts
```diff
@@ -210,10 +206,41 @@ export class EdgelessToolbar extends LitElement {
...
+        Object.assign(state, this.edgeless.selection.lastState);
+        this.edgeless.selection.lastState = null;
+      }
+    } else if (this.edgeless.selection.state.selected.length) {
+      this.edgeless.selection.lastState = this.edgeless.selection.state;
+    }
+
+    this.edgeless.slots.mouseModeUpdated.emit(mouseMode);
+    this.edgeless.slots.selectionUpdated.emit(state);
+  };
```
lawvs, line 242: It would be better if you designed an API in the edgeless selection manager instead of operating the state directly.

packages/blocks/src/__internal__/utils/types.ts
```diff
@@ -241,3 +248,24 @@ export type Detail<T extends keyof WindowEventMap | keyof HTMLElementEventMap> =
     : T extends keyof HTMLElementEventMap
     ? HTMLElementEventDetail<T>
     : never;
+
+export abstract class AbstractSelectionManager<
```
doodlewind, line 252: @Saul-Mirone how do you think about adding this abstraction? Seems the having the DOM `container` for host makes sense to us.

doodlewind (approved):

packages/blocks/src/__internal__/utils/query.ts
```diff
@@ -152,36 +152,20 @@ export function getPreviousBlock(
...
+  return document.querySelector(`[${ATTR}="${model.page.root.id}"]`);
 }
 
 /**
  * If it's not in the page mode, it will return `null` directly.
  */
 export function getDefaultPage(page: Page) {
-  if (!isPageMode(page)) {
-    return null;
-  }
```
lawvs, line 184: What does this operation do? Is it safe?

lawvs (changes requested):

packages/blocks/src/__internal__/utils/query.ts
```diff
@@ -152,37 +152,22 @@ export function getPreviousBlock(
...
   const editor = getEditorContainer(page);
+  if (editor.mode !== 'page') return null;
```
fundon, line 170: @lawvs It should be safe, here also added a check.

doodlewind:
@fundon I think this is good to land once conflicts are resolved, thanks!

fundon:
> @fundon I think this is good to land once conflicts are resolved, thanks!

Solved

packages/blocks/src/__internal__/utils/types.ts
Saul-Mirone: I think the selection manager won't be a long term pattern. I'll redesign it when working on the plugin spec.

PR #2700, opened 2023-05-22
perf(edgeless): use page-level layer to replace block-level transform

fundon:
### before

https://github.com/toeverything/blocksuite/assets/27926/bfddc8ca-04d5-41eb-9002-a6379607ed2f

### after

https://github.com/toeverything/blocksuite/assets/27926/c555f22d-2dfd-4f3a-91a8-831db5d5bce3


doodlewind (approved):
This makes a lot of sense!

Issue #2579, opened 2023-05-15
When the search box is expanded in the database, the interface experiences is shaking


joebeijing:
https://github.com/toeverything/blocksuite/assets/22443345/c2005d86-4ea3-4b7f-a69a-873176342dd9

The New Record button has shifted slightly to the right.

hwangdev97:
I think part of the reason here is due to the inadequate width of the editing area, but specific reasons still need to be verified and checked with @zqran

zqran:
> I think part of the reason here is due to the inadequate width of the editing area, but specific reasons still need to be verified and checked with @zqran

Yes, that's why.

PR #2718, opened 2023-05-24
feat: update editor width

zqran:
related to: https://github.com/toeverything/AFFiNE/issues/2486

close: https://github.com/toeverything/blocksuite/issues/2579


zqran:
In BlockSuite, the value of `720px` is also used in files such as `file-exporter` / `frame block` / `debug-menu` and `E2E tests`.

Do these files also need to be modified to `800px`? /cc @hwangdev97

hwangdev97:
I think so cc @doodlewind

doodlewind:
LGTM

doodlewind (approved):

PR #2723, opened 2023-05-24
fix(store): support moving blocks in multiple frames

fundon:
Try [heavy-whiteboard](https://blocksuite-toeverything.vercel.app/?init=heavy-whiteboard) example in page mode.

### DIFF
--- a/packages/store/src/workspace/page.ts
+++ b/packages/store/src/workspace/page.ts
@@ -377,159 +377,133 @@ export class Page extends Space<FlatBlockMap> {
     return id;
   }
 
-  /**
-   * Moves contiguous blocks to the new parent.
-   *
-   * @example
-   * Moves blocks in a frame only.
-   * Moves `2` to after `0`.
-   * ```
-   * frame A
-   * +--------------+
-   * |  block       |
-   * |  +--------+  |
-   * |  | 0      |  |
-   * |  +--------+  |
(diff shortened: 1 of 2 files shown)


lawvs (dismissed):
Could you add some comments to help us understand it better?

packages/store/src/workspace/page.ts
```diff
@@ -389,44 +389,65 @@ export class Page extends Space<FlatBlockMap> {
...
-      }
+      // after the target
+      let offset = 1;
 
-      const ids = blocks.map(block => block.id);
-      if (insertBeforeSibling) {
-        yChildrenB.insert(nextIdx, ids);
-      } else {
-        yChildrenB.insert(nextIdx + 1, ids);
+      for (const [parent, blocks] of tempMap) {
```
doodlewind, line 422: It seems that each array value of the `tempMap` here only has one item (line 411), when do we need to loop through the `blocks` here?

fundon:
> Could you add some comments to help us understand it better?
Added

packages/store/src/workspace/page.ts
fundon: This is the case in the `heavy-whiteboard` example, but we can edit the frames.

doodlewind (approved):
I refactored the naming a bit, thanks for this feat!

PR #2725, opened 2023-05-24
refactor(database): use `dispatcher` to implement row-level selection

zqran:


Saul-Mirone (approved):

packages/blocks/src/database-block/table/selection-manager/utils.ts
```diff
@@ -57,6 +57,11 @@ export function getClosestDatabaseId(element: Element) {
   return databaseId;
 }
 
+export function isInDatabase(element: Element) {
+  const database = element.closest('affine-database');
+  return database !== null;
+}
+
```
lawvs, line 64: Is it duplicated of `isInsideBlockByFlavour`?

zqran: Their signatures are different, `isInDatabase` is used to determine whether the current element is in the Database.

PR #2754, opened 2023-05-25
refactor(database): use `dispatcher` to implement cell-level selection

zqran:


packages/lit/src/event/keyboard.ts
```diff
@@ -0,0 +1,22 @@
+import { UIEventStateContext } from './base.js';
+import type { UIEventDispatcher } from './dispatcher.js';
+import { KeyboardEventState } from './state.js';
+
+export class KeyboardControl {
+  constructor(private _dispatcher: UIEventDispatcher) {}
+
+  listen() {
+    this._dispatcher.disposables.addFromEvent(document, 'keydown', this._down);
```
Saul-Mirone, line 9: maybe also add keyUp event?

Saul-Mirone (approved):

Issue #2752, opened 2023-05-25
Database row-level selection offset in edgeless mode


doodlewind:
<img width="968" alt="image" src="https://github.com/toeverything/blocksuite/assets/7312949/88f90c36-6ceb-489b-b937-bb0bdbfe1892">

PR #2758, opened 2023-05-25
feat(database): support selections scaling in edgeless mode

zqran:
close: #2752


packages/blocks/src/database-block/table/components/selection/cell-selection.ts
```diff
@@ -39,6 +39,12 @@ export class CellLevelSelection extends WithDisposable(LitElement) {
     this.state = null;
   };
 
+  private get _zoom() {
```
doodlewind, line 42: Before this PR, no `zoom` or `edgeless` keyword exists in the database block. Is there some way that the block UI can simply fit in without special adaptation to the whiteboard state?

doodlewind (approved):

Issue #2757, opened 2023-05-25
Edgeless breaks when deleting all content


ShortCipher5:
- Select all blocks and delete
- The page breaks 

- By going to the title and hitting `Enter` we can fix Page but Edgeless stays broken

![BreakingEdgeless](https://github.com/toeverything/AFFiNE/assets/4605025/da3ad982-8111-449f-81f4-b2e10a848207)

PR #2761, opened 2023-05-25
fix: batch selection contains surface element

Saul-Mirone:
Closes: https://github.com/toeverything/blocksuite/issues/2757


doodlewind (approved):
Good catch!

doodlewind (approved):
Good catch!

PR #2770, opened 2023-05-26
fix(database): page scrolling when cell selection is selected

zqran:
before:

https://github.com/toeverything/blocksuite/assets/15389209/c1f38468-c6d8-4925-8fee-d704ef471068

after:

https://github.com/toeverything/blocksuite/assets/15389209/58603d0f-9a03-4d1b-939d-af26dbd1d16e


doodlewind (approved):
This fix is essential!

PR #2780, opened 2023-05-26
fix(edgeless): keep text size when zooming

### LINKED ISSUE #2748: Text size doesn't count in zoom
When adding text in different zoom, it's always using the same absolute size.

https://github.com/toeverything/blocksuite/assets/7312949/c4989876-e57a-4577-bb37-a734886ab7a9

Flrande:
Close #2748


doodlewind (approved):
Thanks for finally got this fixed!

PR #2808, opened 2023-05-28
feat(phasor): add roughness property

Flrande:


packages/blocks/src/surface-block/surface-model.ts
```diff
@@ -4,7 +4,7 @@ import { literal } from 'lit/static-html.js';
 export const SurfaceBlockSchema = defineBlockSchema({
   flavour: 'affine:surface',
   metadata: {
-    version: 3,
+    version: 4,
```
doodlewind, line 7: Is it possible that we make this non-breaking and just set the `roughness` field a default value if it not exists?

Flrande: Do you mean to make this field optional?

doodlewind: I think it can still be a required field, but maybe we can assign a default value to it, just in case it's missing (in the case we load content from old binaries) 🤔

Flrande: I can add a default value during rendering, but I believe this should only be used as a fallback strategy for migration failures.

doodlewind (approved):

PR #2810, opened 2023-05-28
feat(blocks): add kbd text style

pegak:
This is my first PR. I am trying to learn BlockSuite to contribute more and build on top of it.

It is not perfect in any way. It is almost necessary to alter the code. The decision to be made:

- [ ] How to differentiate between `<code>` and `<kbd>`? Currently, it is "just to be visible".
- [ ] Should we reuse `inlineCodeStyle` or duplicate it? Depends on the style.
- [ ] What keyboard shortcut to use? I picked something, but the decision has to be made.
- [ ] Which icon to use? I just used "something", but it definitely doesn't match the style.

### DIFF
--- a/packages/blocks/src/__internal__/content-parser/file-exporter/exporter-style.ts
+++ b/packages/blocks/src/__internal__/content-parser/file-exporter/exporter-style.ts
@@ -54,6 +54,7 @@ export const globalCSS = css`
     --affine-background-secondary-color: #fbfafc;
     --affine-background-tertiary-color: #e9e9ec;
     --affine-background-code-block: #fafbfd;
+    --affine-background-kbd-block: #fdfbfa;
     --affine-background-modal-color: rgba(0, 0, 0, 0.6);
     --affine-text-primary-color: #424149;
     --affine-text-secondary-color: #8e8d91;
--- a/packages/blocks/src/__internal__/rich-text/virgo/affine-text.ts
+++ b/packages/blocks/src/__internal__/rich-text/virgo/affine-text.ts
@@ -19,17 +19,24 @@ export function affineTextStyles(
   }
 
   let inlineCodeStyle = {};
-  if (props.code) {
+  if (props.code || props.kbd) {
+    const isCode = props.code && !props.kbd;
     inlineCodeStyle = {
       'font-family': 'var(--affine-font-code-family)',
-      background: 'var(--affine-background-code-block)',
+      background: `var(${
+        isCode
+          ? '--affine-background-code-block'
+          : '--affine-background-kbd-block'
+      })`,
       color: 'var(--affine-text-primary-color)',
       'border-radius': '4px',
       padding: '3px 6px',
-      'font-size': 'calc(var(--affine-font-base) - 4px)',
+      'font-size': `calc(var(--affine-font-base) - ${isCode ? '4px' : '2px'})`,
       'font-variant-ligatures': 'none',
       'line-height': 'var(--affine-font-base)',
-      border: '1px solid var(--affine-border-color)',
+      border: `1px solid var(${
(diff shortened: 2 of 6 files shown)


doodlewind:
Thanks for opening PR for this! It can be a great starting point for more text style in AFFiNE.

However it worth clarifying that to ship a new feature like `kbd` style, there are more efforts (UI styles, bug fixes, review...) to be made. So before we push this forward, making sure this is in our scope in our recent iteration may make more sense. Could you elaborate some use case of this feature, and if its functionality can be reached by using the inline code? Thanks!

/cc @joebeijing @JackiePengp

pegak:
Hi,
of course. Use-case I see as really important is if you have a shared knowledge base for a company or some courses, it is important to cover even basic parts like keyboard shortcuts. This is a first-level implementation just have it clearly visible + semantically correct (hence the `kbd` element).
There are definitely things that can improve it to a whole new level - on hover show the keyboard and where to find it (based on the user system) so that it is absolutely clear. But I wanted something simple for a start. :)

joebeijing:
Thank you very much for your PR. From a product perspective, we may temporarily not add this style in the short term. This is mainly because inline code can achieve a similar effect to some extent, and adding this feature will bring more user cognitive and compatibility costs. To avoid distractions from user experience, we will add this feature to the AFFiNE backlog and review it in the future.

pegak:
I understand this. Thank you for the reply! As I said - it was also training for me, to get to know the codebase. I counted on the possibility that it might not be used. :)

I have a question based on this feature request from me - will it be possible to have an "inline-block" with its own logic (as I described - showing the keyboard where the keys are or something) or is it possible only with a full block for now? :)

Flrande:
> I have a question based on this feature request from me - will it be possible to have an "inline-block" with its own logic (as I described - showing the keyboard where the keys are or something) or is it possible only with a full block for now? :)

Hello! This is technically fully achievable. Currently, blocksuite does not provide a way to customize inline elements, but in reality, this functionality relies mainly on our editor core, Virgo. Blocks in blocksuite can smoothly utilize Lit with the help of Virgo to construct a custom inline element.

Regarding the usage of Virgo, there is currently a lack of detailed documentation. However, I can provide you with some simple usage steps:

1. Follow the specifications of Virgo to define [an element with custom logic](https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/__internal__/rich-text/link-node/affine-link.ts). The [default text implementation](https://github.com/toeverything/blocksuite/blob/master/packages/virgo/src/components/virgo-text.ts) in Virgo may be clearer. (It is used in [default render function](https://github.com/toeverything/blocksuite/blob/master/packages/virgo/src/utils/attribute-renderer.ts#L42).)
2. Write a rendering [function that renders the specified element based on the delta](https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/__internal__/rich-text/virgo/attribute-renderer.ts).
3. Define a [schema](https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/__internal__/rich-text/virgo/types.ts#L10) that includes the fields required for rendering this element.
4. When initializing Virgo, [pass the schema and the rendering function](https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/__internal__/rich-text/rich-text.ts#L140-L150).

doodlewind:
Since we don't have plan on built-in keyboard style support (and more breaking change in hotkey is also incoming), this is better to be closed FTM. But the good news is that with ongoing https://github.com/toeverything/blocksuite/issues/3165, we are working on providing APIs to make this kind of customization possible, without modifying source code here. As a draft, the paragraph block should also support `TextRenderer`, so that you don't have to opening PR for adding inline style support. We'll let you know for further progress, thanks again for your interest and support!

Issue #2805, opened 2023-05-28
Support edge cases of cursor keeping


doodlewind:
For example, when cursor is at the start. of the first paragraph block, pressing  `ArrowLeft` should locate the cursor to the page title.

tillkwl:
If this is still due Id like to work on fixing this

doodlewind:
@tillkwl surely welcome!

tillkwl:
#2973

doodlewind:
@Saul-Mirone seems here are regressions about pressing arrow keys in page block title.

doodlewind:
Closing since we've supported `ArrowDown` from the page title.

PR #2812, opened 2023-05-28
fix(blocks): in title line, press key arrowdown will create a new line

### LINKED ISSUE #2807: In title line, press key DownArrow will create a new line
https://github.com/toeverything/blocksuite/assets/41265413/a7da4fe3-c4db-44b8-937a-d0849c97b0d5

donteatfriedrice:
To close #2807 

https://github.com/toeverything/blocksuite/assets/99816898/f5fc6a92-7961-4848-8876-89e45e5757ca

It seems that the first paragraph block is the second child of the page block. So I check the second element of the model's children.

### DIFF
--- a/packages/blocks/src/page-block/default/default-page-block.ts
+++ b/packages/blocks/src/page-block/default/default-page-block.ts
@@ -248,7 +248,7 @@ export class DefaultPageBlockComponent
       return;
     } else if (e.key === 'ArrowDown' && hasContent) {
       e.preventDefault();
-      const firstParagraph = model.children[0].children[0];
+      const firstParagraph = model.children[1].children[0];
       if (firstParagraph) {
         asyncFocusRichText(page, firstParagraph.id);
       } else {
--- a/packages/blocks/src/page-block/default/default-page-block.ts
+++ b/packages/blocks/src/page-block/default/default-page-block.ts
@@ -253,7 +253,9 @@ export class DefaultPageBlockComponent
       return;
     } else if (e.key === 'ArrowDown' && hasContent) {
       e.preventDefault();
-      const firstParagraph = defaultFrame?.children.find(block => block.flavour === 'affine:paragraph');
+      const firstParagraph = defaultFrame?.children.find(
+        block => block.flavour === 'affine:paragraph'
+      );
       if (
         firstParagraph &&
         matchFlavours(firstParagraph, ['affine:paragraph'])
--- a/packages/blocks/src/page-block/default/default-page-block.ts
+++ b/packages/blocks/src/page-block/default/default-page-block.ts
@@ -253,7 +253,7 @@ export class DefaultPageBlockComponent
       return;
     } else if (e.key === 'ArrowDown' && hasContent) {
       e.preventDefault();
-      const firstParagraph = model.children[1].children[0];
+      const firstParagraph = defaultFrame?.children.find(block => block.flavour === 'affine:paragraph');
       if (
         firstParagraph &&
         matchFlavours(firstParagraph, ['affine:paragraph'])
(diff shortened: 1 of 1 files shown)


Flrande (approved):

Flrande:
Good catch!

Flrande:
It seems that we need to more accurately identify the frame block. Here are some reference codes.

```ts
const firstFrame = model.children.find((block) => block.flavour === "affine:frame")
if (!firstFrame) {
	const newFirstFrameId = page.addBlock("affine:frame", {}, null, 0)
	const newFirstParagraphId = page.addBlock("affine:paragraph", {}, newFirstFrameId, 0)
	asyncFocusRichText(page, newFirstParagraphId)
	return
}
const firstParagraph = firstFrame.children.find((block) => block.flavour === "affine:paragraph")
```

packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -248,7 +248,7 @@ export class DefaultPageBlockComponent
       return;
     } else if (e.key === 'ArrowDown' && hasContent) {
       e.preventDefault();
-      const firstParagraph = model.children[0].children[0];
+      const firstParagraph = model.children[1].children[0];
```
doodlewind, line 251: /cc @Saul-Mirone do you have a recommended way to query the first paragraph here?

doodlewind (approved):
I think this is a good catch that fixes the regression in #2554

packages/blocks/src/page-block/default/default-page-block.ts
Saul-Mirone: Please check the flavour of block.

doodlewind:
@donteatfriedrice I think it's good to merge as long as we get CI passed. Could you take a look a the test cases that failed and possibly do some tweak, so as to land this PR? Thanks!

donteatfriedrice:
@doodlewind Sorry, I have a question about the ArrowDown feature, based on the code logic currently, when press down ArrowDown, the cursor will move from title to the first paragraph, it will ignore the other block types like list. Is this how the feature designed? I attach a video below. Thanks for your reply.

https://github.com/toeverything/blocksuite/assets/99816898/e3feaa82-ebae-4a08-9e1f-8e11da946d30

doodlewind:
@donteatfriedrice good catch! I've added a quick update, feel free to modify if necessary!

donteatfriedrice:
> @donteatfriedrice good catch! I've added a quick update, feel free to modify if necessary!

Thanks! I think maybe it can be merged now..

donteatfriedrice:
@doodlewind I am wondering if it should also support code block?

doodlewind:
> @doodlewind I am wondering if it should also support code block?

Surely nice to have!

donteatfriedrice:
@doodlewind Now it supports ArrowDown from title to paragraph, list, code block. But it seems that there are some problems with the key down events of code block itselef. 
1. When the cursor is in the last code line, user press arrowdown, the cursor would not move to next block.
2. When cursor is at the block before code block, then press arrowdown, it will focus on the whole code block, not the first code line of this code block.
3.  When cursor is at the block after code block, then press arrowup, it will focus on the whole code block, not the last code line of this code block. 

https://github.com/toeverything/blocksuite/assets/99816898/a98f2472-c15c-4e1d-8250-cb9295565bf8

Should these bugs fixed by this issue (#2807), or by issue #2805 ?

doodlewind:
@donteatfriedrice great catch, having more standalone issues and PRs following our discussion will be really appreciated!

doodlewind:
This one is good to merge. I hope we can provide test cases regarding the cursor behavior in the upcoming improvements, thanks!

donteatfriedrice:
> This one is good to merge. I hope we can provide test cases regarding the cursor behavior in the upcoming improvements, thanks!

OK, thanks.

PR #2825, opened 2023-05-29
fix(playground): sync debug menu color transition with content area on switching dark mode

Innei:
Before:

https://github.com/toeverything/blocksuite/assets/41265413/55fea3dd-e923-4a04-9ecf-e2218f05203c

After: 

https://github.com/toeverything/blocksuite/assets/41265413/8a73abd8-de9e-4f6e-972a-3267bf65d310


doodlewind:
I tried but not sure if this is the expected styling:

image

doodlewind (approved):
LGTM, thanks for providing such a detailed improvement!

PR #2842, opened 2023-05-29
feat: add page data source

### LINKED ISSUE #2841: Add meta field indicating page data source
Add a marker to indicate where a page was created. In future, we may plan to create a page filter to find all data from an imported source, such as importing from Notion.

zuoxiaodong0815:
close #2841


doodlewind:
This will miss the current beta version, and we need to migration to subdoc format at the next version. It's better to implement it after the subdoc refactoring landed.

/cc @joebeijing

PR #2844, opened 2023-05-29
fix(blocks): quote block cursor movement

### LINKED ISSUE #2834: Last line can be ignored when moving cursor down in quote block
When moving the cursor down within a quote block, if the cursor is at the end of the second-to-last line, the last line is being ignored.

https://github.com/toeverything/blocksuite/assets/41265413/d23d4e7c-4171-4490-bf39-36826254fd34

donteatfriedrice:
To fix #2834

https://github.com/toeverything/blocksuite/assets/99816898/c59e320d-8f3b-403f-9b6a-00b8793f58b5


donteatfriedrice (commented):
When the cursor at the end of the last second line, the nextRangeRect's bottom plus its half of height will always be greater than the containerRect's bottom based on current code. Maybe should check the current range rect's bottom plus the next range rect' s height to see if at the last line. I am not sure about it since there are some e2e test errors... Maybe I should check it again...

PR #2881, opened 2023-05-30
fix: fix cursor position when using cmd+z hotkey

### LINKED ISSUE #2791: Cursor position incorrect when hitting enter at the end of inline code
https://github.com/toeverything/blocksuite/assets/20554850/1f57bcd7-884b-4c2f-8277-2622186ce3aa

smallnine9:
fix #2791


Flrande:
We may need to update [edgeless mode](https://github.com/toeverything/blocksuite/blob/aed319649daa5757b772d8fff39c2f0c5ad622e4/packages/blocks/src/page-block/edgeless/selection-manager.ts#LL246C23-L246C23) as well.

doodlewind (approved):

doodlewind:
Thanks for the fix!

smallnine9:
@doodlewind  But I found edgeless mode will have some special problem, maybe we need open another issue?

doodlewind:
@smallnine9 surely we can open new issue for this, thanks!

PR #2884, opened 2023-05-30
fix(blocks): do not convert markdown in quote block

### LINKED ISSUE #2836: Modifier text characters like `#` causes unexpected conversion of quote block
https://github.com/toeverything/blocksuite/assets/41265413/8887aadb-00ba-442a-a23a-24f245063e32

donteatfriedrice:
To fix #2836 

https://github.com/toeverything/blocksuite/assets/99816898/55a84247-5c33-4da2-abcb-352c1928975a

### DIFF
--- a/packages/blocks/src/__internal__/rich-text/markdown-convert.ts
+++ b/packages/blocks/src/__internal__/rich-text/markdown-convert.ts
@@ -1,5 +1,5 @@
 import { ALLOW_DEFAULT, PREVENT_DEFAULT } from '@blocksuite/global/config';
-import { assertExists, matchFlavours } from '@blocksuite/global/utils';
+import { assertExists, isEqual, matchFlavours } from '@blocksuite/global/utils';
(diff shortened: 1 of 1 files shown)


doodlewind (approved):
Seems to be a neat fix! Hope this can sneak pass CI :)

Innei:
I think it should be allowed to nest other blocks within a Quote Block?

This is typora behavior

CleanShot 2023-05-30 at 5 30 50@2x

donteatfriedrice:
> I think it should be allowed to nest other blocks within a Quote Block?
> 
> This is typora behavior
> 
> CleanShot 2023-05-30 at 5 30 50@2x

It seems that quote block doesn't support nesting other block type now? I try to copy list block into quote, but there are some problems, maybe we should create a new issue? I don't know whether blocksuite support nested quote block...

https://github.com/toeverything/blocksuite/assets/99816898/3d8f4900-6b01-4123-aecc-ee3970a21264

PR #2885, opened 2023-05-30
refactor(edgeless): remove redundant transact

Flrande:


doodlewind (approved):
Nice improvement, we need more cleanup in edgeless!

PR #2888, opened 2023-05-30
fix(blocks): database multi-select tab close icon invisible

### LINKED ISSUE #2887: The Close Icon of Database Multi-Select Tab is Invisible.
database-multi-select-tab-close-icon-invisible

https://github.com/toeverything/blocksuite/assets/99816898/4d4ca36f-95cc-4ff9-b8d9-5e040e5280b3

donteatfriedrice:
To fix #2887 

https://github.com/toeverything/blocksuite/assets/99816898/c612b368-6e55-4d86-9b82-ec501b8d4599

### DIFF
--- a/packages/blocks/src/database-block/table/components/column-type/select/select-cell-editing.ts
+++ b/packages/blocks/src/database-block/table/components/column-type/select/select-cell-editing.ts
@@ -97,12 +97,14 @@ const styles = css`
   .select-selected > .close-icon {
     display: flex;
     align-items: center;
-    fill: var(--affine-icon-color);
   }
 
   .select-selected > .close-icon:hover {
     cursor: pointer;
-    background: var(--affine-hover-color);
+  }
+
+  .select-selected > .close-icon > svg{
+    fill: var(--affine-black-90);
   }
 
   .select-option-new {
--- a/packages/blocks/src/database-block/table/components/column-type/select/select-cell-editing.ts
+++ b/packages/blocks/src/database-block/table/components/column-type/select/select-cell-editing.ts
@@ -98,15 +98,12 @@ const styles = css`
     display: flex;
     align-items: center;
   }
-
   .select-selected > .close-icon:hover {
     cursor: pointer;
   }
-
   .select-selected > .close-icon > svg{
     fill: var(--affine-black-90);
   }
-
   .select-option-new {
     display: flex;
     flex-direction: row;
@@ -138,7 +135,6 @@ const styles = css`
     width: 16px;
     height: 16px;
   }
-
   .select-option {
     display: flex;
     justify-content: space-between;
--- a/packages/blocks/src/database-block/table/components/column-type/select/select-cell-editing.ts
+++ b/packages/blocks/src/database-block/table/components/column-type/select/select-cell-editing.ts
@@ -101,7 +101,7 @@ const styles = css`
   .select-selected > .close-icon:hover {
     cursor: pointer;
   }
-  .select-selected > .close-icon > svg{
+  .select-selected > .close-icon > svg {
     fill: var(--affine-black-90);
   }
   .select-option-new {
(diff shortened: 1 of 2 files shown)


packages/global/src/config/icons/database.ts
```diff
@@ -259,7 +259,6 @@ export const DatabaseSearchClose = html`<svg
   width="16"
   height="16"
   viewBox="0 0 16 16"
-  fill="none"
```
zqran, line 262: We can keep `fill="none"` here.

packages/blocks/src/database-block/table/components/column-type/select/select-cell-editing.ts
```diff
@@ -97,6 +97,12 @@ const styles = css`
   .select-selected > .close-icon {
     display: flex;
     align-items: center;
+    fill: var(--affine-icon-color);
```
zqran, line 100: In our design system, the icon color here should be: 

```suggestion
    fill: var(--affine-black-90);
```

packages/blocks/src/database-block/table/components/column-type/select/select-cell-editing.ts
```diff
@@ -97,6 +97,12 @@ const styles = css`
...
+  }
+
+  .select-selected > .close-icon:hover {
+    cursor: pointer;
+    background: var(--affine-hover-color);
```
zqran, line 105: The hover `background` is not needed here, this is also by design.

packages/blocks/src/database-block/table/components/column-type/select/select-cell-editing.ts
zqran, line 100: In our design system, the icon color here should be:

```suggestion
    fill: var(--affine-black-90);
```

packages/blocks/src/database-block/table/components/column-type/select/select-cell-editing.ts
zqran, line 105: The hover `background` is not needed here, this is also by design.

packages/global/src/config/icons/database.ts
zqran, line 262: We can keep `fill="none"` here.

zqran (dismissed):

packages/blocks/src/database-block/table/components/column-type/select/select-cell-editing.ts
donteatfriedrice: Thanks.

packages/blocks/src/database-block/table/components/column-type/select/select-cell-editing.ts
donteatfriedrice: Thanks again. By the way, should I keep the cursor: pointer or not?

packages/global/src/config/icons/database.ts
donteatfriedrice: It seems when keep fill="none" here, and add  fill: var(--affine-black-90) in .close-icon, the icon will be still not visible. Any suggestion to fix it?

packages/blocks/src/database-block/table/components/column-type/select/select-cell-editing.ts
zqran: > Thanks again. By the way, should I keep the cursor: pointer or not?

Yes. Generally speaking, the clickable area should be indicated by the mouse style(`cursor: pointer`).

packages/global/src/config/icons/database.ts
zqran: That's fine too. Alternatively, you can set the fill style like `.close-icon > svg`.

donteatfriedrice: OK. Thanks for your advice.

packages/blocks/src/database-block/table/components/column-type/select/select-cell-editing.ts
donteatfriedrice: Thanks

doodlewind (approved):

PR #2905, opened 2023-05-31
fix(edgeless): auto add frame on click when no frame exists

### LINKED ISSUE #2846: Deleting the last/only object (in Edgeless) breaks page mode
After deleting the last object/frame in Edgeless mode, the Page mode breaks: it's empty and you cannot type or insert from Block Hub anything to it. You can only turn to Edgeless mode again to add something to recover the Page mode.

feedback (1)

regischen:
Closes #2846

### DIFF
--- a/packages/blocks/src/page-block/default/selection-manager/selection-manager.ts
+++ b/packages/blocks/src/page-block/default/selection-manager/selection-manager.ts
@@ -346,7 +346,9 @@ export class DefaultSelectionManager extends AbstractSelectionManager<DefaultPag
       keys: { shift },
     } = e;
 
-    if (this._ensureFrameExists() && isInsidePageTitle(target)) {
+    const hasAddedFrame = this._ensureFrameExists();
+
+    if (hasAddedFrame && isInsidePageTitle(target)) {
       requestAnimationFrame(() => {
         handleNativeRangeClick(this.page, e, this.container);
       });
--- a/packages/blocks/src/page-block/default/selection-manager/selection-manager.ts
+++ b/packages/blocks/src/page-block/default/selection-manager/selection-manager.ts
@@ -174,7 +174,7 @@ export class DefaultSelectionManager extends AbstractSelectionManager<DefaultPag
       ) {
         event.raw.preventDefault();
       }
-      if (this.page.hasFrameBlock()) {
+      if (this.page.hasFlavour('affine:frame')) {
         this._onContainerPointerMove(ctx);
       }
     });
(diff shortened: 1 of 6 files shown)


packages/blocks/src/page-block/default/selection-manager/selection-manager.ts
```diff
@@ -36,6 +36,7 @@ import {
   isEmbed,
   isImage,
   isInsidePageTitle,
+  isInTitleBlock,
```
doodlewind, line 39: Actually we don't have a title block, it's just the title field in the page block. And can we reuse the `isInsidePageTitle` method instead of adding a new one?

packages/blocks/src/page-block/default/selection-manager/selection-manager.ts
```diff
@@ -194,6 +197,15 @@ export class DefaultSelectionManager extends AbstractSelectionManager<DefaultPag
     );
   }
 
+  private _checkHasAnyFrameBlock() {
```
doodlewind, line 200: Semantically, a `checkSomething` method should not perform mutaions.

packages/store/src/workspace/page.ts
```diff
@@ -203,6 +203,10 @@ export class Page extends Space<FlatBlockMap> {
     );
   }
 
+  hasFrameBlock() {
```
doodlewind, line 206: We should avoid adding page level API that is hard-coded to some specific block types.

packages/blocks/src/page-block/default/selection-manager/selection-manager.ts
regischen: ok, I didn't notice there already has one method

packages/store/src/workspace/page.ts
regischen: ok, so I move this method to selection manager and turn it to private

tests/utils/actions/keyboard.ts
```diff
@@ -28,7 +28,7 @@ export async function type(page: Page, content: string, delay = 50) {
 export async function withPressKey(
   page: Page,
   key: string,
-  fn: () => Promise<void>
+  fn: () => Promise<void> = () => new Promise(resolve => resolve())
```
doodlewind, line 31: This method is designed to run the handler, so if you don't want to pass in the callback, using `pressBackspace` will be semantically a better fit.

regischen: yeah, I should have found there is pressBackspace method.

doodlewind (approved):
I think this is a better idea than our discussion result, good job!

Issue #2946, opened 2023-06-01
After pasting a link to generate a bookmark, undoing should not restore it to the loading state.


joebeijing:
https://github.com/toeverything/blocksuite/assets/22443345/c32c5ae7-4870-4fff-9ab3-753a7471c339

The expected outcome is to return to the link state directly after undoing.

QiShaoXuan:
fixed here in passing: https://github.com/toeverything/blocksuite/pull/2980

PR #2951, opened 2023-06-01
fix: redo/undo should not affect bookmark internal update

QiShaoXuan:
fixed: https://github.com/toeverything/blocksuite/issues/2946

### DIFF
--- a/packages/store/src/space.ts
+++ b/packages/store/src/space.ts
@@ -27,8 +27,6 @@ export class Space<
    */
   protected readonly _ySpace: Y.Map<State[keyof State]>;
 
-  shouldTransact = true;
-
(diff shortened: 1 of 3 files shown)


doodlewind (changes requested):
Adding this boolean parameter to block API may lead to maintenance issues. I'd prefer alternative approaches. 🤔️

packages/store/src/space.ts
```diff
@@ -27,6 +27,8 @@ export class Space<
    */
   protected readonly _ySpace: Y.Map<State[keyof State]>;
 
+  shouldTransact = true;
```
doodlewind, line 30: Should be `private _shouldTransact` 😝

QiShaoXuan: Okay, move _shouldTransact to Page class now~

doodlewind (approved):

PR #2965, opened 2023-06-02
fix(edgeless): fix continuous undo redo lost frame

### LINKED ISSUE #2715: Undo stack item lost after continuous pasting frames
1. Pasting frames for multiple times.
2. Keep undoing.
3. Keep redoing -> only 2 frames can be recovered.

https://github.com/toeverything/blocksuite/assets/7312949/5173f4b7-1e7a-415f-9eed-4d3e311fe67c

The operation of adding shapes or brushes are not effected. So this could possibly relate to redundant state mutation after adding frame, which will flush the undo stack.

regischen:
Closes #2715 
after redo, will trigger once resizeObserver event which cause the history stack chaotic. 
So If there is no change in size, then there is no need to update the size of frame block

### DIFF
--- a/packages/blocks/src/page-block/edgeless/frame-resize-observer.ts
+++ b/packages/blocks/src/page-block/edgeless/frame-resize-observer.ts
@@ -14,9 +14,10 @@ export class FrameResizeObserver {
    * So we need to cache observed element.
    */
   private _cachedElements = new Map<string, Element>();
+  private _lastRect = new Map<string, DOMRect>();
 
   slots = {
-    resize: new Slot<Map<string, DOMRect>>(),
+    resize: new Slot<Map<string, DOMRectReadOnly>>(),
   };
 
   constructor() {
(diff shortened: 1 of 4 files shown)


packages/blocks/src/page-block/edgeless/frame-resize-observer.ts
```diff
@@ -14,9 +14,10 @@ export class FrameResizeObserver {
    * So we need to cache observed element.
    */
   private _cachedElements = new Map<string, Element>();
+  private _lastRect = new Map<string, DOMRectReadOnly>();
```
doodlewind, line 17: `lastRects` aligns sementic better (with "s" that indicates there can be multiple objects). Also, is that somewhere required to cleanup this map?

doodlewind (approved):
Quite a good catch, thanks!

packages/blocks/src/page-block/edgeless/frame-resize-observer.ts
```diff
@@ -35,6 +36,19 @@ export class FrameResizeObserver {
       const blockElement = entry.target.closest(`[${BLOCK_ID_ATTR}]`);
       const id = blockElement?.getAttribute(BLOCK_ID_ATTR);
       if (!id) return;
+      if (this._lastRect.has(id)) {
+        const rect = this._lastRect.get(id);
+        if (
+          rect &&
+          rect.x === entry.contentRect.x &&
```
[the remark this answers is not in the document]
regischen: yes

packages/blocks/src/page-block/edgeless/frame-resize-observer.ts
regischen: yeah, we need to cleanup this map.

PR #2968, opened 2023-06-02
fix: bookmark copy icon and toast info 

### LINKED ISSUE #2966: Toast that appears after clicking the "Copy" button of Bookmark is incorrect
https://github.com/toeverything/blocksuite/assets/22443345/ea5b1dfd-979a-4184-8008-6ab9e956ee92

- It should be “Copied original link to clipboard“
- The Button's name should be “Copy original link” CC @Svaney-ssman

donteatfriedrice:
To fix #2966


donteatfriedrice:
Code in another pull request.

PR #2969, opened 2023-06-02
fix: delete parent lost nested children

### LINKED ISSUE #2899: If the parent node is removed, any nested children nodes will be lost
https://github.com/toeverything/blocksuite/assets/18554747/44aa2b92-e3a9-4cfa-879a-ce5faa18864f

### LINKED ISSUE #2966: Toast that appears after clicking the "Copy" button of Bookmark is incorrect

donteatfriedrice:
To fix #2899 
To fix #2966 

https://github.com/toeverything/blocksuite/assets/99816898/f5213a19-f463-4331-8ce0-b4fb22f1229b

### DIFF
--- a/packages/blocks/src/bookmark-block/components/bookmark-operation-popper.ts
+++ b/packages/blocks/src/bookmark-block/components/bookmark-operation-popper.ts
@@ -73,7 +73,7 @@ const operations: Operation[] = [
   {
     type: 'copy',
     icon: CopyIcon,
-    label: 'Copy',
+    label: 'Copy original link',
     action: (model, callback) => {
       copyBlocks({
         type: 'Block',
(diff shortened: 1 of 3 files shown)


lawvs (commented):
Nice catch! could you add some e2e tests for it?

donteatfriedrice:
> Nice catch! could you add some e2e tests for it?

Sure. I have added an E2E test case.

lawvs (approved):

doodlewind (approved):
This is also bothering me when using AFFiNE, thanks for the fix!

PR #2973, opened 2023-06-04
feat: handling left key press for cursor transition from paragraph to page title

tillkwl:
This pull request involves modifications in the keymap and rich-text-operations files with the aim of refining the handling of Left Key presses.

The main objective behind this implementation is to allow the cursor to move to the Page Title when the Left Key is pressed at the start of the first Paragraph block. To accomplish this, the handlers are set to return PREVENT_DEFAULT when the preceding block is affine:surface, intending to emulate the typical navigation observed in rich text editing interfaces.

However, the current implementation does not yield the expected results - pressing the Left Key at the beginning of the first Paragraph block does not relocate the cursor to the Page Title.

Even though at the handleKeyDown you can transition from Page Title to Paragraph for example doing it like that.

I welcome any insights or advice regarding why this might be occurring, and what potential solutions could resolve this issue. Your expertise would be highly valuable in addressing this, thereby improving the overall user experience in our text interface.

Thank you in advance for your assistance and dedication to enhancing our project.


pnpm-lock.yaml
```diff
@@ -1,4 +1,8 @@
-lockfileVersion: '6.0'
```
doodlewind, line 1: Seems that this should be reverted to prevent CI failure.

packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
```diff
@@ -544,6 +544,20 @@ export function handleLineStartBackspace(page: Page, model: ExtendedModel) {
   handleUnknownBlockBackspace(model);
 }
 
+function handleParagraphBlockLeftKey(page: Page, model: ExtendedModel) {
+  if (!matchFlavours(model, ['affine:paragraph'])) return false;
+  const previousBlockFlavour = getPreviousBlock(model)?.flavour;
+  if (previousBlockFlavour == 'affine:surface') {
```
doodlewind, line 550: Best practice is to keep using `matchFlavours` here. Also we prefer `===` :)

doodlewind:
@Flrande any advice for focusing to the end of page title via our inner API? 🤔

Flrande:
> @Flrande any advice for focusing to the end of page title via our inner API? 🤔

We can use `focusEnd()` in virgo and we can get vEditor for page title by [titleVEditor](https://github.com/toeverything/blocksuite/blob/9927a75adbf11c05975324fb3a19fa48b129d86b/packages/blocks/src/page-block/default/default-page-block.ts#L175)

tillkwl:
Any idea why the pipeline fails?

packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
```diff
@@ -542,6 +543,28 @@ export function handleLineStartBackspace(page: Page, model: ExtendedModel) {
...
+  if (!matchFlavours(model, ['affine:paragraph'])) return false;
+  const pageElement = getDefaultPage(page);
+  if (!pageElement) {
+    // Maybe in edgeless mode
+    return null;
+  }
+  const titleVEditor = pageElement.titleVEditor;
+  const previousBlock = getPreviousBlock(model);
+  if (previousBlock !== null) {
+    if (matchFlavours(previousBlock, ['affine:surface'])) {
```
Flrande, line 556: I feel that the way judgments are made here is somewhat unreliable. Do you have a better approach to handle it? @Saul-Mirone

Flrande:
I made a simple modification to the judgment logic and added a test. Thank you very much for your contribution!

packages/blocks/src/__internal__/rich-text/rich-text-operations.ts
```diff
@@ -542,6 +543,32 @@ export function handleLineStartBackspace(page: Page, model: ExtendedModel) {
...
+  if (!pageElement) {
+    // Maybe in edgeless mode
+    return;
+  }
+  const titleVEditor = pageElement.titleVEditor;
+  const parent = page.getParent(model);
+  if (parent && matchFlavours(parent, ['affine:frame'])) {
+    const frameParent = page.getParent(parent);
+    if (frameParent && matchFlavours(frameParent, ['affine:page'])) {
+      const frameIndex = frameParent.children.indexOf(parent);
```
Saul-Mirone, line 558: The page block may contains other blocks such as `affine:surface`, you can filter the children before check the index of frame, for example:
```ts
frameParent.children.filter(block => matchFlavours(block, ['affine:frame'])).indexOf(parent)
```

PR #2993, opened 2023-06-06
fix: pasting code is not as expected

### LINKED ISSUE #2991: The behavior of pasting code is not as expected
### What happened?

In fact, after copying the following code to the code block, it will be pasted multiple times

```
const axios = require('axios');
const PDFDocument = require('pdfkit');
const fs = require('fs');

async function getIssues(user, repo) {
  const url = `https://api.github.com/repos/${user}/${repo}/issues?state=all`;
  const response = await axios.get(url);
  return response.data;
}

function formatIssues(issues) {
  return issues.map(issue => {
    return `Title: ${issue.title}\n\n${issue.body}\n\n---------\n\n`;
  }).join('\n');
}

function generatePDF(content, filename) {
  const doc = new PDFDocument;
  doc.pipe(fs.createWriteStream(filename));
  doc.text(content);
  doc.end();
}

async function main() {
  const user = 'github';
  const repo = 'github';
  const issues = await getIssues(user, repo);
  const content = formatIssues(issues);
  generatePDF(content, 'issues.pdf');
}

main().catch(console.error);

```
image

### Version

macOS ARM 64

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

_No response_

### Anything else?

_No response_

donteatfriedrice:
To fix #2991 

https://github.com/toeverything/blocksuite/assets/99816898/dca3ae85-8d0e-4646-98a0-3d2590e2ba9c

### DIFF
--- a/packages/blocks/src/__internal__/content-parser/index.ts
+++ b/packages/blocks/src/__internal__/content-parser/index.ts
@@ -231,8 +231,8 @@ export class ContentParser {
 
   public text2blocks(text: string): SerializedBlock[] {
     return text.split('\n').map((str: string) => {
-      const splitText = text.split(this.urlPattern);
-      const urls = text.match(this.urlPattern);
+      const splitText = str.split(this.urlPattern);
+      const urls = str.match(this.urlPattern);
       const result = [];
 
       for (let i = 0; i < splitText.length; i++) {


doodlewind (approved):
Good catch, thanks!

PR #2994, opened 2023-06-06
feat: complement and add some edgeless utilies

regischen:
add some methods in  Point, Bound class
and add Line.intersect method

### DIFF
--- a/packages/phasor/package.json
+++ b/packages/phasor/package.json
@@ -16,7 +16,6 @@
   "author": "toeverything",
   "license": "MPL-2.0",
   "devDependencies": {
-    "jsdom": "^22.1.0",
     "nanoid": "^4.0.1",
     "yjs": "^13.6.0"
   },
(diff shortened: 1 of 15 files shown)


packages/phasor/package.json
```diff
@@ -37,6 +37,7 @@
     }
   },
   "dependencies": {
+    "@blocksuite/blocks": "workspace:*",
```
doodlewind, line 40: This leads to circular deps and should be reverted.

regischen: > This leads to circular deps and should be reverted.

ok, I get it.
because I want to reuse some basic code library without repeating. 
if we want to use the Point Class in block packages, Should we re-write the Point class in phasor package or 
should we extract the some basic codes into a new package that is a very basic packages for use of other packages. Maybe just called basic or math?
Which do you perfer？

PR #2996, opened 2023-06-06
feat(edgeless): add text align button

Flrande:


doodlewind (commented):
I found the text cursor doesn't show up when creating new text element, is it a regression?

Flrande:
> I found the text cursor doesn't show up when creating new text element, is it a regression?

I can't reproduce it. Any screen record?

doodlewind:
I double checked and not reproduced. Good to merge.

Issue #2921, opened 2023-05-31
​forward delete at the end of block to merge


youkq95:
### Description

When editing contents across blocks, someone might want to merge the content of two blocks by:
1. back delete from the starting of the second block, or
2. forward deleted from the ending of the first block.

Now on the affine PC desktop can do it the first way, but not in the second way. 

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

doodlewind:
This sounds very interesting! For macOS users, the only key for deleting content is `Backspace`, which deletes content backward. So deleting content forward was not specifically supported. Hope for PRs that makes this happen!

For anyone interested in this, see #2552 that demonstrates how hotkey binding works.

Alkaidcc:
fixed in #2552 @youkq95

doodlewind:
Closed by #2552, thanks @Alkaidcc for helping clean up our backlog!

PR #2552, opened 2023-05-12
feat(blocks): support delete hotkey

Alkaidcc:
fix #2550

PR #2998, opened 2023-06-06
[WIP] feat: add forward delete keybinding at the end of block

ahonn:
I saw an issue (#2921) on [BlockSuite Project](https://github.com/orgs/toeverything/projects/22/views/10) regarding the unresponsiveness of forward delete at the end of a block, so I tried to solve this problem.

I didn't have a thorough understanding of the implementation details of this project, so I found the solution by tracing breakpoints through DevTools. As shown in the TODO in the code, should I implement corresponding functions such as `handleParagraphBlockDelete`/`handleCodeBlockDelete` for `handleLineEndDelete`? 

If implementing it in this way is feasible or if there are better solutions available, please let me know.


ahonn:
Sorry, I didn't notice that it was duplicated with #2552

PR #3009, opened 2023-06-07
feat(edgeless): add common math utils

regischen:


doodlewind (changes requested):
I think the `Point` class looks like a duplicate of `Vec`, so merging them sounds much better than having such redundancy in our base math dependencies.

packages/phasor/src/utils/line.ts
```diff
@@ -0,0 +1,38 @@
+import { clamp } from './math-utils.js';
+import { almostEqual, EPSILON, MACHINE_EPSILON } from './math-utils.js';
+import type { Point } from './point.js';
+
+export class Line {
```
doodlewind, line 5: Looks better to have `lineIntersects` in `math-utils`  rather than adding a redundant class. Also it's suggested to make it depend on `Vec` instead of `Point`.

doodlewind: `Vec.cpr` and `Vec.lrp` are the existing methods to be reused.

packages/phasor/src/utils/point.ts
```diff
@@ -0,0 +1,90 @@
+import { clamp } from './math-utils.js';
+
+export interface IPoint {
+  x: number;
+  y: number;
+}
+
+export class Point {
```
doodlewind, line 8: Suggest not using this class. Also refactoring the `edgeless-toolbar` and the newly added `Bound` methods without this dependency sounds better. Otherwise we have too much to migrate for enforcing this convention.

packages/phasor/src/utils/line.ts
regischen: okay

packages/phasor/src/utils/point.ts
regischen: ok, looks like the codes are more like functional-style. not OOP style.

doodlewind (approved):
LGTM

Issue #2449, opened 2023-05-08
Support for multiple views in Database Block


zzj3720:
Considering that we already have two view modes, users may need to use both mode in a single data source.
<img width="429" alt="image" src="https://user-images.githubusercontent.com/17165520/236783595-a7e32b4a-d786-43a7-9eba-40ef21d79fa4.png">

There are also sorting and filtering that would benefit from this

I can help implement this if needed

zzj3720:
supported in #3010

Issue #2426, opened 2023-05-06
Enhance the type system of the Database Block


zzj3720:
Currently, The database has simple column types such as https://github.com/toeverything/blocksuite/blob/1a83852d8c582664cda9cffa8753d0818cd9bdd8/packages/blocks/src/database-block/table/types.ts#L14-L20
Trying to implement filtering and sorting and much more may require more powerful type system.
### For example
<div style="display:flex">
<img width="250" alt="image" src="https://user-images.githubusercontent.com/17165520/236609316-89f1e579-7cfd-41fc-bee9-846781138ea5.png">
<img width="250" alt="image" src="https://user-images.githubusercontent.com/17165520/236609321-6e3e30d4-1c59-4135-a3f1-f661bc4877b8.png">
</div>
<img width="349" alt="image" src="https://user-images.githubusercontent.com/17165520/236609329-049978bb-5268-40e5-8638-711fe015eca6.png">
The URL is the same as the Text, File is partially identical to the first two.

If described in terms of "subtyping" in TypeScript, it might look like this
```ts
interface Any {
  isEmpty():boolean;
  isNotEmpty():boolean;
}
interface String extends Any{
  is(str:string):boolean;
  isNot(str:string):boolean;
  contains(subStr:string):boolean;
  doseNotContains(subStr:string):boolean;
  startsWith(subStr:string):boolean;
  endsWith(subStr:string):boolean;
}
interface File extends Any{};
interface Text extends String{};
interface URL extends String{};
```

### Benefits
Once we enhance the type system of the Database, 
- Automatically provide filtering and sorting options for new column type based on data type.
- We can even know the types of Formula columns. such as number+number results in number, string+string results in string, we can then know how to filter sort and format them or refer it again in other Formula column.
- We can support nested data types, such as putting any primitive type into Array type to form a multi-select, and still know how to handle it.
- etc.

If this idea aligns with the goals of the project, I can help design and implement the related feature. I have some practical experience in implementing simple type system on Structural Editor.

zzj3720:
I'm working on a demo.

doodlewind:
Quick idea: we can support the filter mechanism on the "All Pages" view in AFFiNE

zzj3720:
supported in #3010

PR #3010, opened 2023-06-07
feat: refactor to support multi-view and filter for database block

zzj3720:
relate: #2449 #2426 

- [x] Refactor to support for multi-view in database block.
- [x] Add a simple type system that can infer and match functions that match the requirements.
- [x] Add an AST that describes the behavior of the filter.
- [x] Add the AST interpreter.
- [x] Add migration logic


doodlewind (commented):
This is huge! I think we need a feature flag to opt-out this.

zqran:
I found some breaking issues:

1. The UI for `number` needs to be updated.
2. `Number` cell cannot change from cell selection to row selection when `esc` key is pressed.

Issue #2934, opened 2023-06-01
Support modifying tag color for the "Select" field in database


joebeijing:

hwangdev97:
The sub-menu appear as follows:

<img width="655" alt="shot-2023-06-05-18 19 20@2x" src="https://github.com/toeverything/blocksuite/assets/24713927/18222f61-f45b-45b2-a1c6-cf61c4916a65">

@zqran cc @doodlewind

zqran:
> The sub-menu appear as follows:
> 
> <img alt="shot-2023-06-05-18 19 20@2x" width="655" src="https://user-images.githubusercontent.com/24713927/243325082-18222f61-f45b-45b2-a1c6-cf61c4916a65.png">
> @zqran cc @doodlewind

Get!

PR #3015, opened 2023-06-07
feat(database): support select option to modify color

zqran:
close: #2934 

---

Many popups are used in Database, for example, the function of modifying select option color in this PR uses 3 popups.

However, when `Popper` is currently used to process positioning, the calculated position does not meet expectations, which is confusing.

So, [`Floating UI`](https://github.com/floating-ui/floating-ui)(a new version of Popover) is used in this PR to replace `Popper`.

https://github.com/toeverything/blocksuite/assets/15389209/34b396a3-2788-419c-89b0-340418409665


doodlewind (commented):
It ain't not good having two popper libraries co-existing. Is it possible to replace `@popperjs/core` since their API looks somewhat similar?

If possible, this could happen in following refactoring PRs, but I surely won't expect having duplicated deps as a common practice.

zqran:
> It ain't not good having two popper libraries co-existing. Is it possible to replace `@popperjs/core` since their API looks somewhat similar?
> 
> If possible (sounds luring accoring to https://floating-ui.com/docs/migration), this could happen in following refactoring PRs, but I surely won't expect having duplicated deps as a common practice.

Yes, this PR is an attempt to use `floating-ui`, which is better than `@popperjs/core` in terms of usage.

So, next I will replace all other `@popperjs/core` with `floating-ui`.

doodlewind (approved):

PR #3028, opened 2023-06-09
feat: add block on clicking page bottom when necessary

QiShaoXuan:
1. if click bottom of page and the end block is embed, editor should insert a new editable block
2. if click bottom of page and the end block is editable, the last editable block should be focused

### DIFF
--- a/tests/selection/block.spec.ts
+++ b/tests/selection/block.spec.ts
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 import type { Page } from '@playwright/test';
 import { expect } from '@playwright/test';
+import { assert } from 'console';
 
 import {
   activeEmbed,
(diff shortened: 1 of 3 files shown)


doodlewind (approved):
Thanks for the improvement!

tests/selection/block.spec.ts
```diff
@@ -1172,3 +1173,58 @@ test('when shift-click should select correct number of list blocks', async ({
...
+    }
+    viewport.scrollTo(0, 1000);
+  });
+
+  const pageRect = await page.evaluate(() => {
+    const pageBlock = document.querySelector('affine-default-page');
+    return pageBlock?.getBoundingClientRect() || null;
+  });
+
+  assert(pageRect !== null);
```
lawvs, line 1197: You should use `expect` from playwright

PR #3030, opened 2023-06-09
fix(edgeless): correct the connecotor's handle behavior

blurname:
The originController in closure is not updated, then it casued the incorrect calculation of connector path (especially `customized start/end`).
before:

https://github.com/toeverything/blocksuite/assets/50226652/3761d779-43d0-4ec7-b3d6-c6ac35446088

after:

https://github.com/toeverything/blocksuite/assets/50226652/b35eeda2-0f0b-41bb-8d92-6ca47db25a44


doodlewind (approved):
LGTM, thanks!

PR #3041, opened 2023-06-11
feat(virgo): support embed

Flrande:
Make Virgo support making a specific `v-element` only selectable as a whole, by passing a function to Virgo when initing it.

For example, if you want to make a `v-element` of type `code` only selectable as a whole, you can pass a function to Virgo like this.

```ts
const vEditor = new VEditor(yText, {
    embed: delta => !!delta.attributes?.code,
});
```

The effect is as follows.

https://github.com/toeverything/blocksuite/assets/50035259/95f6323a-48ab-4d9f-a0de-bb0466aae596


doodlewind (commented):
Any design description or API documentation to share?

Flrande:
> Any design description or API documentation to share?

Updated.

PR #3070, opened 2023-06-13
fix(edgeless): fit to screen not working

### LINKED ISSUE #3060: Fit to Screen function is not working
Unable to show all content at once in the window.

regischen:
close #3060 

the original implment just consider only one frame block.
so just considering all frame blocks will be fine.


doodlewind (approved):
LGTM

Issue #2389, opened 2023-05-03
Clarify types of providers and support `workspace.isEmpty`


Himself65:
After updating several versions of the code in AFFiNE, we found that there are actually two types of providers: background providers and necessary providers. For the necessary provider, its purpose is to determine whether a workspace is truly empty (`workspace.isEmpty`) when we need to initialize it with the hello world page, rather than whether _items already exist. This is because when the user first enters the interface, all workspace.isEmpty must be true because the data read from indexeddb is later than UI loading. Therefore, we need a skeleton screen and related logic to check whether it is truly empty. The necessary provider is loaded before we can safely determine whether a workspace is empty.

As for the background provider, its purpose is actually very simple. We do not need to consider whether it is loaded; we need to call connect and disconnect on time.

```ts
interface NecessaryProvider {
  flavour: string;
  sync: () => Promise<void>;
}

interface BackgroundProvider {
  flavour: string;
  connect: () => void;
  disconnect: () => void;
}

class Workspace {
  get isEmpty(): Promise<boolean> | boolean {
     // ...
  }
}
```

PR #3073, opened 2023-06-13
feat(store): clarify different types of providers

Himself65:
Fixes: https://github.com/toeverything/blocksuite/issues/2389


doodlewind (changes requested):
Please just wait a while for me here for the naming.

doodlewind (approved):

PR #3081, opened 2023-06-13
feat(edgeless): keep viewport state between mode switch

### LINKED ISSUE #2479: Before and after switching to Edgeless, the previous zoom ratio and position when Edgeless was opened should be remembered
Currently, when frequently switching, Edgeless returns to its default state every time

regischen:
close #2479

### DIFF
--- a/packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
+++ b/packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
@@ -884,7 +884,8 @@ export class EdgelessPageBlockComponent
 
   private _trySetupViewportData() {
     const { viewport } = this.surface;
-    const viewportData = localStorage.getItem(this.page.id);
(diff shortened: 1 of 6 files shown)


packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
```diff
@@ -887,6 +882,63 @@ export class EdgelessPageBlockComponent
     });
   }
 
+  private _trySetupViewportData() {
+    const { viewport } = this.surface;
+    const viewportData = localStorage.getItem(this.page.id);
```
doodlewind, line 887: Would require some convention here. Any recommendation @Himself65?

doodlewind (commented):
Looks pretty good generally, please just wait for a bit for some styling cleanup 😉

doodlewind (approved):

PR #3118, opened 2023-06-15
feat(edgeless): add view layer for surface element

regischen:
add a view layer for surface element.
we can alter the render result without changing the crdt model.
all crdt change will be reflected on view.
and view change will result in re-render.

### DIFF
--- a/packages/phasor/src/elements/surface-element.ts
+++ b/packages/phasor/src/elements/surface-element.ts
@@ -123,7 +123,6 @@ export class SurfaceElement<
     for (const key in updates) {
       this.yMap.set(key, updates[key] as T[keyof T]);
     }
-    this.applyViewUpdate(updates as Partial<K>);
   }
 
   applyViewUpdate(updates: Partial<K>) {
--- a/packages/phasor/src/elements/surface-element.ts
+++ b/packages/phasor/src/elements/surface-element.ts
@@ -13,6 +13,11 @@ export interface ISurfaceElement {
   index: string;
   seed: number;
 }
+export type SurfaceElementKey = keyof ISurfaceElement;
+export interface ISurfaceElementView extends ISurfaceElement {
+  display: boolean;
+  opacity: number;
+}
 
 export interface HitTestOptions {
   expandStroke: boolean;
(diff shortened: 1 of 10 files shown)


doodlewind:
I haven't seen a clear definition and difference between the concept of _element_ and _view_ yet. Could you elaborate?

regischen:
> I haven't seen a clear definition and difference between the concept of _element_ and _view_ yet. Could you elaborate?

the idea is how to change the render result of the element without chaning the model.

So I design the concept, view.
The view belongs to the element, is a property of element, reflects how is the element view property or render result.
It's not crdt driven. and now only view property changed, the renderer re-renders.
meantime, the changes to the model of the element will sync to the the view of the element.

doodlewind:
Do you have a new feature to implement based on this design? I think it would be better to review based on the prototype, rather than adding new infra design without actual usage. (That said, these previously added unused util methods are also something I am reluctant to merge.)

regischen:
> Do you have a new feature to implement based on this design? I think it would be better to review based on the prototype, rather than adding new infra design without actual usage. (That said, these previously added unused util methods are also something I am reluctant to merge.)

yes, totally, All these codes I have committed is useful for the following feature development.
the reason why I am doing so is I think these are bascis. so I split them apart in order to not make the pr so big.
for example the view will be helpful for the connector and earser tool. The previous math utils will be used in eraser tool.
If you think it would be better that feature and design are together committed. I can also accept that!

packages/phasor/src/elements/surface-element.ts
```diff
@@ -13,6 +13,11 @@ export interface ISurfaceElement {
   index: string;
   seed: number;
 }
+export type SurfaceElementKey = keyof ISurfaceElement;
+export interface ISurfaceElementView extends ISurfaceElement {
+  display: boolean;
```
doodlewind, line 18: I don't feel right about making these two properties local-only. What if they need to be collaboratively shared?

So I suggest not to introduce this _view_ concept. Instead, what I think about is introducing concepts like _LocalElementModifier_s that can "shadow" CRDT state field by local ones. Each modifier is identified by an element id, and its fields have higher priority than CRDT element props when an element is being rendered.

/cc @Flrande

doodlewind (changes requested):

packages/phasor/src/elements/surface-element.ts
regischen: now making these two properties local-only is because we haven't use these two properties in today's functionality.(
for exmaple, we havn't support that element can change its opacity, but eventually it will I believe.)
So Ii it needs to be collaborative, then we just add them to crdt model. And the view interface inherits the model interface.

regischen: Yes, what we need is to "shadow" CRDT field.

Flrande: One simple approach I can think of is to maximize the shareability of states, so that transparency, which does not affect interactions and only affects visual effects, can be directly incorporated into CRDT?

Flrande: It seems that it would be more reasonable to place the display field I previously added only within the TextElement. Instead of using a public field for the surface to control the rendering of this element, I should directly control the text rendering through the render function in TextElement.

PR #3126, opened 2023-06-15
feat(edgeless): edgeless earser tool

### LINKED ISSUE #2943: Support eraser tool in the whiteboard

regischen:
closes #2943


Flrande:
If we incorporate transparency into CRDT, would there be any need for an additional view layer?

regischen:
> If we incorporate opacity into CRDT, would there be any need for an additional view layer?
It's definite that transparency would be into CRDT in the future. 
But in this case, I think it still needs. 
because when we use eraser tool to erase element. The element be erased will show opacity. It's a state indicator instead of changing of actual opacity.  It's a middle state.
And I think in the future, we still need some "middle state" in view which actually doesn't not change the crdt model.
Because some interactions need to be processed, we perhaps shouldn't change the CRDT's model before these interactions are completed. However, during the process of these interactions, rendering may potentially change.

regischen:
> If we incorporate transparency into CRDT, would there be any need for an additional view layer?

It's definite that transparency would be into CRDT in the future.
But in this case, I think it still needs.
because when we use eraser tool to erase element. The element be erased will show opacity. It's a state indicator instead of changing of actual opacity. It's a middle state.
And I think in the future, we still need some "middle state" in view which actually doesn't not change the crdt model.
Because some interactions need to be processed, we perhaps shouldn't change the CRDT's model before these interactions are completed. However, during the process of these interactions, rendering may potentially change.

Flrande:
>  we still need some "middle state" in view which actually doesn't not change the crdt model.

I understand that there might be such an intermediate state, but I'm unsure if it requires adding a global code hierarchy. Would placing this intermediate state as a field within the element suffice for this purpose? My main concern is whether the view would overlap with the existing responsibilities of the surface since both essentially involve CRUD operations on whiteboard elements. Can you provide other examples where this intermediate state is necessary?

regischen:
> > we still need some "middle state" in view which actually doesn't not change the crdt model.
> 
> I understand that there might be such an intermediate state, but I'm unsure if it requires adding a global code hierarchy. Would placing this intermediate state as a field within the element suffice for this purpose? My main concern is whether the view would overlap with the existing responsibilities of the surface since both essentially involve CRUD operations on whiteboard elements. Can you provide other examples where this intermediate state is necessary?

the view is not involved with crdt operations. right?
another example is

https://github.com/toeverything/blocksuite/assets/58546692/43bd68b4-4dfe-4d8f-b3e2-e155dc31fdac

if in the future, we have mindmap. When we drag one child element to change it's position. I think there is a middle state.

doodlewind (commented):
Should wait for #3118

doodlewind (approved):

PR #3128, opened 2023-06-15
refactor: migrate slash menu to widget component

lawvs:
## Design

THIS DESIGN IS OUTDATED!

```ts
const editor = document.querySelector("editor-container");
const defaultPage = editor.querySelector("affine-default-page");
const edgelessPage = editor.querySelector("affine-edgeless-page");

// import { SlashMenu } from '@blocksuite/blocks'
const SlashMenu = defaultPage.components.SlashMenu;

class CustomSlashMenu extends SlashMenu {
  override get menuGroups() {
    return super.menuGroups.slice(0, 1);
  }
}
// Fix `Illegal constructor` error
// see https://stackoverflow.com/questions/41521812/illegal-constructor-with-ecmascript-6
customElements.define("custom-slash-menu", CustomSlashMenu);

defaultPage.components.SlashMenu = CustomSlashMenu;
edgelessPage.components.SlashMenu = CustomSlashMenu;
```

Screenshot 2023-06-16 at 3 43 30 AM

### DIFF
--- a/packages/blocks/src/page-block/default/default-page-block.ts
+++ b/packages/blocks/src/page-block/default/default-page-block.ts
@@ -137,7 +137,7 @@ export class DefaultPageBlockComponent
    */
   components = {
     dragHandle: <DragHandle | null>null,
-    SlashMenu: <typeof SlashMenu | null>SlashMenu,
+    slashMenu: <typeof SlashMenu | null>SlashMenu,
   };
 
   mouseRoot!: HTMLElement;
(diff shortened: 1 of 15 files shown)


packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
```diff
@@ -184,6 +185,7 @@ export class EdgelessPageBlockComponent
   components = {
     dragHandle: <DragHandle | null>null,
     toolbar: <EdgelessToolbar | null>null,
+    SlashMenu: <typeof SlashMenu | null>SlashMenu,
```
himself65: lowercase?

packages/blocks/src/page-block/default/default-page-block.ts
```diff
@@ -128,6 +129,7 @@ export class DefaultPageBlockComponent
    */
   components = {
     dragHandle: <DragHandle | null>null,
+    SlashMenu: <typeof SlashMenu | null>SlashMenu,
```
Saul-Mirone, line 132: Should be `slashMenu`

PR #3146, opened 2023-06-16
feat(edgeless): support drag canvas with middle button of mouse

### LINKED ISSUE #3033: Pressing the mouse wheel can activate Hand mode and allow dragging of the canvas
- This refers to pressing the mouse wheel, not scrolling the mouse wheel
- Pressing activates Hand mode, and releasing returns to Select mode.

doouding:
Closes #3033 

There're two things worth sharing:

1. Unlike pressing the primary mouse button, when the middle button is pressed and the mouse moves outside the document, it does not trigger the `pointermove` event.
2. After switching back to the current tab, the `pointermove` event won't be triggered if the middle button is pressed without first clicking on the page.


doodlewind (approved):
This looks pretty neat, thanks!

PR #3148, opened 2023-06-17
fix: handling backspace key press for cursor transition from paragraph to page title

### LINKED ISSUE #2976: Cannot remove the first line using delete or forward delete
https://github.com/toeverything/blocksuite/assets/70709113/c5110f66-6ad7-458f-96e2-8db9a3d15e2c

Try deleting the first line using delete or forward-delete while your cursor is in the first line. It does not work - but it does work in other editors like notion.

I'm expecting it to work with **both** delete and forward-delete

zhengjitf:
Fix #2976, about the delete action. (For forward-delete fixed by https://github.com/toeverything/blocksuite/pull/2552)

### DIFF
--- a/.github/CLA.md
+++ b/.github/CLA.md
@@ -78,3 +78,4 @@ Example:
 - Till, @tillkwl, 2023/06/04
 - Yuexun Jiang, @ahonn, 2023/06/06
 - Hyden Liu, @HydenLiu, 2023/06/11
+- zhengjitf, @zhengjitf, 2023/06/17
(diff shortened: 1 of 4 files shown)


doodlewind (changes requested):
Thanks for your contribution! But I think for the first paragraph block, we should not simply delete it in this case. Otherwise there will be no content inside the note container, which will make it unable to focus to the content area:

image

So as an alternative, I think simply focus the cursor to the end of the title (without deleting the first paragraph block) can be preferable. If you are open to further investigation, feel free to ask for help!

zhengjitf:
Added a judgment, if the paragraph has no next sibling, won't delete it, just clear text. Is this OK?


https://github.com/toeverything/blocksuite/assets/17249775/488e17b3-59b4-4e67-b1df-b1b8bd9c181e

doodlewind (approved):

doodlewind (commented):
Sorry for the delay, thanks for your contribution! Also feel free to join our CN community at https://t.me/AffineDevCN if that helps :)

PR #3149, opened 2023-06-17
feat(phasor): support text in shape element

### LINKED ISSUE #2441: Support adding text above the shape

Flrande:
Close #2441


Flrande:
continue in #3423

Issue #2939, opened 2023-06-01
surface `elements` property should has prefix like others


Flrande:
![df7b95fe-1a87-4968-91bf-f03e22f5070c](https://github.com/toeverything/blocksuite/assets/50035259/fa370064-ad44-48af-af96-ab48448a5d33)

Flrande:
related to:
https://github.com/toeverything/blocksuite/blob/81b4c523ae71343b7c2f09d1387a46bcdf096aa1/packages/store/src/utils/utils.ts#L75
https://github.com/toeverything/blocksuite/blob/4ab9958701fcef765b2ab38c31eb1b55cecadfc5/packages/store/src/workspace/workspace.ts#L378

PR #3156, opened 2023-06-18
refactor(blocks): make elements as prop of surface block

Saul-Mirone:
Closes: https://github.com/toeverything/blocksuite/issues/2939


doodlewind (commented):
This is so damn good!

PR #3157, opened 2023-06-18
feat(store): let user control subdoc load timing after page meta added

Saul-Mirone:


tests/utils/actions/misc.ts
```diff
@@ -369,8 +379,11 @@ export async function initEmptyDatabaseWithParagraphState(
   page: Page,
   pageId?: string
 ) {
-  const ids = await page.evaluate(pageId => {
+  const ids = await page.evaluate(async pageId => {
     const { page } = window;
+    await page.waitForLoaded();
+    page.trySyncFromExistingDoc();
```
doodlewind, line 385: Is this API call required for the application side? Would be better to keep this `syncExistingDoc` method process private.

Saul-Mirone: Done

PR #3158, opened 2023-06-18
fix: cursor position on changing block types

### LINKED ISSUE #3152: The caret will be at wrong position after transforming paragraph block to another type and input new character.
Steps to reproduce:
1. Input in a paragraph block, like text, h1, h2...
2. Transform block type to another paragraph type using slash menu (type '/' )
3. Input new character.

https://github.com/toeverything/blocksuite/assets/99816898/295a0f7e-8c5d-4ff7-a3c8-7d671e418d94

It seems like this problem only occurs when convert type within paragraph block. 

When transform paragraph block to other flavour blocks (like list, code) or transform from other flavour blocks to paragraph, it works as expected.

smallnine9:
fix #3152 
Thanks for your review 🫶


Flrande:
Can you add a test for it? 😉

smallnine9:
sure

packages/blocks/src/page-block/utils/container-operations.ts
```diff
@@ -550,11 +550,14 @@ export function handleKeydownAfterSelectBlocks({
 }
 export async function onModelTextUpdated(
   model: BaseBlockModel,
-  callback: (text: RichText) => void
+  callback?: (text: RichText) => void
 ) {
+  console.log('onModelTextUpdated');
```
lawvs, line 555:
```suggestion
```

smallnine9: will delete it in next commit, thanks for pointing it! I should be more careful

smallnine9:
``Hey guys, I need your help. 👏
The test which I added needs to import `updateBlockType`.
But when I run playwright test I found this error:

> Decorators must be placed *after* the 'export' keyword. Remove the 'decoratorsBeforeExport: false' option to use the '@decorator export class {}' syntax. (27:0)

>  26 | @customElement('affine-reference')
 27 | export class AffineReference extends WithDisposable(ShadowlessElement) {
  28 |   static override styles = css`
  29 |     .affine-reference {
  30 |       white-space: nowrap;
When I delete updateBlockType, playwright works fine. And I didn't find decoratorsBeforeExport option in our tsconfig.json, so how to resolve this problem?

doodlewind:
Playwright is an E2E testing tool, generally you only trigger keyboard and mouse events against a web page, rather than importing a method and run it. So I'm not sure it's best practice to import `updateBlockType` (and all other non `import type` imports of `@blocksuite/blocks`) in playwright test case. Instead, it may be easier to press menus and perform mouse selections inside the test case.

smallnine9:
Apologies for the delayed response. I have submitted a new test based on the suggestions from Yifeng. Your reviews and feedback are welcome.

doodlewind:
@smallnine9 could you please take a look at the test case regression (https://github.com/toeverything/blocksuite/actions/runs/5472630294/jobs/9965048544?pr=3158 and https://github.com/toeverything/blocksuite/actions/runs/5472630294/jobs/9965048292?pr=3158)? Should not be hard to get them reproduced locally.

smallnine9:
ok

doodlewind (approved):

doodlewind:
Thanks for passing the test, we should land this long awaiting fix!

PR #3161, opened 2023-06-19
feat(edgeless): keep brush state in one session

### LINKED ISSUE #2649: Hope that the color, thickness, and other settings of the previous Pen can be remembered
The settings should be recorded locally and should be recorded separately for different pages.

regischen:
closes #2649


doodlewind (commented):
How about migrating the viewport state to session storage altogether?

regischen:
> How about migrating the viewport state to session storage altogether?

okay

doodlewind (approved):

PR #3166, opened 2023-06-19
refactor: use transform instead of position on selected-blocks

fundon:


doodlewind (commented):
Does this benefit the performance?

fundon:
https://stackoverflow.com/questions/7108941/css-transform-vs-position
https://www.paulirish.com/2012/why-moving-elements-with-translate-is-better-than-posabs-topleft/

doodlewind:
This makes sense, thanks for the perf improvement!

doodlewind (approved):

PR #3175, opened 2023-06-20
feat(block-std): add block spec

Saul-Mirone:


doodlewind (approved):
LGTM! we can update the docs afterwards.

PR #3179, opened 2023-06-20
build: release package `block-std`

himself65:

PR #3180, opened 2023-06-20
chore: update publish script

Saul-Mirone:


himself65:
see https://github.com/toeverything/blocksuite/pull/3179

PR #3184, opened 2023-06-21
feat(edgeless): connector

### LINKED ISSUE #2654: Refactor connector

regischen:
close #2654 

* [x] straight connector auto path
* [x] straight connector anchor path
* [x] elbow connector auto path
* [x] elbow connector anchor path  
* [x] indicator
* [x] sync update when other element dimension changes
* [x] tests 
* [x] remove connector package 
straight line auto
anchoring straight
auto elbow


packages/blocks/src/page-block/edgeless/connector-manager.ts
```diff
@@ -0,0 +1,919 @@
...
+
+export class EdgelessConnectorManager {
+  private _algorithm: AStarAlgorithm | undefined;
+  private _connectionOverlay = new ConnectionOverlay();
+  constructor(private _edgeless: EdgelessPageBlockComponent) {
+    this._edgeless.surface.viewport.addOverlay(this._connectionOverlay);
+    this._connectionOverlay.surface = this._edgeless.surface;
+  }
+
+  searchConnection(point: IVec, excludedIds: string[] = []) {
```
doodlewind, line 78: Please add necessary blank lines or split sub methods to improve the readability of this method.

packages/phasor/src/renderer.ts
```diff
@@ -69,6 +72,10 @@ export class Renderer implements SurfaceViewport {
     this.ctx = this.canvas.getContext('2d') as CanvasRenderingContext2D;
     this.rc = new RoughCanvas(canvas);
   }
+  clientToModelCoord(vec: IVec): IVec {
```
doodlewind, line 75: This method lacks consistency with existing ones (`toViewCoord` / `toModelCoord`). Having the caller side compute the bounding rect makes more sense. Please DO care about consistency, otherwise you'll break the codebase in a very fast way, which is NOT acceptable as our official team member.

Also, notice that DO NOT write newly added methods in such arbitrary position, they should be placed relative to the ones with close semantics.

packages/blocks/src/page-block/edgeless/connector-manager.ts
```diff
@@ -0,0 +1,944 @@
...
+
+  private filterConnectablePoints(points: IVec[], bound: Bound | null) {
+    return points.filter(point => {
+      if (!bound) return true;
+      return !bound.isPointInBound(point);
+    });
+  }
+
+  private getConnectablePoints(
+    osp: IVec,
```
doodlewind, line 510: Do not use overly simplified variable naming, or at lease add comment on their naming. we are NOT writing C.

doodlewind: This method doesn't use any `this` properties so it should be extracted as standalone utils. We are NOT writing Java.

regischen: tomorrow, I will continue improve this method

packages/blocks/src/page-block/edgeless/tool-controllers/connector-tool.ts
```diff
@@ -1,39 +1,20 @@
...
-
-  private _pickBy(
-    x: number,
-    y: number,
-    filter: (element: Selectable) => boolean
-  ) {
-    const { surface } = this._edgeless;
-    return pickBy(surface, this._page, x, y, filter);
-  }
+  private _connector: ConnectorElement | undefined = undefined;
```
doodlewind, line 15: We don't recommend using explicit `undefined` literals, this simply looks weird. Or else use `_connector?` as an alternative.

regischen: ok, I'll notice that in the following

doodlewind (approved):

Issue #2667, opened 2023-05-19
Add link field for Database Block


joebeijing:
- Automatically recognize any content entered as a link by default, and keep the Text and Link consistent. 
- Display the 'Edit' button as soon as the mouse hovers over the link. 
- Clicking on the link will directly redirect to the link.

PR #3187, opened 2023-06-22
feat(database): support `link` column type

zqran:
close: #2667

https://github.com/toeverything/blocksuite/assets/15389209/b77cd85f-90ca-4db1-b4b6-1cf5c359adec


packages/blocks/src/database-block/table/components/column-type/link.ts
```diff
@@ -0,0 +1,271 @@
...
+        value = this._initYText(this.value);
+      } else {
+        value = this.value;
+      }
+    }
+
+    this.vEditor = new VEditor(value, {
+      active: () => activeEditorManager.isActive(this),
+    });
+    setupVirgoScroll(this.column.page, this.vEditor);
```
zzj3720, line 110: Can we not depend on Virgo here

packages/blocks/src/database-block/table/components/selection/selection.ts
```diff
@@ -96,11 +96,12 @@ export class DatabaseSelectionView extends WithDisposable(ShadowlessElement) {
...
     this._disposables.add({
       dispose: this.eventDispatcher.add('click', context => {
-        const target = context.get('pointerState').event.target;
+        const event = context.get('pointerState').event;
+        const target = event.target;
         if (target instanceof Element && this.isCurrentDatabase(target)) {
           const cell = target.closest('affine-database-cell-container');
           if (cell) {
-            cell?.cell?.enterEditMode();
+            cell?.cell?.enterEditMode(event);
```
zzj3720, line 104: enterEditMode may be called without an `event` in future, Can we not depend an `event` here

zqran: Here I want to trigger the click of the link when clicking the external area of the link, so I need to get the relevant elements.

image

---

Related code:

https://github.com/toeverything/blocksuite/pull/3187/files#diff-7f1b6d33e4075b6a3dc56f7a12e6d0bfca7122157175c37bfe3eaad36c347ec7R127-R135

zqran: Get!

packages/blocks/src/database-block/table/components/column-type/link.ts
zqran: ok

PR #3212, opened 2023-06-26
feat(edgeless): support select element below the hollow shape

### LINKED ISSUE #2034: Hope to be able to select other components below the stroke
The expected result is that the suggested region for selecting strokes should be a fixed margin around the stroke rather than the entire rectangular area, so that other components can be selected within the stroke area as well as non-stroke parts of the area.

CleanShot 2023-04-12 at 11 55 31@2x

cc @HeJiachen-PM  @doodlewind

doouding:
close #2034 

Sorry for the long wait. I spent some time investigating several open source projects to find a proper solution. However, I found that all of them require a lot of effort. I believe that we need an overall solution for this type of geometric problem (just like how excalidraw do it), but it may not be our top priority right now. So I just used some simple math and native APIs to fulfill our needs.


doodlewind:
Just saying that you don't have to add module and type labels to PRs, they are mainly for issues :)

doouding:
It seems break some tests. I'll fix them soon.

doodlewind (approved):
Great to land this, thanks!

PR #3216, opened 2023-06-26
fix: support copy image to other app

### LINKED ISSUE #3056: Support copy image to word
Now we have not implemented copying the image to the clipboard and paste it into word or excel.

zuoxiaodong0815:
close #3056


packages/blocks/src/__internal__/clipboard/utils/pure.ts
```diff
@@ -102,6 +106,25 @@ export function performNativeCopy(items: ClipboardItem[]): boolean {
   return success;
 }
 
+function _copyImage(imageSrc: string): boolean {
```
doodlewind, line 109: We don't use `_foo` for module-level functions, only `this._foo` for private class methods and properties.

packages/blocks/src/__internal__/clipboard/utils/pure.ts
```diff
@@ -102,6 +106,25 @@ export function performNativeCopy(items: ClipboardItem[]): boolean {
...
+    range.selectNode(tempImageElem);
+    selection?.removeAllRanges();
+    selection?.addRange(range);
+    try {
+      document.execCommand('copy');
+    } finally {
+      document.body.removeChild(tempImageElem);
+    }
+  };
+  document.body.appendChild(tempImageElem);
```
lawvs, line 124: Please hide the image element before appending it to the body

PR #3261, opened 2023-06-29
feat(block-std): scaffolding new selection manager

Saul-Mirone:


packages/block-std/src/selection/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './base.js';
+export * from './manager.js';
+export * from './variant/index.js';
```
doodlewind, line 3: `variant` -> `variants` :D

doodlewind (approved):

packages/block-std/src/selection/index.ts
Saul-Mirone: Renamed.

Issue #3246, opened 2023-06-29
In Edgeless mode, the Bookmark menu does not respond or display completely


joebeijing:
<img width="1003" alt="CleanShot 2023-06-29 at 14 49 45@2x" src="https://github.com/toeverything/blocksuite/assets/22443345/4b61b49a-21d5-410d-bcd6-e1dd6a9918f1">

@JackiePengp mentioned that clicking "More" doesn't respond. I couldn't reproduce it, but after clicking, it doesn't display completely. It shouldn't be covered by the boundary of the Note.

doodlewind:
/cc @QiShaoXuan since it works for image:

<img width="874" alt="image" src="https://github.com/toeverything/blocksuite/assets/7312949/12560465-cd1c-4805-aa77-f6354aa4dd37">

PR #3267, opened 2023-06-30
fix: bookmark toolbar hidden by the frame in edgeless mode

QiShaoXuan:
fixed: https://github.com/toeverything/blocksuite/issues/3246

### DIFF
--- a/packages/blocks/src/bookmark-block/bookmark-block.ts
+++ b/packages/blocks/src/bookmark-block/bookmark-block.ts
@@ -224,8 +224,9 @@ export class BookmarkBlockComponent extends BlockElement<BookmarkBlockModel> {
       }
 
       this._toolbarHoverState = {
-        inToolbar: inToolbar || currentToolbarState,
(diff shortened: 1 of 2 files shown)


packages/blocks/src/bookmark-block/bookmark-block.ts
```diff
@@ -266,13 +297,22 @@ export class BookmarkBlockComponent extends BlockElement<BookmarkBlockModel> {
...
+    const toolbar = showToolbar
       ? html`<bookmark-toolbar
           .model=${this.model}
           .onSelected=${this._onToolbarSelected}
+          .root=${this}
+          .toolbarHoverStateSlot=${this._toolbarHoverStateSlot}
         ></bookmark-toolbar>`
       : nothing;
 
+    // const toolbar = html`<bookmark-toolbar
```
doodlewind, line 309: Should be removed here

doodlewind (approved):

PR #3273, opened 2023-06-30
fix: missing title when exporting markdown

### LINKED ISSUE #3247: when export the file with the markdown format, and import it again, it fails to do so and the imported file's title is not consistent with the original one
https://github.com/toeverything/blocksuite/assets/22443345/d313454b-44ec-495e-9b73-f16490955c6e

zuoxiaodong0815:
close #3247 

fix: missing title when exporting markdown
fix: export only one text node if BlockSuite has more than one text node.


doodlewind (approved):
Thanks for the fix!

Issue #2601, opened 2023-05-17
Incorrect opened link URL


doodlewind:
<img width="547" alt="image" src="https://github.com/toeverything/blocksuite/assets/7312949/4f47db12-0689-440a-919e-c7ad3dec2b28">

QiShaoXuan:
Note: in the fixed pr, only prepend  https protocol when bookmark click, but not reloading

PR #3277, opened 2023-06-30
fix: prepend "https" if url without protocol when bookmark click to open

QiShaoXuan:
fixed: https://github.com/toeverything/blocksuite/issues/2601


doodlewind (approved):
Thanks for maintaining this one!

PR #3281, opened 2023-06-30
feat: update edgeless toolbar ui and zoom bar position

donteatfriedrice:
update edgeless toolbar ui and zoom bar position


doodlewind (approved):
Thanks for the timely fix!

PR #3285, opened 2023-07-02
fix(edgeless): optimize elbow connector path structure

regischen:
Optimize some elbow path

before
6x8F2nJflG
after
img_v2_bbdbaf75-82c8-4a74-86d5-82859ff9803g

before
image

after
SU2NauTQHu

### DIFF
--- a/packages/phasor/src/utils/a-star.ts
+++ b/packages/phasor/src/utils/a-star.ts
@@ -7,16 +7,16 @@ import { PriorityQueue } from './priority-queue.js';
 import { type IVec } from './vec.js';
 
 export class AStarRunner {
-  private _cameFrom = new Map<IVec, IVec | null>();
+  private _cameFrom = new Map<IVec, { from: IVec[]; indexs: number[] }>();
   private _frontier!: PriorityQueue<
     IVec,
     [diagonalCount: number, pointPriority: number, distCost: number]
   >;
 
   private _graph: Graph;
-  private _costSoFar = new Map<IVec, number>();
-  private _diagonalCount = new Map<IVec, number>();
-  private _pointPriority = new Map<IVec, number>();
+  private _costSoFar = new Map<IVec, number[]>();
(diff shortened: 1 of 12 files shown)


packages/phasor/src/utils/a-star.ts
```diff
@@ -32,48 +32,60 @@ export class AStarRunner {
...
     if (almostEqual(a[1], last[1]) && almostEqual(a[1], last2[1])) return 0;
     return 1;
   }
 
   private _almostEqual(a: IVec, b: IVec): boolean {
-    return almostEqual(a[0], b[0]) && almostEqual(a[1], b[1]);
+    return almostEqual(a[0], b[0], 0.02) && almostEqual(a[1], b[1], 0.02);
+  }
+
+  private _neighbors(cur: IVec) {
```
doodlewind, line 76: These non-`this` dependent methods should be top-level utils, including `_compare`, `_heuristic` and more. Please try making less inconsistencies that has been addressed in former review.

regischen: ok，I'll put more attention on it

tests/edgeless/connector.spec.ts
```diff
@@ -396,36 +336,75 @@ test('path #1', async ({ page }) => {
...
+    [250, 0],
   ]);
 });
 
-test('path #2', async ({ page }) => {
-  await enterPlaygroundRoom(page);
-  await initEmptyEdgelessState(page);
-  await switchEditorMode(page);
-  await deleteAll(page);
+test('path #2, the top-right point is overlapped with the bottom-left point of another, and anchor from top to bottom of another', async ({
```
doodlewind, line 366: Please migrate all test cases in this file according to 702bc472200dd0b587016026a489982737999b9f. Current writeup is too verbose. Also this kind of reassigning `let start, end` is unnecessary and should be avoided when possible.

doodlewind (approved):

PR #3310, opened 2023-07-04
feat: hide toolbar when drag element move(#3271)

### LINKED ISSUE #3271: Hide component toolbar when drag element move
The component toolbar is a bit distracting
hide toolbar

Yukiniro:
Closes #3271


doodlewind (approved):
Thanks for the trying!

PR #3315, opened 2023-07-04
Initial Dockerfile and docker-compose.yml for Docker support

as-shura:
Is it possible to review this in a way that we can continue using this inside of Docker?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,27 @@
+# Multi-stage
+# 1) Node image for building frontend assets
+# 2) Nginx stage to serve frontend assets
+
+# Name the node stage "builder"
+FROM node:18 AS builder
+# Install pnpm
+RUN npm install --global pnpm
+# Set working directory
+WORKDIR /app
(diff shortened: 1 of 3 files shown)


himself65 (commented):
/cc @darkskygit

doodlewind:
I don't think the BlockSuite framework and the playground is an app to be reused via Docker. We will keep using the native tool chain for building and testing it. But we have planned docker support for http://affine.pro and this will be landed very soon tho.

doodlewind:
Or else, could you elaborate your use case?

doodlewind:
It's recommended to use [StackBlitz](https://stackblitz.com/github/toeverything/blocksuite) and [CodeSandbox](https://codesandbox.io/p/github/toeverything/blocksuite/master) for hosting cross-platform debugging environment (and we are open to PR that reflects this in our BUILDING.md docs), and this one should be closed since for BlockSuite, we don't have further plan (including docs and regression maintenance) on docker support.

PR #3318, opened 2023-07-04
feat: supports copying images and cross app pasting

fundon:
closes: #1468,  #3056


packages/blocks/src/page-block/default/utils.ts
```diff
@@ -329,65 +327,6 @@ function getBlockAndRect(blocks: BaseBlockModel[], mid: number) {
   };
 }
 
-export async function downloadImage(model: BaseBlockModel) {
```
doodlewind, line 332: Thanks for help migrating legacy utils!

doodlewind (approved):

PR #3322, opened 2023-07-04
fix: todo block display error after importing notion

### LINKED ISSUE #3304: When importing the Notion data to AFFiNE it did not show the same results
img_v2_63fabb56-4c6d-4cb3-b643-ce41655bb89g
ecbb9799-e833-4155-9a2e-287fea126f46

zuoxiaodong0815:
close #3304


fourdim:
I insist on not to fix that, as it will break many things.

zuoxiaodong0815:
We consider whether to fix a problem first in terms of product experience, then in terms of technical cost.

packages/blocks/src/__internal__/content-parser/index.ts
```diff
@@ -354,7 +354,47 @@ export class ContentParser {
         return `<code>${token.text}</code>`;
       },
     };
-    marked.use({ extensions: [underline, inlineCode] });
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
doodlewind, line 358: Should define the interface of `token`

zuoxiaodong0815: I have fixed this problem.

doodlewind (approved):

Issue #3320, opened 2023-07-04
Unable to copy edgeless content as PNG in playground


doodlewind:
https://github.com/toeverything/blocksuite/assets/7312949/3b7dee29-73e8-4efd-a3c3-3f2b53165fb2

Console error:

```
block-hub.ts:556 Uncaught TypeError: Cannot read properties of undefined (reading 'page')
    at new BlockHub (block-hub.ts:556:26)
    at DocumentCloner2.createElementClone (document-cloner.ts:156:28)
    at DocumentCloner2.cloneNode (document-cloner.ts:318:32)
    at DocumentCloner2.appendChildNode (document-cloner.ts:284:40)
    at DocumentCloner2.cloneChildNodes (document-cloner.ts:301:22)
    at DocumentCloner2.cloneNode (document-cloner.ts:340:22)
    at DocumentCloner2.appendChildNode (document-cloner.ts:284:40)
    at DocumentCloner2.cloneChildNodes (document-cloner.ts:301:22)
    at DocumentCloner2.cloneNode (document-cloner.ts:340:22)
    at new DocumentCloner2 (document-cloner.ts:69:37)
```

/cc @zuoxiaodong0815 to take a look or assign.

zuoxiaodong0815:
<img width="560" alt="image" src="https://github.com/toeverything/blocksuite/assets/53252747/f69429da-4808-4775-8409-8d8fae81319f">
@fundon Need to implement this method 'ignoreElements'.  Please modify according to this code.

zuoxiaodong0815:
The backgroundColor is needed to deal.

PR #3331, opened 2023-07-04
fix: console error when copying as image

fundon:
closes: #3320
related to: https://github.com/toeverything/AFFiNE/pull/3005

* [x] fix: throw an error
* [x] fix: update clipboard api on electron


doodlewind (approved):
This is fast, thanks for the fix!

PR #3335, opened 2023-07-04
feat(edgeless): add overlay for note and shape tool

donteatfriedrice:
- [ ] Optimize edge cases

### DIFF
--- a/packages/blocks/src/__internal__/utils/types.ts
+++ b/packages/blocks/src/__internal__/utils/types.ts
@@ -185,6 +185,7 @@ export type NoteTool = {
   background: CssVariableName;
   childFlavour: NoteChildrenFlavour;
   childType: string | null;
+  tip: string;
 };
 
 export type ConnectorTool = {
--- a/packages/blocks/src/page-block/edgeless/components/toolbar/note/note-menu.ts
+++ b/packages/blocks/src/page-block/edgeless/components/toolbar/note/note-menu.ts
@@ -124,7 +124,8 @@ export class EdgelessNoteMenu extends LitElement {
 
   private _updateNoteTool(
     childFlavour: NoteChildrenFlavour,
-    childType: string | null
+    childType: string | null,
+    tip: string
   ) {
     if (this.edgelessTool.type !== 'note') return;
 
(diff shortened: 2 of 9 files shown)


regischen:
The positon of the overlay of the shape other than rect didn't overlap with the real shape when we click the mouse.
This may improve a little.

https://github.com/toeverything/blocksuite/assets/58546692/438c7439-fbcb-421f-8294-95f264c943ea

regischen:
It would be great if the overlay's position of note is same with the real created note

donteatfriedrice:
> The positon of the overlay of the shape other than rect didn't overlap with the real shape when we click the mouse. This may improve a little.
> 
>  Screen.Recording.2023-07-05.at.16.20.05.mov

I will improve this. Now I have a question, I should modify the overlay style to rough, how can I achieve this? Should I modify the Overlay abstract class , add a override render method with RoughCanvas for it? Then when I implement the draw method, use the Rough style?

regischen:
> > The positon of the overlay of the shape other than rect didn't overlap with the real shape when we click the mouse. This may improve a little.
> > Screen.Recording.2023-07-05.at.16.20.05.mov
> 
> I will improve this. Now I have a question, I should modify the overlay style to rough, how can I achieve this? Should I modify the Overlay abstract class , add a override render method with RoughCanvas for it? Then when I implement the draw method, use the Rough style?

image
this will meet your expectation

donteatfriedrice:
> > > The positon of the overlay of the shape other than rect didn't overlap with the real shape when we click the mouse. This may improve a little.
> > > Screen.Recording.2023-07-05.at.16.20.05.mov
> > 
> > 
> > I will improve this. Now I have a question, I should modify the overlay style to rough, how can I achieve this? Should I modify the Overlay abstract class , add a override render method with RoughCanvas for it? Then when I implement the draw method, use the Rough style?
> 
> image this will meet your expectation

Ok, I will try. Thanks!!!

doodlewind (approved):
Several quick feedback:

* Incorrect overlay position on panning edgeless viewport (see the recording).
* Besides dragging, clicking on canvas should also add new shape element.
* Using the `RoughCanvas` style overlay for shape will be better. Also, it may also be good for the shape overlay to inherit the last shape element style edited, so that we can keeping adding "blue shape with red fill color".
* Minor padding issue for the text overlay.

https://github.com/toeverything/blocksuite/assets/7312949/6983b8db-22dc-4989-8af9-3be771fae9b2

image

But the current implementation is already awesome, let's get this landed!

Issue #3332, opened 2023-07-04
Pressing enter in database tag input doesn't flush tag input


hwangdev97:
https://github.com/toeverything/blocksuite/assets/24713927/a0c18999-2eec-44ed-be70-d25662ea1ef0

Just like it was shown in the video,  after pressing enter, it appears that two tags were created and some suggested words are left behind.

doodlewind:
On pressing enter, adding a new tag named `c p ` makes sense (since before confirming in IME, that's what you actually input in the input box). But the input box should be cleared for new tag content at the moment.

zzj3720:
It looks like the `select` action should not be triggered during the `CompositionEvent`

PR #3336, opened 2023-07-05
fix(database): prevent tag from being selected when compsing

zqran:
close: #3332

https://github.com/toeverything/blocksuite/assets/15389209/cee35cba-5d0c-48ec-90b0-62252f3d84a5


doodlewind (approved):
Thanks for the timely fix!

PR #3394, opened 2023-07-06
feat(edgeless): support connecting rotated elements

regischen:
support connector connects rotated elements.
And considering slope.

https://github.com/toeverything/blocksuite/assets/58546692/75198127-4773-4b45-abae-830740c7a53c

Also support triangle and diamond.

https://github.com/toeverything/blocksuite/assets/58546692/81533d3c-33c2-4df5-ac65-dc0fa9f34b65

TODO: support connects ellipsis edges. now only can connect anchors.


packages/phasor/src/utils/point-location.ts
```diff
@@ -0,0 +1,28 @@
+import { type IVec } from './vec.js';
+
+export class PointLocation extends Array<number> implements IVec {
```
doodlewind, line 3: Not sure if this is always used as marking relative position? If true, this can be better named as `RelativeLocation`.

doodlewind (approved):

packages/phasor/src/utils/point-location.ts
regischen: the point in PointLocation is stiil canvas point, it's not related to something.
But it brings the tangent if the point is on outline of some surfaceElements.

and maybe in the future, it will bring more infomation.

Issue #3346, opened 2023-07-05
Format bar should not expose card view conversion button inside database title column


doodlewind:
inner feedback row id: 10

<img width="558" alt="image" src="https://github.com/toeverything/blocksuite/assets/7312949/7fce8ebd-ebae-4dbb-9f37-f40a565d6ef5">

We can hide the `Turn into card view` inside title column.

PR #3397, opened 2023-07-06
fix: hide turn card btn when the link inside database

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/3346


doodlewind (approved):
Thanks for the fix!

PR #3408, opened 2023-07-06
feat(block-std): add text selection in selection manager

Saul-Mirone:


packages/blocks/src/services.ts
```diff
@@ -2,20 +2,20 @@ import type { UnionToIntersection } from '@blocksuite/global/types';
...
-import { ParagraphBlockService } from './paragraph-block/index.js';
+import { LegacyParagraphBlockService } from './paragraph-block/index.js';
 
 export const blockService = {
-  'affine:code': CodeBlockService,
+  'affine:code': LegacyCodeBlockService,
   'affine:database': DatabaseBlockService,
-  'affine:paragraph': ParagraphBlockService,
-  'affine:list': ListBlockService,
+  'affine:paragraph': LegacyParagraphBlockService,
```
doodlewind, line 17: What's the difference between legacy and non-legacy providers?

Saul-Mirone: The legacy service is not part of block spec and will be replaced by the new service in a progressive way.

packages/block-std/package.json
```diff
@@ -16,7 +16,8 @@
     "@blocksuite/store": "workspace:*"
   },
   "dependencies": {
-    "@blocksuite/global": "workspace:*"
+    "@blocksuite/global": "workspace:*",
+    "lib0": "^0.2.74"
```
doodlewind, line 20: Not sure if this will install lib0 twice? We even use `peerDeps` to make sure yjs only installs once.

PR #3423, opened 2023-07-07
feat(edgeless): support adding text in shape

### LINKED ISSUE #2441: Support adding text above the shape

Flrande:
Close #2441

https://github.com/toeverything/blocksuite/assets/50035259/36a8a143-1b99-4f80-a506-6a5794037be8


packages/blocks/src/page-block/edgeless/components/text/edgeless-shape-editor.ts
```diff
@@ -0,0 +1,192 @@
...
+import { customElement, query } from 'lit/decorators.js';
+import { styleMap } from 'lit/directives/style-map.js';
+import * as Y from 'yjs';
+
+import { isCssVariable } from '../../../../__internal__/theme/css-variables.js';
+import type { EdgelessPageBlockComponent } from '../../edgeless-page-block.js';
+import { getSelectedRect } from '../../utils/query.js';
+
+@customElement('edgeless-shape-editor')
+export class EdgelessShapeEditor extends WithDisposable(ShadowlessElement) {
```
doodlewind, line 16: -> `EdgelessShapeTextEditor`

packages/phasor/src/elements/shape/types.ts
```diff
@@ -19,6 +26,17 @@ export interface IShape extends ISurfaceElement {
...
+  text?: Y.Text;
+  color?: string;
+  fontSize?: SHAPE_TEXT_FONT_SIZE;
+  fontFamily?: string;
+  textHorizontalAlign?: 'left' | 'center' | 'right';
+  textVerticalAlign?: 'top' | 'center' | 'bottom';
+}
+
+export interface IShapeLocalRecord extends ISurfaceElementLocalRecord {
+  textDisplay?: boolean;
```
doodlewind, line 39: `displayText`

Should also comment on when is this required.

doodlewind (approved):
Let's land this, thanks for the initial effort in shape text!

PR #3424, opened 2023-07-07
fix(edgeless): unexpected history record added when updating note height

### LINKED ISSUE #3385: Unable to redo after cliping
To reproduce:

1. Clip note
2. Undo (things work fine by the moment)
3. Try to redo but the history stack will be cleared

https://github.com/toeverything/blocksuite/assets/7312949/91e0448e-981d-4930-9fba-26d185bbb01b

This should be related to redundant state update after undo, so the root cause should also be #2475

doouding:
close #3385

### Idea behined the solution
Note height is updated every time its related DOM element's rect changes. However, we cannot tell what triggers the change in the DOM rect (it could be triggered by local updates or undo/redo operations). Therefore, I have decided to make the undoManager not track updates to the note height.
If you think this solution is feasible, then I'll add some unit tests in the next days. Please let me know if you have any suggestions, thanks.

### DIFF
--- a/packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
+++ b/packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
@@ -33,6 +33,7 @@ import { styleMap } from 'lit/directives/style-map.js';
 
 import { EdgelessClipboard } from '../../__internal__/clipboard/index.js';
 import {
+  almostEqual,
   asyncFocusRichText,
   type BlockComponentElement,
   bringForward,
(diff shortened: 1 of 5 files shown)


packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
```diff
@@ -1062,10 +1065,35 @@ export class EdgelessPageBlockComponent
...
+    this._disposables.add(
+      this.page.slots.rootAdded.on(root => listenChildrenUpdate(root))
+    );
+  }
+
   override firstUpdated() {
     this._initSlotEffects();
     this._initDragHandle();
     this._initResizeEffect();
+    this._initNoteResize();
```
doodlewind, line 1096: This is duplicate with the `_initResizeEffect` above. Can you try to get these 2 methods merged?

doodlewind:
This is a recommended pattern that should work! I would try to advocate this in future watch-based model mutations afterwards, so that these effects won't pollute the history stack.

packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
doouding: Technically those two methods are not same. The name maybe a little confusion, I'll rename it.

doodlewind (approved):

doodlewind (approved):

PR #3425, opened 2023-07-08
fix(edgeless): add size limit for grid gap

### LINKED ISSUE #2651: The Grid on the whiteboard should not become very dense dots after scaling
CleanShot 2023-05-18 at 5 16 07@2x

Yukiniro:
Closes #2651 
Add `GRID_GAP_MIN` and `GRID_GAP_MAX` to limit the value of grid gap


doodlewind (approved):
Thanks for looking into this!

PR #3439, opened 2023-07-10
refactor(edgeless): remove redundant callback param

fundon:


doodlewind (approved):
Thanks for the refactoring, looking for more cleanups like this!

PR #3454, opened 2023-07-11
fix(edgeless): position of selected-rect when resizing after rotation

fundon:
before

https://github.com/toeverything/blocksuite/assets/27926/bead8233-8035-4286-8fb4-d1d2ecb6f8ed

after

https://github.com/toeverything/blocksuite/assets/27926/53194b49-051f-45c6-9980-e15adc5ae144


doodlewind (approved):

doodlewind:
@fundon is it possible to pass the failing test? I'd really appreciate if we can land this as a great cleanup.

fundon:
I think #3665 would be a better implementation.

PR #3460, opened 2023-07-11
fix(edgeless): scissors z-index bug after bump into AFFiNE

doouding:
https://github.com/toeverything/blocksuite/assets/9301743/04382761-3825-4c36-bde3-7ad5dad81443


packages/blocks/src/page-block/edgeless/components/note-cut/cut-button.ts
```diff
@@ -139,7 +139,7 @@ export class NoteScissorsButton extends WithDisposable(LitElement) {
         position: absolute;
         left: 0;
         top: 0;
-        z-index: 1;
+        z-index: calc(var(--affine-z-index-popover, 0) + 2);
```
doodlewind, line 142: Is it possible to define this `calc` result as a new CSS variable like `--affine-z-index-edgeless-scissor`?

PR #3465, opened 2023-07-11
feat(edgeless): complete canvas text component toolbar

donteatfriedrice:


donteatfriedrice:
It seems that the selected rectangle would not resize when font family change, should fix it.

doodlewind (approved):
LGTM

PR #3475, opened 2023-07-11
fix(edgeless): change image caption background to transparent

### LINKED ISSUE #3472: Caption background in Edgeless mode
### What happened

When adding the caption for an image, the caption background should be transparence

Image

donteatfriedrice:
To close #3472 

Screenshot 2023-07-12 at 00 30 51


doodlewind (approved):
Thanks!

PR #3526, opened 2023-07-13
fix: don't forcefully change list type

### LINKED ISSUE #3468: Add number prefix to a todo item should not forcefully change it into numbered list, vice versa
https://github.com/toeverything/blocksuite/assets/132558/45243302-280c-4de4-b799-a1407ac73249

smallnine9:
fix #3468
Thanks for your review ：）


doodlewind (approved):
Thanks for the timely fix!

PR #3558, opened 2023-07-15
refactor(page): use slot instead of dispatchEvent

AliceLanniste:


packages/blocks/src/code-block/code-block.ts
```diff
@@ -266,6 +266,15 @@ export class CodeBlockComponent extends BlockElement<CodeBlockModel> {
   }
 
   override updated() {
+    const langList = this.querySelector('lang-list');
```
doodlewind, line 269: How about using `@query` (https://lit.dev/docs/components/shadow-dom/#query) here?

PR #3584, opened 2023-07-17
feat(virgo): ctrl delete to delete one word forward

### LINKED ISSUE #3566: Support CTRL + Del to delete one word forward
### Description

Delete one word to the right like in any other program.

### Use case

I rely on this keybind a lot.

### Anything else?

https://defkey.com/what-means/ctrl-del

lzlme:
resolve #3566


Flrande (changes requested):
We should handle native [inputType](https://www.w3.org/TR/input-events-2/#interface-InputEvent-Attributes) in [transform-input](https://github.com/toeverything/blocksuite/blob/75669fcb78d7f8f11132d9245b5030504d006cba/packages/virgo/src/utils/transform-input.ts#L154).

lzlme:
> We should handle native [inputType](https://www.w3.org/TR/input-events-2/#interface-InputEvent-Attributes) in [transform-input](https://github.com/toeverything/blocksuite/blob/75669fcb78d7f8f11132d9245b5030504d006cba/packages/virgo/src/utils/transform-input.ts#L154).

Well, what is a good way to let `transformInput` know that `ctrl-delete` is pressed? Should `_ctrlDelete` be passed as a parameter to `transformInput`? And then handle it like this ?
```javascript
case 'deleteContentBackward':
case 'deleteByCut': {
  if (ctrlDelete) {
    handleWordDelete(editor, vRange);
    return
  }
  handleDelete(vRange, editor);
  return;
}
```

Flrande:
> Well, what is a good way to let transformInput know that ctrl-delete is pressed? Should _ctrlDelete be passed as a parameter to transformInput? And then handle it like this ?

If it is a native behavior you don't need to know `ctrl-delete` and you can judge it using `deleteWordForward` (see it in [inputType](https://www.w3.org/TR/input-events-2/#interface-InputEvent-Attributes))

It it is not a native behavior you should not implement it in virgo and you should implement it in [defineKeyMap](https://github.com/toeverything/blocksuite/blob/75669fcb78d7f8f11132d9245b5030504d006cba/packages/blocks/src/__internal__/service/index.ts#L157).

lawvs (commented):
Could you add some tests for it?

lzlme:
> Could you add some tests for it?

Sure

Flrande (approved):

PR #3599, opened 2023-07-18
fix(edgeless): the newly created block should retain its formatting

### LINKED ISSUE #3578: Support retain formatting for block slicer
When using Block Splitter the newly created block will not retain the formatting - such as background colour.

https://github.com/toeverything/AFFiNE-Community/assets/4605025/fc61ed76-114f-4ff4-9c25-5a2d8a29c871

doouding:
close #3578


doodlewind (approved):
This is quick, thanks!

PR #3602, opened 2023-07-18
perf: lazy provider for loading subdocs

pengx17:
related https://github.com/toeverything/AFFiNE/issues/3029
Add an abstraction for provider to support lazy loading sub docs in blocksuite

----

#### RFC
In my current design, the user is responsible for binding listeners to a document with a data source in a provider. However, after internal discussions, we realized that this approach may not work when a subdocument is used within another page, which goes beyond our observability.

Since our provider implementation only has access to y.doc and not blocksuite abstractions such as Page or Workspace, we will continue utilizing y.doc as the bridge between the document and datasource.

When invoking `subdoc.load`, events related to subdocuments (load, added, removed) will be emitted at the parent doc. However, this level of granularity is too broad for our needs and requires alternative solutions.

The key aspect of the lazy load provider is maintaining a reference count for "connected" and "disconnected" events on a y.doc by calling `waitForLoaded` and `unload` (to be determined) on Page/Space objects.


pengx17:
Close due to no need to make change to blocksuite for now

Issue #3368, opened 2023-07-05
Support resizing shape element on edge


doodlewind:
The note block supports resizing on edge, would be great if this applies to the shape element, both horizontally and vertically:

https://github.com/toeverything/blocksuite/assets/7312949/9bc8d9dd-dc52-4bd2-a63c-ad69d8934216

The note block is not expected to be resized vertically, but this is not the case for shape elements.

For our initial implementation, resizing on rotate shapes can be ignored if this requires too much complexity.

doodlewind:
Closed by #3604

PR #3604, opened 2023-07-18
feat(edgeless): support resizing on edge the shape element

Yukiniro:
Related to #3368


packages/phasor/src/utils/math-utils.ts
```diff
@@ -499,3 +499,21 @@ export function isOverlap(
     )
   );
 }
+
+export function calcPointToPointWithAngle(
```
regischen, line 503: there already has `rotatePoints` method in math-utils. you can use it to avoid duplication

Yukiniro: Ok. I will update this PR later

packages/blocks/src/page-block/edgeless/components/resize/resize-manager.ts
```diff
@@ -210,6 +254,38 @@ export class HandleResizeManager {
       dp.x += deltaX;
       dp.y += deltaY;
 
+      if (
```
doodlewind, line 257: Thanks for the contribution! But here I think "resizing on edge" refers more to following effect:

https://github.com/toeverything/blocksuite/assets/7312949/89554cbb-567c-4914-ad60-ba5d3c88cd6f

With draggable edges, we don't have to add more visible handles :)

Yukiniro: By the way, can I understand that the edge can be dragged in horizontal and vertical directions, but the circular handles are not displayed on the edge?

doodlewind: Yep!

doodlewind (approved):
Found a minor regression that the bottom right cursor should be `nwse-resize` instead of `ew-resize`.

But it's already quite good to land, will be great if we can expect following fixes, thanks!

Yukiniro:
> Found a minor regression that the bottom right cursor should be `nwse-resize` instead of `ew-resize`.
> 
> But it's already quite good to land, will be great if we can expect following fixes, thanks!

Thanks for reviewing!  I will fix it later.

Issue #2668, opened 2023-05-19
Support date field for database block


houjoe0829:
User can set the display format of the date, including whether to include the time, and in what format to display it.

xell:
I generally agree with the suggestion. Perhaps using the HTML standard input like

```
<input type="date">
```

would be enough for providing a basic date functionality?

PR #3608, opened 2023-07-19
feat(database): add `date` column

zqran:
close: #2668

https://github.com/toeverything/blocksuite/assets/15389209/80c7fb9f-a346-4eff-bb24-c9428e057c55


packages/blocks/src/database-block/common/column-manager.ts
```diff
@@ -198,6 +198,12 @@ export const linkHelper = columnManager.register<string>('link', {
   configRender: () => html``,
   cellToString: data => data?.toString() ?? '',
 });
+export const dateHelper = columnManager.register<string>('date', {
```
zzj3720, line 201: I think timestamp is better than date string

zzj3720 (approved):

PR #3637, opened 2023-07-20
feat: add path and view map in lit root

Saul-Mirone:


doodlewind (approved):

packages/lit/src/element/block-element.ts
```diff
@@ -67,6 +67,27 @@ export class BlockElement<
...
+
+  get pathname(): string {
+    return this.path.join('|');
+  }
+
+  get parentPath(): string[] {
+    return this.path.slice(0, -1);
+  }
+
+  parentBlockElement = <Block extends BlockElement>(): Block | undefined => {
```
doodlewind, line 81: This looks like a getter? Also the `T | null` typing should be more favorable in our codebase.

Saul-Mirone: Updated

PR #3644, opened 2023-07-21
feat(block-std): bind selection manager to lit root

Saul-Mirone:


doodlewind (approved):

packages/playground/src/main.ts
```diff
@@ -97,6 +98,11 @@ async function main() {
   window.Y = Workspace.Y;
   window.std = std;
   window.ContentParser = ContentParser;
+  Object.defineProperty(globalThis, 'root', {
```
doodlewind, line 101: Perhaps it's better to update `env.d.ts` so that we use `window.root` here?

Saul-Mirone: I just updated it.

PR #3655, opened 2023-07-21
fix(page): cursor & selection after pasting image

### LINKED ISSUE #3639: The cursor and highlight didn't behave as expected after pasting the image
### What happened?

paste-with-image

1. Copy the cover image from: https://affine.pro/blog/whats-new-affine-0630
2. Paste in Notion: The embed block is highlighted, and I can use the up/down arrow key to navigate between blocks.
3. Paste in AFFiNE: No block is highlighted, and when the up/down arrow key are pressed, the cursor does not move, but the entire document scrolls.

Expect：

1.  The embed block is highlighted correctly.
2. When the up or down arrow keys are pressed, the cursor will behave in the same way as it does when navigating through paragraphs.

### Distribution version

macOS x64

Chrome

### Relevant log output

_No response_

### Anything else?

Version 0.7.0-canary.47 (0.7.0-canary.47)

- [X] Yes I'd like to help by submitting a PR!

youhua:
try to fix #3639 

1. Is there any function to wait block ready when focusBlockByModel?
2. pageBlock.selection break assertImageOption, I temporarily passed the test case by adding a click event.


tests/clipboard.spec.ts
```diff
@@ -139,6 +141,56 @@ test(
...
+      type: 'issue',
+      description: 'https://github.com/toeverything/blocksuite/issues/3639',
+    });
+    await enterPlaygroundRoom(page);
+    await initEmptyParagraphState(page);
+    await focusRichText(page);
+
+    // set up clipboard data using html
+    const clipData = {
+      'text/html': `<p>符合 Markdown 格式的 URL 放到笔记中，此时需要的格式如下：</p>
```
doodlewind, line 157: Would be better to use English as test case content, unless it's related to IME or emoji.

packages/blocks/src/__internal__/service/json2block.ts
```diff
@@ -141,7 +145,10 @@ export async function json2block(
         length: 0,
       });
     } else {
-      // TODO: set embed block selection
+      requestAnimationFrame(() => {
+        // TODO: wait block ready
+        focusBlockByModel(lastModel);
+      });
```
fourdim, line 151: What does this mean?

fourdim: Why do you want to focus an image?

youhua: You can find the context from #3639 

Most document tools, such as Notion and Pages, will focus on the image after pasting, allowing users to easily move, resize, delete and perform other operations on the image. 

While focusing on the image, the system is also able to respond appropriately to multi-paste events.

fourdim: Unfortunately, the selection manager module is currently undergoing a process of refactoring.
This part is temporarily reverted due to a compatibility issue. We will add it back after the refactoring is done.

PR #3666, opened 2023-07-23
feat: allow null value in editor container

himself65:


doodlewind:
Closing due to conflicts and potential API updates in `EditorContainer`, we may split it into multiple editors specific to different modes, and provide a new `ComposedEditor` as API surface for AFFiNE.

/cc @Saul-Mirone

PR #3695, opened 2023-07-25
feat(page): new drag handle

donteatfriedrice:


packages/blocks/src/widgets/drag-handle/index.ts
```diff
@@ -0,0 +1,580 @@
+import type {
```
doodlewind, line 1: Is it possible to remove the legacy `blocks/src/components/drag-handle.ts` for now? Regressions are acceptable here.

donteatfriedrice: I will verify whether the new drag handle can work properly under edgeless mode, if it works well, I will remove the legacy drag-handle component

packages/blocks/src/widgets/drag-handle/index.ts
```diff
@@ -0,0 +1,566 @@
...
+    const { target } = state.raw;
+    const element = this._captureEventTarget(target);
+
+    const inside = !!element?.closest('.affine-drag-handle-container');
+    if (!inside || !this._hoveredBlockId || !this._hoveredBlockPath) {
+      return;
+    }
+
+    // Get current hover block eleemnt by path
+    const hoverBlockElement = this.root.viewStore.getViewByPath(
```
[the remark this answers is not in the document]
donteatfriedrice: OK, I will update code and use the new store API. Thanks!

doodlewind (approved):

PR #3701, opened 2023-07-26
feat(edgeless): frame

### LINKED ISSUE #2325: Support new frame element in edgeless
The new frame element should not be confused with the legacy frame block. Frame elements can be used to "group" visual contents inside in edgeless mode.

Previous goal:

<del>It can be simply understood as the presentation effect of Miro's presentation based on frames. We will invest more design into it and create even greater highlights.</del>

Related: https://github.com/toeverything/blocksuite/issues/2196

regischen:
closes #2325 

# Some new concepts introduced: 

naming could be discussed.

1. ```type EdgelessElement = TopLevelBlockModel | PhasorElement```

2. surface-element has a new getter called ```gridBound```
I think the concept rotate should better not exposed to `Grid`. So we should provide unified api to it.
And in this case, frame have title, and we can select frame by clicking title. So the area of title should be considered in `Grid`.
But not in `xywh`

3.  `Page` has a new Slot called `yBlockUpdated`.
Because I want yblock event is emitted from page instead of emitted from individual model. and we can get the old value and the new value from the event.


regischen:
> Some new concepts introduced:
> 
> 1. `type EdgelessElement = TopLevelBlockModel | PhasorElement`
> 2. surface-element has a new getter called `gridBound`
>    I think the concept rotate should better not exposed to `Grid`. So we should provide unified api to it.
>    And in this case, frame have title, and we can select frame by clicking title. So the area of title should be considered in `Grid`.
>    But not in `xywh`
> 3. `Page` has a new Slot called `yBlockUpdated`.
>    Because I want yblock event is emitted from page instead of emitted from individual model. and the props updated, we can get the old value and the new value.

pnpm-lock.yaml
```diff
@@ -30,109 +30,109 @@ importers:
         version: link:packages/virgo
       '@commitlint/cli':
         specifier: ^17.6.7
-        version: 17.6.7
+        version: registry.npmmirror.com/@commitlint/cli@17.6.7
```
doodlewind, line 33: This should be reverted. The unnecessary diff looks very careless.

packages/blocks/src/page-block/edgeless/components/toolbar/edgeless-toolbar.ts
```diff
@@ -159,9 +200,156 @@ export class EdgelessToolbar extends WithDisposable(LitElement) {
...
+      const viewport = this.edgeless.surface.viewport;
+      const frame = frames[current];
+      if (frame) {
+        const bound = Bound.deserialize(frame.xywh);
+        viewport.setViewportByBound(bound, [40, 60], true);
+      }
+    }
+  }
+
+  private get prensentContent() {
```
doodlewind, line 217: The `_foo` convention also applies for methods.

packages/blocks/src/page-block/edgeless/components/toolbar/edgeless-toolbar.ts
```diff
@@ -98,10 +102,47 @@ export class EdgelessToolbar extends WithDisposable(LitElement) {
...
+
+    .edgeless-present-frame {
+      width: 145px;
+      text-align: center;
+      display: flex;
+      align-items: center;
+      justify-content: center;
+    }
+
+    .edgeless-prensent-frame-title {
```
doodlewind, line 114: Should be `.edgeless-frame-navigator-title`. Please remove all the *present* vocabularies here (and replace them as *frame navigator*), since *present* it's a verb that doesn't fit well, and its noun form means gift, which is also wrong in our case.

packages/blocks/src/page-block/edgeless/components/toolbar/edgeless-toolbar.ts
doodlewind: Also, I think we should extract the frame navigator toolbar as a component.

regischen: At first, I also was intended to make a new component for frame navigator.
But in this way, I can't reuse it's parent (toolbar container) style because lit doesn't have sort of `Fragment`.

regischen: > The `_foo` convention also applies for methods.
> 
> But here, this variable and the `defaultContent` should both be extracted as pure template function (with `MyTemplate` naming convention), instead of private methods.

defaultContent depends on many states of `EdgelessToolbar`, is it better to be extracted as a pure template function?

packages/blocks/src/page-block/edgeless/components/toolbar/edgeless-toolbar.ts
```diff
@@ -159,9 +200,161 @@ export class EdgelessToolbar extends WithDisposable(LitElement) {
     }
   };
 
+  protected override updated(changedProperties: PropertyValues) {
+    const { type } = this.edgelessTool;
+    if (
+      changedProperties.has('_currentFrameIndex') &&
+      type === 'frameNavigator'
+    ) {
+      const frames = this.edgeless.surface.getElementsByType('frame');
```
doodlewind, line 209: So when every time the toolbar gets updated, we need to iterate over all elements?

regischen: uh, it's not good, I'll fix it.
since we can't add or delete frame on presentation, I'll cach frames

doodlewind (approved):

PR #3729, opened 2023-07-27
feat(block-std): add `isCollapsed` and `isInSameBlock` in `TextSelection`

Flrande:


packages/block-std/src/selection/variants/text.ts
```diff
@@ -69,6 +70,14 @@ export class TextSelection extends BaseSelection {
       to: json.to as TextRangePoint | null,
     });
   }
+
+  isCollapsed(): boolean {
+    return this.to === null;
+  }
+
+  isInSameBlock(): boolean {
```
Saul-Mirone, line 78: Should `isInSameBlock` return `true` if the range is collapsed?

PR #3731, opened 2023-07-27
feat(edgeless): support batch in surface

regischen:
support `Batch` to render frame and nonFrame properly.
also useful for grid search.


packages/blocks/src/note-block/note-model.ts
```diff
@@ -21,6 +21,8 @@ export const NoteBlockSchema = defineBlockSchema({
     background: DEFAULT_NOTE_COLOR,
     index: 'a0',
     hidden: false,
+    frame: null,
```
doodlewind, line 24: Won't require persisting this to represent frame.

packages/phasor/src/Layer.ts
```diff
@@ -0,0 +1,37 @@
+export class Layer<T extends { index: string }> {
```
doodlewind, line 1: Instead of defining _layer_, I suggest using the concept of "_batch_" that allows distributing elements in a coarse grained way. The comparison of indexes only makes sense when two elements are in the same batch. Each batch has a string index, so that the indexes are comparable between batches.

But this should not be a `static` property per element type (which is correct according to our current requirement, but too ad-hod as a framework level assumption). What if we want two text elements to be used in two different batches?

Instead, every element should be allowed to have an optional CRDT `batch` property to indicate its batch index (aka batch id). If this property doesn't exist, we use the default batch. So in this way, all frame elements should have a `batch` property that explicitly mark them to be rendered beneath the other elements.

So what if we compare two elements in this way?

```
TextElement (batch = null)
FrameElement (batch = `a0`)
```

Here the default batch index should be declared as `const` (rather than CRDT property) and set using API like `surface.setDefaultBatch`, so that all element indexes could be comparable without involving the element type.

Also `Layer.ts` -> `batch.ts`, please keep in mind with the capitalization convention.

doodlewind (approved):

PR #3741, opened 2023-07-27
fix(page): add selection indicator for embed block

lawvs:
## New component

```ts
@customElement('affine-block-selection')
export class BlockSelection extends LitElement
```

## Usage

 ```ts
 class Block extends LitElement {
   state override styles = css`
     :host {
       position: relative;
     }
   `
   render() {
     return html`${this.selected?.is('block')
       ? html`<affine-block-selection></affine-block-selection>`
       : null}`;
   };
 }
 ```

https://github.com/toeverything/blocksuite/assets/18554747/2751fed3-c611-4eaa-8ec9-32aaa18d82dc

### DIFF
--- a/packages/blocks/src/divider-block/divider-block.ts
+++ b/packages/blocks/src/divider-block/divider-block.ts
@@ -47,7 +47,7 @@ export class DividerBlockComponent extends BlockElement<DividerBlockModel> {
     </div>`;
 
     return html`
-      <div class="affine-divider-block-container" }>
+      <div class="affine-divider-block-container">
         <hr />
         ${children}
         ${this.selected?.is('block')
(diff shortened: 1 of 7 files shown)


packages/blocks/src/divider-block/divider-block.ts
```diff
@@ -46,9 +47,12 @@ export class DividerBlockComponent extends BlockElement<DividerBlockModel> {
     </div>`;
 
     return html`
-      <div class=${`affine-divider-block-container`}>
+      <div class="affine-divider-block-container" }>
```
Saul-Mirone, line 50: The } bracket seems need to be removed.

lawvs: nice catch

doodlewind (approved):

PR #3746, opened 2023-07-28
refactor(page): migrate to pdf-lib

zuoxiaodong0815:
The ’jsPDF‘ modify the ‘then’ and ’catch‘ methods. But in Affine, modifying the prototype chain is not allowed at runtime.


doodlewind (approved):
LGTM

Issue #2945, opened 2023-06-01
Support toggling for headings and list blocks


houjoe0829:
PRD：[Internal doc](https://toeverything.feishu.cn/wiki/Ds3BwBM1ViKxC8kmDaTcHrt0nFc)

houjoe0829:
In terms of implementation priority, Lists > Headings.

lawvs:
## Caveats

- Undo/Redo
- Collaboration
- Copy/Paste
- Update or rerender

xell:
@lawvs please see the [internal figma design](https://www.figma.com/file/hVxJQtEqyv5xw1AbCzc50M/%F0%9F%97%92%EF%B8%8F-AFFiNE-2023?type=design&node-id=5659-121035&mode=design&t=wLNBCRWgYINIigeE-4) for the details.

PR #1139, opened 2023-02-11
feat(blocks): support `toggle` list type

gotjoshua:
This continues from #908 with a cleaner git tree.

Current roadmap:

P0
- [x] Switch `showChildren` to non-CRDT UI state
- [x] Fix hover DOM error
- [x] Revert hard-coded states on `page`
- [x] Revert placeholder
- [x] use common icon in slash menu and ui
- [ ] Fix test case

P1
- [ ] Fix wrong selection state and console error when there are nested content inside collapsed list
- [ ] Fix weird hide/show state change
- [ ] Forward compatibility (make sure unknown `list.type` has its fallback, since the toggle type will be a breaking change for Abbey Wood version - we'd better make sure we can be compatible with future updates)

PR #3770, opened 2023-07-30
feat(page): support universal list toggle

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/2945

- [x] Add toggle for list
- [x] Update editor layout
- [x] Update drag handle
- [x] Fix tests
- [x] Add new tests for toggle list
- [x] Update the `theme` package
  - See https://github.com/toeverything/design/pull/90
- [x] Update mobile padding
  - `@media screen and (max-width: 640px)`

Credits to https://github.com/toeverything/blocksuite/pull/1139

## Side effects

- Now a new class `with-drag-handle` will be added to block elements when the drag handle hovering
- Update `--affine-editor-width` from 800px to 944px
- Update `affine-doc-page-block-container` padding from 24px to 96px
- Update playwright viewport from 900px to 960px

---

https://github.com/toeverything/blocksuite/assets/18554747/868a95d6-5102-41ed-8618-bc8e2c5785fa

### DIFF
--- a/packages/blocks/package.json
+++ b/packages/blocks/package.json
@@ -18,7 +18,7 @@
     "@blocksuite/block-std": "workspace:*",
     "@blocksuite/lit": "workspace:*",
     "@blocksuite/store": "workspace:*",
-    "@toeverything/theme": "^0.7.15",
+    "@toeverything/theme": "^0.7.16",
     "nanoid": "^4"
   },
   "dependencies": {
@@ -78,7 +78,7 @@
     "@blocksuite/block-std": "workspace:*",
     "@blocksuite/lit": "workspace:*",
     "@blocksuite/store": "workspace:*",
-    "@toeverything/theme": "^0.7.15",
+    "@toeverything/theme": "^0.7.16",
     "@types/marked": "^4.3.1",
     "@types/sortablejs": "^1.15.1"
   }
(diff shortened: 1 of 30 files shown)


hwangdev97:
I think when we active drag handle, the toggle icon should display. 

shot-2023-09-13-19 55 17@2x

hwangdev97 (approved):

hwangdev97:
For UI/UX, it’s ok 👍👍👍👍

donteatfriedrice:
In edgeless mode, drag handle looks like being out of note when there is a toggle list, maybe need to optimize the offset left in edgeless mode.

https://github.com/toeverything/blocksuite/assets/99816898/0bb18a99-a5ba-4e69-a96c-cf8b9674a8c4

hwangdev97:
> In edgeless mode, drag handle looks like being out of note when there is a toggle list, maybe need to optimize the offset left in edgeless mode or expand the note padding?
> 
>  Screen.Recording.2023-09-15.at.11.08.01.mov

We will modify the note editing state of edgeless so that it can be better compatible with related content, so there is no need to process it in this PR.

doodlewind (approved):
This has been a long journey and we should land this, thanks for the effort!

PR #3776, opened 2023-07-31
fix(edgeless): default entry playground should use multi-user compatible generator

### LINKED ISSUE #3764: Empty content in new default playground
image

doouding:
close #3764


doodlewind (approved):
Cheers!

PR #3779, opened 2023-08-01
chore: move @toeverything/theme to peer dependency

lawvs:
Fix the issue with possibly inconsistent theme versions of the blocksuite and affine.


doodlewind (approved):

doodlewind:
Thought this could work and you can merge this for the moment, I don't think it's good if our in-house package also has be installed _explicitly_ for third party devs to use BlockSuite 🤔

himself65 (approved):
This might cause the build error in AFFiNE when api changes. But it's fine

PR #3784, opened 2023-08-01
fix(edgeless): rotate text incorrectly after resizing

### LINKED ISSUE #3783: Rotate text incorrectly after resizing
Resize the text, then rotate it, the center of text rotation might be wrong (the selected rect is right).

https://github.com/toeverything/blocksuite/assets/17249775/9fc53d43-2933-4231-9796-d9d7a1dc7dbd

zhengjitf:
Close #3783


doodlewind (approved):
Seems that this is fixed in master, but let's keep the record to mark the effort as happened, thanks!

PR #3788, opened 2023-08-01
feat(page): support adding attachment by drag and drop

lawvs:
I believe DND should be refactored into a widget.


doodlewind (approved):
We should land this but also keep in mind with the widget stuff.

PR #3790, opened 2023-08-02
feat(database): kanban UI

zzj3720:
[design preview](https://www.figma.com/file/rfH3P9Wbz4IEp2Ai18O8dV/%E2%96%B6%EF%B8%8F-Affine.2023.Jul?type=design&node-id=4021-58147&mode=dev)
image


packages/blocks/src/database-block/kanban/kanban-view.ts
```diff
@@ -1,6 +1,7 @@
 import './group.js';
 import './header.js';
 import './drag.js';
+import '../common/groupBy/define.js';
```
doodlewind, line 4: Please use `group-by` as naming convention for files.

Issue #3626, opened 2023-07-20
database block data incompatibility


zzj3720:
#3421

PR #3793, opened 2023-08-02
refactor(database): v2 -> v3 migration

zzj3720:
close: #3626

In the v2 version of the data, the title column is not a normal column and does not exist in the view data, which makes it necessary to always write special logic for the title, and the title cannot have different widths in different views.

In the v3 version of the data, the title is stored as a normal column, with only special handling of the column type, in line with other special columns that may be available in the future.

example:
image


Flrande (approved):

zzj3720:
> Could you clarify what's the difference between V2 and V3 in the PR description? This should help in our release note. Thanks!

done

PR #3816, opened 2023-08-03
feat(edgeless): add shell component for surface block

Saul-Mirone:


packages/blocks/src/surface-block/surface-block.ts
```diff
@@ -0,0 +1,13 @@
+import { BlockElement } from '@blocksuite/lit';
+import { customElement } from 'lit/decorators.js';
+
+import type { SurfaceBlockModel } from './surface-model.js';
+
+@customElement('affine-surface')
+export class SurfaceBlockComponent extends BlockElement<SurfaceBlockModel> {}
```
doodlewind, line 7: Having a `render` method that returns `nothing` may be semantically better 🌚

doodlewind (approved):

packages/blocks/src/surface-block/surface-block.ts
Saul-Mirone: Done

PR #3820, opened 2023-08-03
test(edgeless): add frame tests

regischen:


doodlewind (approved):
LGTM, good to see tests here!

PR #3821, opened 2023-08-03
refactor(page): bind range-manager and range-synchronizer to edgeless

Flrande:


packages/blocks/src/page-block/keyborad/keyboard-manager.ts
```diff
@@ -195,10 +196,8 @@ export class Keyboard {
   ) {
     const current = selections[0];
     const first = this._page.getBlockById(current.blockId);
-    const firstElement = this.host.root.viewStore.viewFromPath(
-      'block',
-      current.path
-    );
+    const firstElement = this.host.root.viewStore.getViewByPath(current.path)
```
Saul-Mirone: Can be replaced with `const firstElement = this.host.root.viewStore.viewFromPath('block', current.path)`.

Saul-Mirone (approved):

PR #3830, opened 2023-08-03
feat: add attachment options

lawvs:


packages/blocks/src/attachment-block/components/options.ts
```diff
@@ -0,0 +1,264 @@
...
+      left + width
+    }px - 100%), calc(${top}px - 100% - 4px))`;
+  };
+
+  const viewportElement = getViewportElement(model.page);
+  if (viewportElement) {
+    viewportElement.addEventListener('scroll', updatePosition);
+  }
+
+  setTimeout(() => {
```
doodlewind, line 87: Is this required for all portal component users?

packages/blocks/src/components/portal.ts
```diff
@@ -41,6 +45,34 @@ export class Portal extends LitElement {
...
 }
 
+/**
+ * Similar to `<affine-portal>`, but only renders once when called.
+ *
+ * The template should be a **static** template since it will not be re-rendered.
+ *
+ * See {@link Portal} for more details.
+ */
+export function createLitPortal({
```
Saul-Mirone, line 55: I think we can make the `affine-portal` classname as part of options, and move this into `@blocksuite/lit`

packages/blocks/src/attachment-block/components/options.ts
lawvs: The key to `setTimeout` lies in the `containerRef`. In order to update the position of the HTMLElement, specific DOM information is required.

doodlewind: Is it possible to encapsulate these logic into some util methods together with `createLitPortal`? The current API surface looks a bit too verbose and fragmented.

packages/lit/src/portal.ts
```diff
@@ -0,0 +1,96 @@
...
+ * @example
+ * ```ts
+ * render() {
+ *   return html`${showPortal
+ *     ? html`<affine-portal .template=${portalTemplate}></affine-portal>`
+ *     : null}`;
+ * };
+ * ```
+ */
+@customElement('affine-portal')
```
Saul-Mirone, line 26: Maybe `blocksuite-portal` ?
Should we keep affine as prefix for our common elements? @doodlewind

packages/blocks/src/attachment-block/components/options.ts
lawvs: We can encapsulate a higher-level API together as a higher-level API in the future.

```ts
createPopoverPortal({
  reference,
  template, // <- portal template
  positionOptions,
  ... createLitPortalOptions
})
```

Ref API
```ts
import { computePosition, offset } from '@floating-ui/dom';

  computePosition(reference, menu, {
    placement: 'top',
    middleware: [
      offset({
        mainAxis: 10,
      }),
    ],
  }).then(({ x, y }) => {
    Object.assign(menu.style, {
      left: `${x}px`,
      top: `${y}px`,
    });
  });
```

We can also add lifecycle to `createLitPortal`, but this will reduce the understanding and convenience of createLitPortal.

```ts
const createLitPortal({
  render: ({ onMount, befortUnmount }) => TemplateResult<1>;
  container?: HTMLElement;
  abortController?: AbortController;
}): HTMLDivElement
```

packages/lit/src/portal.ts
doodlewind: Yes should be `blocksuite-portal`. Having this prefix sounds like a good convention to follow, when providing components that are expected to be used outside of the editor.

PR #3838, opened 2023-08-04
feat(page): support bookmark block in exporting html/markdown

### LINKED ISSUE #3839: support bookmark block in exporting html/markdown

zuoxiaodong0815:
close #3839

### DIFF
--- a/packages/blocks/src/__internal__/content-parser/file-exporter/exporter-style.ts
+++ b/packages/blocks/src/__internal__/content-parser/file-exporter/exporter-style.ts
@@ -121,9 +121,118 @@ export const globalCSS = css`
     --affine-palette-shape-black: #000;
     --affine-palette-shape-grey: #c2c2c2;
     --affine-tooltip: #424149;
+    --affine-shadow-1: 0px 0px 4px 0px rgba(66, 65, 73, 0.14);
+    --affine-card-background-blue: linear-gradient(
+      180deg,
+      #f0f3fd 0%,
+      #fcfcfd 100%
+    );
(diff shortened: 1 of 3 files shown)


packages/blocks/src/__internal__/content-parser/file-exporter/exporter-style.ts
```diff
@@ -121,9 +121,118 @@ export const globalCSS = css`
     --affine-palette-shape-black: #000;
     --affine-palette-shape-grey: #c2c2c2;
     --affine-tooltip: #424149;
+    --affine-shadow-1: 0px 0px 4px 0px rgba(66, 65, 73, 0.14);
```
doodlewind, line 124: This naming looks a bit off 🤔

zuoxiaodong0815: The properties and values here refer to this code, https://github.com/toeverything/design/blob/main/packages/theme/src/index.ts
image

zuoxiaodong0815: I haven't come up with a final method for css variables in exporting file.

The first way is to rewrite css variables like the current way.

The second way is to import theme package in blocksuite.

The third way is to collect all css variables when exporting file. Like following,
```typescript
  const cssAffineVars = {};
  const styles = window.getComputedStyle(document.documentElement);

  for (const prop of styles) {
    if (prop.startsWith("--affine-")) {
      cssAffineVars[prop] = styles.getPropertyValue(prop);
    }
  }
```

doodlewind (approved):
Let's just land this for the moment.

PR #3856, opened 2023-08-06
test: cleanup invalid test cases

zhengjitf:
- `page.evaluate()` returns a serializable value or undefined, in this case, `Array.from(notes)` will be `[]`
- loop promises mistakenly


doodlewind (approved):
Thanks for the help maintaining our test cases. More updates here are welcomed!

PR #3860, opened 2023-08-06
feat(database): optimize native selection

zqran:
1. Prevent native selection from being selected while dragging.
2. Execute the default copy function of the browser when editing columns such as `number/date/link/select`.

### DIFF
--- a/packages/blocks/src/database-block/table/clipboard.ts
+++ b/packages/blocks/src/database-block/table/clipboard.ts
@@ -25,6 +25,8 @@ type TableViewClipboardConfig = {
   data: DataViewManager;
 };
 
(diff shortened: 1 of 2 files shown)


packages/blocks/src/database-block/table/clipboard.ts
```diff
@@ -25,6 +25,8 @@ type TableViewClipboardConfig = {
   data: DataViewManager;
 };
 
+const columnsContainInput = ['number', 'date', 'link', 'select'];
```
zzj3720, line 28: Consider support for third-party columns

zzj3720 (approved):

PR #3865, opened 2023-08-07
feat: use new migration api

Saul-Mirone:
/cc @pengx17 @himself65 @doodlewind @Flrande 

Align the API with #3826 .

```ts
import { Schema } from '@blocksuite/store';

const schema = new Schema();
schema.register(blockSchemas);

// run workspace level migration
schema.upgradeWorkspace(rootYDoc);

// run migration for pages
schema.upgradePage(blockVersions, pageYDoc);
```


himself65 (commented):
This is just put the code in another place but solve nothing

Saul-Mirone:
> This is just put the code in another place but solve nothing

Align the API with https://github.com/toeverything/blocksuite/pull/3826

PR #3410, opened 2023-07-06
fix(edgeless): note menu scroll too fast

donteatfriedrice:
To fix #3404 

https://github.com/toeverything/blocksuite/assets/99816898/0aaab560-b810-462f-bf4f-3348eedb8745

PR #3870, opened 2023-08-07
feat: add `favoritePages` to workspace meta to support favorite page list sortable

teobler:
related to affine [#3410](https://github.com/toeverything/AFFiNE/issues/3410)

### DIFF
--- a/packages/store/src/workspace/meta.ts
+++ b/packages/store/src/workspace/meta.ts
@@ -25,6 +25,7 @@ type PagesPropertiesMeta = {
 };
 type WorkspaceMetaState = {
   pages?: unknown[];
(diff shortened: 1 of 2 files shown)


lawvs:
Blocksuite is an editor for general purposes, so I don't think we should add an additional special purpose field to it.

In fact, affine has overridden this type and has included a "favorite" field within it.

See also https://github.com/toeverything/AFFiNE/blob/d96263fde95f15ade3f801ef7f112405acf76069/packages/hooks/src/use-block-suite-page-meta.ts#L8-L22

teobler:
Okay, thanks for clarification, just thought blocksuite was built for affine, I will back to affine then.

PR #3872, opened 2023-08-07
fix(page): support chinese slash

### LINKED ISSUE #3871: Symbol '、' cannot be replaced when  `_handleClickItem` execute
After closing #2416, AFFiNE now supports '、' to open the slash menu. But the symbol '、' didn't be replaced when executing the `cleanSpecifiedTail` function.

Kapture 2023-08-07 at 20 24 09

yuusheng:
close #3871

### DIFF
--- a/packages/blocks/src/widgets/slash-menu/slash-menu-popover.ts
+++ b/packages/blocks/src/widgets/slash-menu/slash-menu-popover.ts
@@ -1,11 +1,12 @@
 import { WithDisposable } from '@blocksuite/lit';
-import { type BaseBlockModel } from '@blocksuite/store';
+import { assertExists, type BaseBlockModel } from '@blocksuite/store';
 import { html, LitElement, nothing } from 'lit';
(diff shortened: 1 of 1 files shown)


yuusheng:
There are two ways to solve this:
1. pass a correct slash when calling `cleanSpecifiedTail`.
2. handle Chinese slash in `cleanSpecifiedTail`.

I think we should pass a correct slash to make sure the `cleanSpecifiedTail` function didn't change.

lawvs:
Due to the trigger key of the slash menu being configurable, it may not only contain a single character, so we need a more generic but less elegant way to handle this case.

See also https://github.com/toeverything/blocksuite/blob/4012030acab70bc8334154c12c769a76c7c5c74f/packages/blocks/src/widgets/slash-menu/index.ts#L75-L96

Closed in favor of https://github.com/toeverything/blocksuite/pull/3878, thanks for your effort anyway!

PR #3884, opened 2023-08-08
test(edgeless): fix frame test

Flrande:


doodlewind (approved):

regischen:
This can be reverted if #3887 fixed

PR #3891, opened 2023-08-08
feat(edgeless): support new drag handle in edgeless

donteatfriedrice:


doodlewind (approved):
Great to see this finally works, thanks!

Issue #2952, opened 2023-06-01
Support attachment block


houjoe0829:
Attachment preview is not within the scope of this task.

Svaney-ssman:
[internal doc](https://toeverything.feishu.cn/wiki/X8Bkwj9triEVqbkxMImc7e3Hnwd)

Svaney-ssman:
[internal figma](https://www.figma.com/file/hVxJQtEqyv5xw1AbCzc50M/AFFiNE-2023?type=design&node-id=6306-131013&mode=design) @joebeijing

doodlewind:
This is closed by #3622, we'll track new related issues after new canary releases.

PR #3896, opened 2023-08-08
feat(page): add caption for attachment

lawvs:
Part of https://github.com/toeverything/blocksuite/issues/2952

https://github.com/toeverything/blocksuite/assets/18554747/cc165947-c8b5-4fbe-b52b-0180b7f80921


packages/blocks/src/attachment-block/components/moreMenu.ts
doodlewind, line 1: Please keep the `more-menu.ts` naming style convention.

PR #3897, opened 2023-08-08
fix(page): enable and fix some tests

Flrande:


doodlewind (approved):
Great to see tests back!

Issue #3883, opened 2023-08-03
Wrong position of the icon of Linked to this page


xell:
It should align horizontally with the "Linked to this page", instead of being vertically in the whole height

![CleanShot 2023-08-03 at 09 35 04@2x](https://github.com/toeverything/AFFiNE/assets/132558/d2464daa-b159-4edf-b967-908a3bd5b350)

JimmFly:
> @xell can you please provide me just the the page url or the file name in which I should make the changes, anyone of it will workout for me



> > @xell can you please provide me just the the page url or the file name in which I should make the changes, anyone of it will workout for me
> 
> You're welcome to contribute to the repo. And perhaps @DiamondThree or @JimmFly can help you about this.

This is Blocksuite‘s issue, and I transferred it to  Blocksuite.

RajeevDash007:
@JimmFly can you please send me the url of this page, so that I can see the output please,
or else just say the flow to reach the destination section after entering the landing page

JimmFly:
> @JimmFly can you please send me the url of this page, so that I can see the output please, or else just say the flow to reach the destination section after entering the landing page

https://github.com/toeverything/blocksuite/blob/ef3445a9f6f32b9481ec81f0a36458dc41c66101/packages/blocks/src/page-block/doc/meta-data/meta-data.ts#L396
It is rendered here, you can use the '@' symbol to create a page on a blank line, and then the two pages will be concatenated.

https://github.com/toeverything/blocksuite/assets/102217452/abc27dfd-1e79-486b-a3af-c0fe37713a8d

PR #3905, opened 2023-08-08
fix: align arrow icon in page info

RajeevDash007:
fixed: #3883 
BlockSuite-Playground

If you need any changes then let me know.
Thank you so much for your assistance.

### DIFF
--- a/packages/blocks/src/page-block/doc/meta-data/meta-data.ts
+++ b/packages/blocks/src/page-block/doc/meta-data/meta-data.ts
@@ -140,7 +140,7 @@ export class PageMetaData extends WithDisposable(LitElement) {
 
     .meta-data-expanded-item .arrow-icon svg {
       fill: var(--affine-icon-color);
-       height: 25px;
+      height: 25px;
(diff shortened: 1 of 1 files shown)


RajeevDash007:
@JimmFly just check this PR once

JimmFly:
Splendid, @xell , is this precisely what you desire?

RajeevDash007:
@xell will that be okay?
Screenshot (7)

xell (commented):
The alignment is better now. But the class name would be better to use `link-icon` or something like that to reflect the fact that it is associated with an icon named `DualLinkIcon`, by which it's beneficial for the future code explanation, searching and debug, among others.

doodlewind:
@xell for naming related reviews, it's recommended to directly push to the PR branch, this could make the review process more efficient.

xell (approved):

doodlewind:
Should land this for further testing in AFFiNE, thanks @RajeevDash007 for the collaboration!

Issue #3781, opened 2023-08-01
Wrong font in the More menu of toolbar in Edgeless


xell:
The font should stick to the UI font family instead of changing with the editor.

![image](https://github.com/toeverything/blocksuite/assets/132558/8eb9e5b3-9282-491f-a8ce-d604648967ec)

hwangdev97:
I believe not just this menu, but all aspects of the edgeless UI with font settings seem to be problematic:

![shot-2023-08-09-17 37 20@2x](https://github.com/toeverything/blocksuite/assets/24713927/715c17a4-47ab-4dd7-9bfe-50d50494700b)

![shot-2023-08-09-17 39 51@2x](https://github.com/toeverything/blocksuite/assets/24713927/d1da8b4e-f501-4897-aa6a-e63da497387a)

doodlewind:
Could @RajeevDash007 try to fix these related menu styling issue together in the existing PR? Thanks!

RajeevDash007:
Yup, I am on it

PR #3910, opened 2023-08-09
fix: added base UI font family to the Menu of edgeless

RajeevDash007:
fixed: #3781 
Screenshot (10)
@xell check this out

### DIFF
--- a/packages/blocks/src/page-block/edgeless/components/component-toolbar/more-button.ts
+++ b/packages/blocks/src/page-block/edgeless/components/component-toolbar/more-button.ts
@@ -58,6 +58,7 @@ export class EdgelessMoreButton extends WithDisposable(LitElement) {
       display: block;
       color: var(--affine-text-primary-color);
       fill: currentColor;
+      font-family: var(--sl-font-sans);
     }
 
     .more-actions-container {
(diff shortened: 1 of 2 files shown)


RajeevDash007:
@xell @doodlewind 
added font to the Toolbar too.
Screenshot (24)

doodlewind (approved):
Thanks for the quick response! It's land-able since CI passes and the patch looks pretty neat, but I'd suggest just keep this open for a while, since we are trying the workflow that for styling issues labeled with `mod:style`, we await for the design team to verify the effect. So /cc @xell @hwangdev97 to confirm.

xell (approved):

packages/blocks/src/page-block/edgeless/components/component-toolbar/more-button.ts
```diff
@@ -58,6 +58,7 @@ export class EdgelessMoreButton extends WithDisposable(LitElement) {
       display: block;
       color: var(--affine-text-primary-color);
       fill: currentColor;
+      font-family: var(--sl-font-sans);
```
doodlewind, line 61: @RajeevDash007 though this works, but seems that the `--sl-font-sans` is the [Shoelace](https://shoelace.style/) variable that only works in the BlockSuite playground environment (we use Shoelace for debug menu here, which is not included in the exported editor), but we intend to fix this in AFFiNE. So here the correct fix should be using the `--affine` prefixed variable.

RajeevDash007:
@doodlewind check  #3954

PR #3911, opened 2023-08-09
perf(edgeless): remove over rendering on mouse move and viewport updates

### LINKED ISSUE #3916: Redundant re-render in edgeless mode
Currently the block UI components are re-rendered on every mouse move and viewport panning / zooming. The expected behavior is to make it on-demand, just like in doc mode.

We should not use `requestUpdate` too frequently. We've applied this pattern in #1472 and #1520.

regischen:
Close #3916


doodlewind (approved):
Thanks for the optimization! We should land this for now, since the perf regression on panning viewport and mouse move has been solved.

But there are still over-rendering in local editing, which can be addressed in following PRs.

Doc mode:

https://github.com/toeverything/blocksuite/assets/7312949/70241c88-be8e-4523-8897-f029f2f2c9a4

Edgeless mode:

https://github.com/toeverything/blocksuite/assets/7312949/5cbe105f-ce51-4ec6-880b-6d97faca11ca

(Just add a `console.log` in `render` method in paragraph block to check the effect).

regischen:
> Thanks for the optimization! We should land this for now, since the perf regression on panning viewport and mouse move has been solved.
> 
> But there are still over-rendering in local editing, which can be addressed in following PRs.
> 
> Doc mode:
> 
>  Screen.Recording.2023-08-09.at.19.41.40.mov 
> Edgeless mode:
> 
>  Screen.Recording.2023-08-09.at.19.42.07.mov 
> (Just add a `console.log` in `render` method in paragraph block to check the effect).

I see. Hope this can be improved later

Issue #3918, opened 2023-08-09
Incorrect style of block hub


Garfield550:
<img width="123" alt="image" src="https://github.com/toeverything/blocksuite/assets/3471836/100c7414-a198-4ae3-9ac9-abfa78cc8ce3">

PR #3921, opened 2023-08-09
chore: use LitElement in WidgetElement

lawvs:
Related to https://github.com/toeverything/blocksuite/issues/3918

The `ShadowlessElement` will add the style to the global, causing unnoticed style leaks.


Flrande (approved):

lawvs:
In favor of https://github.com/toeverything/blocksuite/pull/3934

PR #3939, opened 2023-08-10
test: add test utils in window

Flrande:


packages/playground/apps/starter/env.d.ts
```diff
@@ -17,6 +18,7 @@ declare global {
     ContentParser: typeof ContentParser;
     Y: typeof Workspace.Y;
     std: typeof std;
+    testUtils: TestUtilsq;
```
doodlewind, line 21: Typo here

Flrande: #3944

PR #3947, opened 2023-08-10
fix: remain editing state when drag and drop in edgeless mode

donteatfriedrice:
https://github.com/toeverything/blocksuite/assets/99816898/e5acb89e-66e3-4d2f-b5cc-71f1bcc5dcdc


doodlewind (approved):
Cool fix, thanks!

Issue #3963, opened 2023-08-10
The `Table View` cell still hovers while not activated


yuusheng:
When moving out from a cell, the previously active cell is still staying in active style (except for the column type is `text`).
Is that a correct behavior?

![Kapture 2023-08-11 at 01 13 20](https://github.com/toeverything/blocksuite/assets/64511565/811727cc-9042-4916-8a6b-6afada3e4bc5)

zqran:
Thanks for the feedback, the problem has been fixed and cannot be reproduced in [Playground - database](https://blocksuite-toeverything.vercel.app/starter/?init=database).

PR #3968, opened 2023-08-11
fix: only the active cell shows the blue border

RajeevDash007:
#3963 
https://github.com/toeverything/blocksuite/assets/96860040/e1ae872a-22a5-457a-b5dd-9ced8a953375

Check it out!
if any further change is needed, then let me know.
@doodlewind @yuusheng

### DIFF
--- a/packages/blocks/src/database-block/common/columns/title/cell-renderer.ts
+++ b/packages/blocks/src/database-block/common/columns/title/cell-renderer.ts
@@ -1,7 +1,6 @@
 import type { TemplateResult } from 'lit';
 import { css } from 'lit';
-import { customElement, state } from 'lit/decorators.js';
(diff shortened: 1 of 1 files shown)


doodlewind:
Thanks for the trying! But see the CI log, directly deleting the logic makes the `should show or hide database toolbar` and `support drag and drop the add button to insert row` test cases failed. So this might not be a valid solution. This TDD workflow applies in our project, which prevents regression for us.

Also the lint job failed, checkout VSCode config to avoid code warning should help.

But regarding this feature, I think @zzj3720 is working on a better cell rendering strategy to avoid conflict like #3963, so awaiting for the more comprehensive refactoring (in #3790) should be the plan. Cheers.

Issue #3936, opened 2023-08-10
Import modal dialog window UI bugs (especially buttons)


xell:
We notice that there're still bugs in the https://github.com/toeverything/blocksuite/issues/3798 , especially the UI of button and perhaps fonts.

<img width="536" alt="CleanShot 2023-08-10 at 09 39 46@2x" src="https://github.com/toeverything/blocksuite/assets/132558/46096931-c80e-4e24-853d-e0066827ec4e">

v.s. design layout:

<img width="600" alt="CleanShot 2023-08-10 at 09 40 29@2x" src="https://github.com/toeverything/blocksuite/assets/132558/549fff0d-5c5e-4e95-bca5-6d296be5a80f">

RajeevDash007:
@xell can you please help me to get the import dialog box?

DiamondThree:
> @xell can you please help me to get the import dialog box?

hi, you can input "@" on the page and choose import. Like this:
![20230810231235](https://github.com/toeverything/blocksuite/assets/24630517/6a095ee6-ac14-487a-ac8e-4fee118e8bfc)

RajeevDash007:
Thank you for the assistance

RajeevDash007:
could you provide the figma design?

DiamondThree:
> 

https://www.figma.com/file/hVxJQtEqyv5xw1AbCzc50M/%F0%9F%97%92%EF%B8%8F-AFFiNE-2023?type=design&node-id=3104-67070&mode=design&t=aVDAtHenVzdUr2rP-0

xell:
> could you provide the figma design?
> It would help in replecating the UI.
> @xell @DiamondThree

Because the figma design files are not open sourced, please reply my email sent to you yesterday for further details. Thanks.

PR #3995, opened 2023-08-11
fix: updated the UI of the buttons of Import Dialog Box

RajeevDash007:
Fixed: #3936 
Screenshot (33)
The Font family is perfectly rendering in Brave browser as per the UI design,
but it varies from browser to browser ie. It gets changed when we see the same thing in Mozilla FireFox or Google chrome.

@xell check this out


packages/blocks/src/components/import-page/styles.ts
```diff
@@ -55,10 +55,11 @@ export const styles = css`
   .button-container icon-button {
     padding: 8px 12px;
     justify-content: flex-start;
-    gap: 8px;
+    gap: 12px;
     width: 190px;
     height: 40px;
     box-shadow: var(--affine-shadow-1);
+    border-radius:10px;
```
doodlewind, line 62: This fails the linter CI job again. The suggestion is to use VSCode with the [recommended extensions](https://github.com/toeverything/blocksuite/blob/master/.vscode/extensions.json) installed, so that you get proper lint warnings.

RajeevDash007: Thank you

RajeevDash007:
@doodlewind done

doodlewind (approved):
I checked roughly and the styling seems to be improved. /cc @xell if double check is required.

Note that we surely ensure the UI styling in Chrome and Electron, for other browsers, minor inconsistencies should be acceptable.

PR #4008, opened 2023-08-12
chore: check element before apply range

lawvs:


lawvs:
Closed due to its widespread use.

PR #4009, opened 2023-08-12
test: mark native test case as fail

lawvs:


tests/selection/native.spec.ts
```diff
@@ -752,41 +752,42 @@ test('should forwardDelete divider works properly', async ({ page }) => {
...
-  await assertRichTexts(page, ['123', '456', '789']);
-
-  const rect = await page.evaluate(() => {
-    const secondRichText = document.querySelector(
-      '[data-block-id="3"] .virgo-editor'
-    );
-    if (!secondRichText) {
-      throw new Error();
-    }
+test.fixme(
```
Saul-Mirone, line 755: test.fixme will just skip the case, I think maybe add test.fail is enough.

PR #4014, opened 2023-08-13
fix(edgeless): image jitters on resizing in edgeless mode

### LINKED ISSUE #4013: Image jitters on resizing in edgeless mode
https://github.com/toeverything/blocksuite/assets/17249775/fb1d4a2b-b529-448b-9592-e2a7e833469d

zhengjitf:
Close #4013 

As below after fixing:

https://github.com/toeverything/blocksuite/assets/17249775/91ce69a1-79fb-484c-82d1-cff15a499ab0

### DIFF
--- a/packages/blocks/src/image-block/image/image-resize-manager.ts
+++ b/packages/blocks/src/image-block/image/image-resize-manager.ts
@@ -1,23 +1,34 @@
 import type { PointerEventState } from '@blocksuite/block-std';
-import { assertExists, throttle } from '@blocksuite/global/utils';
+import { assertExists } from '@blocksuite/global/utils';
 
 import {
   type BlockComponentElement,
   getClosestBlockElementByElement,
   getModelByElement,
 } from '../../__internal__/utils/query.js';
+import { EdgelessPageBlockComponent } from '../../page-block/edgeless/edgeless-page-block.js';
+import { getClosestPageBlockComponent } from '../../page-block/utils/query.js';
 
 export class ImageResizeManager {
   private _activeComponent: BlockComponentElement | null = null;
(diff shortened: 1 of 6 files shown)


doodlewind (commented):
Thanks for the fix and this indeed solves jittering! However I found a regression here that occasionally cause the image being not draggable:

https://github.com/toeverything/blocksuite/assets/7312949/5b18d47d-3865-4120-b384-ac9f5a386726

In doc mode, this seems to be related to conflicts between block level selection and image dragging:

https://github.com/toeverything/blocksuite/assets/7312949/1d8d4128-1e54-40d0-992e-c9712058ab41

Any idea? /cc @Saul-Mirone for suggestion.

packages/blocks/src/image-block/image/image-resize-manager.ts
```diff
@@ -1,23 +1,34 @@
...
+  private _zoom = 1;
 
   onStart(e: PointerEventState) {
     const eventTarget = e.raw.target as HTMLElement;
     this._activeComponent = getClosestBlockElementByElement(
       eventTarget
     ) as BlockComponentElement;
+
+    const pageElement = getClosestPageBlockComponent(this._activeComponent);
+    if (pageElement instanceof EdgelessPageBlockComponent) {
```
Saul-Mirone, line 26: Don't use instance of to check the `pageElement` type. Please use tag.

doodlewind:
@zhengjitf how about try dragging image in doc mode? I think the reason could be the image block component being re-rendered during hover or mouse move, so its selection box state is cleared when you starts dragging.

lawvs (approved):
Great works!

doodlewind (approved):

PR #4027, opened 2023-08-14
fix(edgeless): presentation ui

regischen:
fix some presentation mode ui problems


doodlewind (commented):
Note that this commit log should not be `style`, which indicates to coding style (convention, lint, naming...) improvements, rather than UI stying (https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type).

regischen:
> https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type

ok get it

Issue #4055, opened 2023-08-14
No gap between 2 styles of Shapes


RajeevDash007:
![Screenshot (37)](https://github.com/toeverything/blocksuite/assets/96860040/5f14eaa6-a893-4025-89ef-df4019047d75)

PR #4057, opened 2023-08-14
fix(edgeless): add space between shapes buttons

RajeevDash007:
Fixed: #4055 

Before:
Screenshot (37)

After:
Screenshot (38)


doodlewind (approved):
Good catch, thanks!

Edit: seems that the spacing may still be a bit off from the [internal doc](https://www.figma.com/file/hVxJQtEqyv5xw1AbCzc50M/%F0%9F%97%92%EF%B8%8F-AFFiNE-2023?type=design&node-id=1-10&mode=design&t=rCncyLAeUwwGKqOG-0), could you tweak it according to the expected style?

RajeevDash007:
> Good catch, thanks!
> 
> Edit: seems that the spacing may still be a bit off from the [internal doc](https://www.figma.com/file/hVxJQtEqyv5xw1AbCzc50M/%F0%9F%97%92%EF%B8%8F-AFFiNE-2023?type=design&node-id=1-10&mode=design&t=rCncyLAeUwwGKqOG-0), could you tweak it according to the expected style?

Done!! changeed 10px to 8px as per the UI

RajeevDash007:
> Member

OK, let me fix it

RajeevDash007:
@doodlewind Fixed both of them, Check it and if you need more changes then let me know

doodlewind (changes requested):
The fixes so far doesn't seem to be very thoughtful. There are still three different height in three different menus, and the spacing between buttons are still not consistent. Are you sure this is what the design drafted?

https://github.com/toeverything/blocksuite/assets/7312949/41091a5b-07bc-4583-b028-5e14e41f25af

RajeevDash007:
https://github.com/toeverything/blocksuite/assets/96860040/b07dbd70-da3a-4087-9615-302042776fbc

I have checked the figma design, tried to replicate the UI, kindly check it out.
let me know, if something is wrong
@doodlewind

RajeevDash007 (commented):
Done

xell (changes requested):
There's obvious difference between the design and the implementation so far. For example, the height of the submenu of shapes should be 40px instead of 50px, and the container of the icon, 24px instead of 36px, among others.

Here's a overlay image with the greens showing the design.

CleanShot 2023-08-15 at 16 33 41@2x

The implementation:

CleanShot 2023-08-15 at 16 34 24@2x

versus the design in Figma:

CleanShot 2023-08-15 at 16 35 02@2x

RajeevDash007:
Okay I will look into it

doodlewind:
Closing in favor of #4139 since it includes the fix here and is more comprehensive. Thanks for the effort anyway!

Issue #4059, opened 2023-08-14
Incorrect style in the database "New Record" button


Zheaoli:
### What happened?

![Screenshot_604](https://github.com/toeverything/AFFiNE/assets/7054676/58b4db21-ea7e-4c76-a04b-3c11dac38a30)

I'm not sure whether the `New Record` display right or not

### Distribution version

Linux

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

doodlewind:
Thanks for the feedback, this looks awkward and should be fixed! /cc @zzj3720 @zqran

zqran:
How to reproduce?

doodlewind:
@zqran try playground?

<img width="900" alt="image" src="https://github.com/toeverything/blocksuite/assets/7312949/a19a3451-876f-4635-9844-97a04a08fc7b">

zqran:
<img width="1375" alt="image" src="https://github.com/toeverything/blocksuite/assets/15389209/9565aa86-c9f3-4290-a56e-2a865e41f8b5">

My display here is correct.

doodlewind:
Also reproduced in AFFiNE internal:

<img width="500" alt="image" src="https://github.com/toeverything/blocksuite/assets/7312949/022e5598-9550-4dd9-aa12-7f6966ae0f45">

@zqran also in your screenshot, the right padding of the button seems to be less the left, maybe we can just fix it as an UI issue.

PR #4067, opened 2023-08-15
fix(database): `New Record` UI

zqran:
close: https://github.com/toeverything/blocksuite/issues/4059


doodlewind (approved):
Works here, thanks!

PR #4105, opened 2023-08-15
feat(database): support block level copy & paste

zqran:


tests/clipboard.spec.ts
```diff
@@ -582,14 +582,16 @@ test('cut will delete all content, and copy will reappear content', async ({
   );
 });
 
-test.fixme(scoped`should copy and paste of database work`, async ({ page }) => {
+test(scoped`should copy and paste of database work`, async ({ page }) => {
```
zqran, line 585: This test is fixed.

@Saul-Mirone @zuoxiaodong0815

Saul-Mirone (approved):

Issue #4108, opened 2023-08-16
Incorrect corner border style in page level tag selector


RajeevDash007:
![Firefox_Screenshot_2023-08-15T16-42-39 287Z](https://github.com/toeverything/blocksuite/assets/96860040/e1f5d0d4-d5a5-4b53-8f35-7d1c2a54a8ad)

triangle shaped edge appears in the input selector top left and right.

PR #4111, opened 2023-08-16
fix: added border radius to page tag selector

RajeevDash007:
Fixed: #4108 

before:
Firefox_Screenshot_2023-08-15T16-42-39 287Z

after:
BlockSuite-Playground (2)


xell (approved):
It looks good.

PR #4119, opened 2023-08-16
feat(store): add snapshot v2 API

Saul-Mirone:


doodlewind (commented):
Why not just remove the legacy `importSnapshot`? This is not a public API for the moment.

Saul-Mirone:
> Why not just remove the legacy `importSnapshot`? This is not a public API for the moment.

I'm still polishing it and it seems that affine is using it:
https://github.com/toeverything/AFFiNE/blob/08f01ea1b31082bf7572bc80d82a4fdbebc8da44/packages/infra/src/blocksuite/index.ts#L186

PR #4123, opened 2023-08-16
fix: unable to copy and paste under multi lines of text selection

### LINKED ISSUE #4120: Unable to copy and paste under multi lines of text selection
To reproduce, each of the line should be a paragraph block, not list:

image

zuoxiaodong0815:
close #4120


doodlewind (commented):
I think we need a test for this kind of basic case 🫡

doodlewind (approved):

PR #4126, opened 2023-08-16
fix(page): range should not be changed by remote-selection

Flrande:


packages/lit/src/utils/range-manager.ts
```diff
@@ -192,6 +196,29 @@ export class RangeManager {
       `Cannot find virgo element in block ${point.path.join(' > ')}}`
     );
 
+    const maxLength = startVirgoElement.virgoEditor.yText.length;
```
Saul-Mirone, line 199: Can we extract the index and length calculation to another method?

Flrande: I will create other pr

Flrande: https://github.com/toeverything/blocksuite/pull/4129

PR #4181, opened 2023-08-17
feat(page): add left right hotkey

Saul-Mirone:


doodlewind (commented):
I guess adding a test here could be better?

Saul-Mirone:
> I guess adding a test here could be better?

Done

PR #4186, opened 2023-08-17
feat(virgo): make `setVRange` support null

Flrande:


packages/virgo/src/services/range.ts
```diff
@@ -24,6 +24,18 @@ export class VirgoRangeService<TextAttributes extends BaseTextAttributes> {
   onVRangeUpdated = ([newVRange, origin]: VRangeUpdatedProp) => {
     this._vRange = newVRange;
 
+    if (this._vRange === null) {
```
Saul-Mirone, line 27: Maybe put if after origin type check? This may cause infinite loop with range manager.

Flrande: How it cause?🤔

Saul-Mirone: Just want to make sure the `silent` origin works as expect.

Saul-Mirone (approved):

PR #4187, opened 2023-08-17
fix: optimize hasChanged for block element

lawvs:
The `widgets` and `content` will cause a block re-render even if nothing has changed.

https://github.com/toeverything/blocksuite/blob/3bbe6d9ebd36e8e230ed72c47233cb4b72ffca8f/packages/lit/src/element/lit-root.ts#L143-L150

The pointless re-render is the root cause of the issue at https://github.com/toeverything/blocksuite/issues/4013

> In child expressions, `undefined`, `null`, `''`, and `nothing` all behave the same and render no nodes.

> hasChanged: A function that indicates if a property should be considered changed when it is set. The function should take the `newValue` and `oldValue` and return true if an update should be requested.

### DIFF
--- a/packages/lit/src/element/lit-root.ts
+++ b/packages/lit/src/element/lit-root.ts
@@ -9,7 +9,7 @@ import type {
 import { BlockStore } from '@blocksuite/block-std';
 import { assertExists } from '@blocksuite/global/utils';
 import type { BaseBlockModel, Page } from '@blocksuite/store';
-import type { PropertyValues, TemplateResult } from 'lit';
+import { nothing, type PropertyValues, type TemplateResult } from 'lit';
 import { customElement, property } from 'lit/decorators.js';
 import { repeat } from 'lit/directives/repeat.js';
 import type { StaticValue } from 'lit/static-html.js';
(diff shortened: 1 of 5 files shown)


packages/lit/src/element/lit-root.ts
```diff
@@ -103,22 +102,22 @@ export class BlockSuiteRoot extends WithDisposable(ShadowlessElement) {
...
   }
 
-  renderModel = (model: BaseBlockModel): TemplateResult => {
+  renderModel = (model: BaseBlockModel): TemplateResult | null => {
     const { flavour, children } = model;
     const schema = this.page.schema.flavourSchemaMap.get(flavour);
     if (!schema) {
       console.warn(`Cannot find schema for ${flavour}.`);
-      return html`${nothing}`;
+      return null;
```
Saul-Mirone, line 110: Maybe we should keep `nothing`[^1] here.

[^1]: [`nothing` vs `null`](https://github.com/lit/lit/issues/784#issuecomment-460086850)

packages/blocks/src/image-block/image-block.ts
```diff
@@ -481,6 +487,10 @@ export class ImageBlockComponent extends BlockElement<ImageBlockModel> {
     return ImageSelectedRectsContainer();
   }
 
+  override willUpdate(changedProperties: PropertyValues) {
+    console.log('will update image block', changedProperties);
```
Saul-Mirone, line 491: console.log

packages/blocks/src/image-block/image-block.ts
```diff
@@ -491,6 +501,7 @@ export class ImageBlockComponent extends BlockElement<ImageBlockModel> {
       resizeImgStyle.width = `${width}px`;
       resizeImgStyle.height = `${height}px`;
     }
+    console.log('render image block', this.model);
```
Saul-Mirone, line 504: console.log

Saul-Mirone (approved):

PR #4193, opened 2023-08-17
feat(database): support the copy and paste of the row selection

zqran:
https://github.com/toeverything/blocksuite/assets/15389209/16539c3d-00d9-4060-8c5f-649a7d0efbf3


packages/blocks/src/database-block/table/clipboard.ts
```diff
@@ -49,7 +47,12 @@ export class TableViewClipboard implements BaseViewClipboard {
       uiEventDispatcher.add(
         'copy',
         ctx => {
+          const selection = getDatabaseSelection(this._root);
+          const tableSelection = selection?.getSelection('table');
+          if (!tableSelection) return false;
+
           this._onCopy(ctx);
```
zqran, line 54: @Saul-Mirone 

This event handler contains an asynchronous operation.
In the current scenario, if it is a copy inside the database, `return true` can be used to prevent bubbling.

doodlewind (approved):

PR #4194, opened 2023-08-17
fix(edgeless): make note submenu consistent with other submenu

### LINKED ISSUE #4070: Inconsistent scroll/expand effects in Edgeless
When using the Pen tool - the expand button will move the view to show the extra colours.

When using the Note tool - the expand button will move individually through the options. 

I think these two interactions should be the same/similar. (Personally prefer the colour interaction in pen tool)

donteatfriedrice:
Close #4070


doodlewind (approved):

xell (approved):
It looks good.

PR #4195, opened 2023-08-17
feat(edgeless): slicer should remain its size when zooming

### LINKED ISSUE #4116: Block slicer size should not follow zooming

doouding:
close #4116


doodlewind (approved):
Good to see the naming change together!

Issue #3836, opened 2023-08-04
Text `foo.hello` wrongly identified as link


zanwei:
### **Description**

**Now:** When inputting Heading content, if a user types '1.' and continues typing without a space, the text is displayed in link style

**Should:** Keep input style with Heading

**Notice:** The situation is repeated in all Heading styles

https://github.com/toeverything/AFFiNE/assets/40333372/ea8150d4-c14b-4008-acbc-5d70ea9b4206

### Distribution version

macOS ARM 64 (Apple Silicon)

### What browsers are you seeing the problem on if you're using web version?

Chrome

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

fourdim:
The thing is you have to add a space.
1.test is a valid link as you can see it here: https://en.wikipedia.org/wiki/.test
Many links like `release.zip` are valid domains which will redirect you to Rick Roll.
Due to technical reasons, we cannot match all the [TLD](https://en.wikipedia.org/wiki/Top-level_domain)s concerning that it will take a long time to determine whether it needs to be a link as the TLD list is too long.

See also:
https://medium.com/@bobbyrsec/the-dangers-of-googles-zip-tld-5e1e675e59a5

zanwei:
Got it, then another issue occurs, please check the video

https://github.com/toeverything/blocksuite/assets/40333372/cb5c8dd7-99f4-4d82-bec6-085d79a53baa

fourdim:
> Got it, then another issue occurs, please check the video
> bug.mov

Fixed in #3875.

doodlewind:
I think the link identify condition looks too permissive, basically it turns any `[word].[word]` into link.

May relate to `autoIdentifyLink` introduced together with bookmark block, /cc @QiShaoXuan

QiShaoXuan:
> I think the link identify condition looks too permissive, basically it turns any `[word].[word]` into link.
> 
> May relate to `autoIdentifyLink` introduced together with bookmark block, /cc @QiShaoXuan

Get, based on current user feedback, the link check should only verify whether the character ending is .com, .cn, .org, and the like.

PR #4202, opened 2023-08-17
fix(page): strict check when identity link

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/3836

The `autoIdentifyLink` still has **many** bugs to handle.

### DIFF
--- a/packages/blocks/src/__internal__/utils/url.ts
+++ b/packages/blocks/src/__internal__/utils/url.ts
@@ -46,7 +46,8 @@ const URL_REGEX = new RegExp(
     '[a-z0-9\\u00a1-\\uffff]\\.' +
     ')+' +
     // TLD identifier name, may end with dot
-    '(?:[a-z\\u00a1-\\uffff]{2,}\\.?)' +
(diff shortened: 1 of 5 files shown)


doodlewind (approved):
This naming is great, we should use "isStrict" policy for auto identification, instead of "isValid" policy that could be overkill. Great work!

packages/blocks/src/__internal__/utils/url.ts
```diff
@@ -80,7 +84,58 @@ export const isValidUrl = (str: string) => {
...
+    } catch {
+      return false;
+    }
     return true;
   }
   return URL_REGEX.test(url);
 };
+
+// https://en.wikipedia.org/wiki/Top-level_domain
+const COMMON_TLDS = [
```
xell, line 98: Perhaps we could add more domains here as listed by some statistics like [1] or [2].

[1] https://dnsinstitute.com/research/popular-tld-rank/
[2] https://w3techs.com/technologies/overview/top_level_domain

doodlewind: I think the "auto identify" feature should be *conservative*, since user can always mark a piece of text as link.

Issue #4164, opened 2023-08-17
The "add new record" button in database should not add to first row


Yanxi0202:
https://github.com/toeverything/blocksuite/assets/134897361/7ee67b63-b0c3-4024-b3d0-33488760bf6e

Yanxi0202:
unable to meet user expectations: if you click the add new record, it should be at the bottom line.

doodlewind:
> unable to meet user expectations: if you click the add new record, it should be at the bottom line.

Yes, and I think if a cell or row is selected, clicking this button should inserted the new row after the selected one. /cc @houjoe0829 to confirm.

zqran:
What should I do if multiple cells are selected?

<img width="824" alt="image" src="https://github.com/toeverything/blocksuite/assets/15389209/78ab68fc-9bb1-486e-bb85-39d2a846ca06">

xell:
> What should I do if multiple cells are selected?
> 
> <img alt="image" width="824" src="https://user-images.githubusercontent.com/15389209/261469448-78ab68fc-9bb1-486e-bb85-39d2a846ca06.png">

I think the new row could just be right after the last row of the selection.

PR #4204, opened 2023-08-18
feat(database): add row according to the selection

zqran:
close: https://github.com/toeverything/blocksuite/issues/4164

https://github.com/toeverything/blocksuite/assets/15389209/8608de74-e7d3-406e-a461-93cccd4bbb9f

### DIFF
--- a/packages/blocks/src/database-block/common/data-view.ts
+++ b/packages/blocks/src/database-block/common/data-view.ts
@@ -7,6 +7,7 @@ import type { UniComponent } from '../../components/uni-component/uni-component.
 import type { DatabaseBlockModel } from '../database-model.js';
 import type { InsertPosition } from '../types.js';
 import type { DataViewManager } from './data-view-manager.js';
+import type { DatabaseSelection } from './selection.js';
 
 export interface DataViewProps<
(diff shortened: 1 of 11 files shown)


doodlewind (approved):
Finally we have this, thanks!

packages/blocks/src/database-block/database-block.ts
```diff
@@ -272,11 +272,18 @@ export class DatabaseBlockComponent extends BlockElement<DatabaseBlockModel> {
         models.forEach(model => this.page.deleteBlock(model));
       },
     };
+    const getSelection = () => {
+      return this.root.selectionManager.value.find(selection =>
+        selection.is('database')
```
zzj3720, line 277: use Pathfinder to make sure the selection is in current block

packages/blocks/src/database-block/common/data-view.ts
```diff
@@ -24,13 +25,15 @@ export interface DataViewProps<
 
   setSelection: (selection?: Selection) => void;
 
+  getSelection: () => DatabaseSelection | undefined;
```
zzj3720, line 28: DataView should not depend on database block.

Issue #4075, opened 2023-08-15
UI issues when editing database row properties


ShortCipher5:
- Close `X` does not close the window
- BlockHub and Help icons are overlayed (but heading area, such as page title is not)

![Image](https://github.com/toeverything/blocksuite/assets/4605025/417d8fb4-0a8c-495e-8b1a-374351531296)

pengx17:
`buttons` in AFFiNE may be covered by Electron's draggable region and thus button tag is added with 'no-drag' attribute. I think you may want to use standard button tag for clickable elements, or add role="button". It is not sepcific to solving this particular issue, but could also provide better a11y for the users.

ShortCipher5:
Just to clarify "window" refers to the sidebar window - rather than the OS (Windows).

zqran:
> - Close X does not close the window

Fixed by `0.8.0-canary.25`.  Related to https://github.com/toeverything/AFFiNE/pull/3783/commits/df0307813992841f37c736e0f1a723e5c1eb0741

PR #4208, opened 2023-08-18
fix(database): ui issues when editing database row properties

zqran:
close: https://github.com/toeverything/blocksuite/issues/4075

### DIFF
--- a/packages/blocks/src/components/menu/menu.ts
+++ b/packages/blocks/src/components/menu/menu.ts
@@ -498,7 +498,7 @@ export const createModal = () => {
   div.style.top = '0';
   div.style.width = '100vw';
   div.style.height = '100vh';
-  div.style.zIndex = '10';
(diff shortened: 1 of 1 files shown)


packages/blocks/src/components/menu/menu.ts
```diff
@@ -498,7 +498,7 @@ export const createModal = () => {
   div.style.top = '0';
   div.style.width = '100vw';
   div.style.height = '100vh';
-  div.style.zIndex = '10';
+  div.style.zIndex = '1001';
```
doodlewind, line 501: 😂😂😂

doodlewind (approved):

packages/blocks/src/components/menu/menu.ts
zqran: Because the `z-index` of the block-hub given by Affine is 1000.

Issue #3611, opened 2023-07-19
Pasted database has different column sequence


Yanxi0202:
​

https://github.com/toeverything/blocksuite/assets/134897361/9f4fb2e5-d0aa-481a-a181-80220d0972b1

doodlewind:
I can still reproduce this and requires @zqran to take a look. Seems that the column data are preserved, but not the column sequence.

zqran:
I'm working on it.

PR #4240, opened 2023-08-21
fix(database): preserve column order when copying

zqran:
close: https://github.com/toeverything/blocksuite/issues/3611

https://github.com/toeverything/blocksuite/assets/15389209/e7ff4c10-7a0c-4114-a97f-f52cf6197e6f


doodlewind (approved):

doodlewind (commented):
This seems to be an easy one to be guarded by test case. Hope that we can add test cases about database copying and pasting in following PRs.

PR #4244, opened 2023-08-21
fix(edgeless): height of note collapse

### LINKED ISSUE #4166: Unstable edgeless note height on switch between editor modes
This happens more frequently for a page with many image-included notes.

doouding:
close #4166 

### Notable change
 - The note's height is not explicitly set when rendering in edgeless, it is directly determined by its content
 - The note's height is now unidirectionally synchronized to its 'xywh' property, which means that changing the height will not affect the actual rendered height of the note (perhaps we should introduce a new flag or property to indicate the height is manually changed by user)
 - Border width is taken into account when synchronize height of note to its 'xywh' property.


doodlewind (approved):
Good catch, thanks for diving into this!

PR #4251, opened 2023-08-21
feat(edgeless): enable toolbar actions for frame

### LINKED ISSUE #4225: The menu of the Frame does not have the expected effect
Current
CleanShot 2023-08-18 at 17 43 09@2x

Expected effect：
Refer to [internal doc](https://toeverything.feishu.cn/docx/JNm5dZ7n2oeH60xUCfXcxMD6nHh#ZVVMdDvaPo6xi0xag2xcUmWmnFb)

regischen:
close #4225

### DIFF
--- a/packages/blocks/src/page-block/edgeless/utils/general.ts
+++ b/packages/blocks/src/page-block/edgeless/utils/general.ts
@@ -1,0 +1,88 @@
+import { assertExists } from '@blocksuite/global/utils';
+import {
+  Bound,
+  FrameElement,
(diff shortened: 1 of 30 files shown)


packages/blocks/src/page-block/edgeless/utils/general.ts
doodlewind, line 1: Having a `general` module here is too opaque. Why not put most of the methods here as clipboard utils? The `edgelessElementsBound` and `getGridBound` should be elsewhere.

Issue #4248, opened 2023-08-21
Image popup overlaps with import modal


doodlewind:
<img width="769" alt="image" src="https://github.com/toeverything/blocksuite/assets/7312949/ad0280a6-93af-4e9d-b2f5-70f005948807">

PR #4260, opened 2023-08-21
fix(page): image format bar overlaps import modal

RajeevDash007:
fixed: #4248 

https://github.com/toeverything/blocksuite/assets/96860040/9ccf4cc9-98b0-4426-ab9f-04ceaa705505


doodlewind (changes requested):
https://www.codegram.com/blog/stop-ab-using-z-index/

lawvs:
I believe we just need to add a simple `overlay mask` to prevent any hover operation in the editor.

RajeevDash007:
> I believe we just need to add a simple `overlay mask` to prevent any hover operation in the editor.

 I guess, still we have to give the z-index value more than 1000 as most of the z-index are set to 1000

lawvs:
> I guess, still we have to give the z-index value more than 1000 as most of the z-index are set to 1000

z-index only competes with sibling elements within the same [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context).

Screenshot 2023-08-24 at 14 29 36

lawvs (approved):

doodlewind (approved):

Issue #4183, opened 2023-08-17
The document sequence of bookmark unexpectedly changed after adding caption


Yanxi0202:

Yanxi0202:
Detailed video can not be uploaded, because of the length limited, will direct send to the responsible people through Feishu.

doodlewind:
If you are referring to this behavior, I guess it should be by design:

https://github.com/toeverything/blocksuite/assets/7312949/0acec876-7e78-4128-864a-91d0bbea9df7

doodlewind:
Closing since this is not reproduced in latest BlockSuite playground. If anyone encounters this, please let us know.

PR #4261, opened 2023-08-21
fix(page): click at blank area should not change note sequence

donteatfriedrice:
Before: every time click blank area to add a tail paragraph will reverse the page model.children

https://github.com/toeverything/blocksuite/assets/99816898/8b5599dc-2386-463c-af2f-869ff3ce8a91

After:

https://github.com/toeverything/blocksuite/assets/99816898/5fc4413a-54e0-4a05-83e4-2eba6db86ff1


doodlewind (approved):
So this should actually fix #4183?

donteatfriedrice:
> So this should actually fix #4183?

I guess it should fix that issue

PR #4269, opened 2023-08-22
feat(edgeless): new component more button

### LINKED ISSUE #3780: Edgeless UI bugs
There're several inconsistencies with the Figma design:

- [x] The hover background of page "Hidden" is incorrect, which should be the same as that of other buttons, for example, the "More" button.

Untitled

- [x] The activated state of the Pen tool is inconsistent with the design (left)

Untitled

- [x] The font size of zoom percent should be scaled downward a tier, i.e. from `sm` 14px to `xs` 12px. And it would be better to use css variable instead of hard coding number to avoid future changes. 

CleanShot 2023-08-01 at 15 24 32@2x

- [ ] The "More" menu needs polished (left is the design layout with all the icons). By the way, the font family issue will be addressed in another issue.

Untitled

regischen:
partly closes #3780 
image


Svaney-ssman (changes requested):
1. wrong frame icon color. correct color: icon color
image

2. list item should not be wrap
image

Svaney-ssman (changes requested):
font size/ menu width incorrect
image
check [design](https://www.figma.com/file/hVxJQtEqyv5xw1AbCzc50M/%F0%9F%97%92%EF%B8%8F-AFFiNE-2023?type=design&node-id=12832-92266&mode=design&t=DozPvXqgRMj6tugA-4)
image

Svaney-ssman (changes requested):
icon size 20*20
image

regischen:
> icon size 20*20 image

image
which version do we use?

Svaney-ssman:
> > icon size 20*20 image
> 
> image which version do we use? so we should not use the icon from the library?

You need to adjust the size if you collect icons from the icon library (all icons are 24*24) . In this menu, we use 20*20.

Svaney-ssman (approved):
LGTM

Issue #4172, opened 2023-08-17
E2E tests cases fail locally


dengquan:
xx@localhost blocksuite % pnpm --version
8.6.7
xx@localhost blocksuite % node --version
v18.17.1
xx@localhost blocksuite % npx playwright --version
Version 1.36.2

When I run E2E tests by the command `pnpm test`, there are 26 failed. Is this a problem with my environment or with the code?

```
26 failed
    bookmark.spec.ts:191:1 › copy url to create bookmark in edgeless mode ──────────────────────────
    clipboard.spec.ts:67:1 › clipboard copy paste ──────────────────────────────────────────────────
    clipboard.spec.ts:374:1 › copy partially selected text ─────────────────────────────────────────
    clipboard.spec.ts:395:1 › copy & paste outside editor ──────────────────────────────────────────
    clipboard.spec.ts:526:1 › cut should work for multi-block selection ────────────────────────────
    clipboard.spec.ts:547:1 › pasting into empty list should not convert the list into paragraph ───
    clipboard.spec.ts:763:1 › copy when text note active in edgeless ───────────────────────────────
    clipboard.spec.ts:805:1 › copy and paste to selection block selection ──────────────────────────
    clipboard.spec.ts:822:1 › should keep paragraph block's type when pasting at the start of empty paragraph block except type text 
    code.spec.ts:253:1 › drag copy paste ───────────────────────────────────────────────────────────
    code.spec.ts:269:1 › keyboard selection and copy paste ─────────────────────────────────────────
    code.spec.ts:325:1 › use code block copy menu of code block copy whole code block ──────────────
    code.spec.ts:370:1 › code block copy button can work ───────────────────────────────────────────
    edgeless/basic.spec.ts:113:1 › option/alt mouse drag duplicate a new element ───────────────────
    edgeless/note.spec.ts:173:1 › edgeless arrow up/down ───────────────────────────────────────────
    edgeless/note.spec.ts:295:1 › drag handle should work across multiple notes ────────────────────
    edgeless/shape.spec.ts:348:1 › dbclick to add text in shape ────────────────────────────────────
    edgeless/shape.spec.ts:385:1 › auto wrap text in shape ─────────────────────────────────────────
    edgeless/text.spec.ts:56:1 › add text element in text mode ─────────────────────────────────────
    edgeless/text.spec.ts:89:1 › copy and paste ────────────────────────────────────────────────────
[...]

PR #4278, opened 2023-08-22
feat(page): keyboard operation to move block

### LINKED ISSUE #1686: Support keyboard operation to move Block up or down
This is a supplement to a very small editing operation, with the aim of ensuring the functional scope of mouse and keyboard operations are as equal as possible.

(1) Shortcut key format
⌘/ctrl + ⌥/Alt + ↑  Move Up
⌘/ctrl + ⌥/Alt + ↓  Move Down

(2) Slash Commands
Add Move Up and Move Down to the Action in the Slash Commands, and place these two options at the top of the Action area.

(3) Add corresponding explanations in the shortcut key instructions
This is located at the end of Keyboard Shortcuts, with attention given to distinguishing between Mac and Windows.
CleanShot 2023-03-13 at 6 20 35@2x

Additionally, the cursor remains at the original position of the manipulated Block after it has been moved.

lzlme:
resolve #1686


packages/blocks/src/page-block/const/move-block-config.ts
```diff
@@ -0,0 +1,92 @@
...
+        const prevBlock = getPrevBlock(block);
+        if (!prevBlock) return;
+        const parent = blockElement.page.getParent(prevBlock.model);
+        if (!parent || parent.flavour === 'affine:page') return;
+        blockElement.page.moveBlocks(
+          [block.model],
+          parent,
+          prevBlock.model,
+          true
+        );
```
Saul-Mirone, line 51: You should return `true` if the operation success to prevent other handlers for the same key map.

lzlme: OK, I've done.

lzlme:
I don't know why the test case failed, but my local test passed normally.

doodlewind:
> I don't know why the test case failed, but my local test passed normally.

@lzlme this looks like the reverse of #4172. You can checkout CI screen recording of failed test cases following this guide:

https://github.com/toeverything/blocksuite/assets/7312949/b489dd2e-df9a-4681-aab6-2cc373459f45

Saul-Mirone (approved):

PR #4279, opened 2023-08-22
feat(virgo): add keyboard util

Flrande:


packages/playground/examples/virgo/test-page.ts
```diff
@@ -1,20 +1,24 @@
...
 } from '@blocksuite/virgo';
-import { css, html } from 'lit';
-import { customElement, query } from 'lit/decorators.js';
+import { css, html, nothing } from 'lit';
+import { customElement, property, query } from 'lit/decorators.js';
 import { styleMap } from 'lit/directives/style-map.js';
 import * as Y from 'yjs';
 import { z } from 'zod';
 
+import { markdownMatches } from './markdown';
```
doodlewind, line 20: Requires `.js` suffix here

doodlewind (commented):
What's the purpose of adding keyboard APIs in Virgo? Any related PRs?

Flrande:
> What's the purpose of adding keyboard APIs in Virgo? Any related PRs?

Will be used to replaced original [keyboard](https://github.com/toeverything/blocksuite/blob/848cdbec890c296e15ca072eafc01dbf9207eecc/packages/blocks/src/__internal__/rich-text/keyboard.ts) (old api is based on block model) and make `rich-text` can be used in database rich-text cell as a general component.

Actually it is a step to refactor `rich-text`. After all finished `rich-text` will be based on yText rather than on block model. It means that it shouldn't contain block-level logic itself (such as hard enters, indent, etc.), and should only contain inline style logic.

Issue #4165, opened 2023-08-17
Canvas text layout shifts when panning to viewport edge


doodlewind:
https://github.com/toeverything/blocksuite/assets/7312949/a5305e30-05a4-444d-8de4-5ab3310777de

Handed over to @regischen but hope @Flrande could provide help if needed.

regischen:
seems it didn't occur

doodlewind:
Closed by #4281

PR #4281, opened 2023-08-22
fix(edgeless): canvas text layout shifts when panning to viewport edge

LittleSound:
fix issue: #4165


doodlewind (approved):

Issue #4271, opened 2023-08-22
The name and size should be preserved when converting an image to a attachment block in AFFiNE


lawvs:

lawvs:
Fixed in https://github.com/toeverything/blocksuite/pull/4282

PR #4282, opened 2023-08-22
fix: retention status between images and attachments

LittleSound:
fix issue: #4271


lawvs:
Thanks for catching this! But placing all miscellaneous data into a block scheme is not a good idea for several reasons:

- It will make type-checking unreliable. The growing amount of empty data will affect the development experience.
- They are mostly useless in most scenarios.
- It can also cause unexpected privacy leaks.

After further discussion, we have decided to only save temporary data in memory. This means that if you refresh the page, the data carried over from the previous block will be lost.

I will update your pull request to reflect this change. Thank you once again for your contribution.

PR #4288, opened 2023-08-22
fix: editor doesn't use block presets from property

Vija02:
`editor-container` accepts the `pagePreset` and `edgelessPreset` property. However, they are not actually used. `block-suite-root` is instead given the default preset.

This is fixed by simply pointing to the correct reference.


lawvs (approved):
Nice catch！

doodlewind (approved):
Good catch, thanks for your contribution!

PR #4292, opened 2023-08-22
feat: expose `refreshPageIndex` API

himself65:


himself65:
@Saul-Mirone This is just a hotfix for https://github.com/toeverything/blocksuite/issues/4294

PR #4295, opened 2023-08-22
fix: search might not work

himself65:
Fixes: https://github.com/toeverything/blocksuite/issues/4294

### DIFF
--- a/packages/store/src/indexer/search.ts
+++ b/packages/store/src/indexer/search.ts
@@ -112,9 +112,11 @@ export class SearchIndexer {
     if (!(yBlocks instanceof YMap)) {
       return;
     }
-    yBlocks.forEach((_, key) => {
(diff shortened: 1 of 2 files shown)


packages/store/src/indexer/search.ts
```diff
@@ -112,17 +112,19 @@ export class SearchIndexer {
...
+        this._refreshIndex(pageId, key, 'add', yBlocks.get(key));
+      });
+    }
   }
 
   private _handlePageIndexing(pageId: string, page?: Doc) {
     if (!page) {
       return;
     }
     const yBlocks = page.get('blocks');
```
dmonad, line 126: @himself65  You can simply do `page.get('blocks', Y.Map)` and always retrieve a Y.Map (even if another user uses this as a `Y.Text`).

himself65: See https://github.com/toeverything/blocksuite/pull/4296

PR #4317, opened 2023-08-23
fix: should not sync range to text-selection when it is out of note

### LINKED ISSUE #4305: Cursor jump to note when typing in canvas text or frame title
https://github.com/toeverything/blocksuite/assets/58546692/acad5104-2f1a-4b8a-aec4-87ef28a7b721

https://github.com/toeverything/blocksuite/assets/58546692/fba0feaa-7aee-4c76-bf27-4ccf37477613

img_v2_07d22616-aa26-49f7-94a7-922845549a9g

Flrande:
Close #4305

### DIFF
--- a/packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
+++ b/packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
@@ -1264,7 +1264,7 @@ export class EdgelessPageBlockComponent
       if (!surface) return;
 
       const el = this.surface.pickById(surface.elements[0]);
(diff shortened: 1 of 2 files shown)


Saul-Mirone (changes requested):
This seems not right. A module in lit should not relay on an element name. Instead of check the tag name, a better way might be:
1. Check the data attribute. For example `[data-sync-range="true"]'.
2. Make the elements should be synced as an option/config of this module.

Saul-Mirone (approved):

Issue #4308, opened 2023-08-23
After copied, drawings and notes can't move or delete together at once


Yanxi0202:
https://github.com/toeverything/blocksuite/assets/134897361/2b8791b6-4aad-436e-bf30-cfd7b0853650

you need manually refresh the page to see the changes, for more details please see the video.

regischen:
it seems werid that you don't select the text, In the video I saw the selected box is boxing the textElement.
By the way which version do you use @Yanxi0202

Yanxi0202:
I use this version: 0.0.0-internal.20230823003802-9d6d559, I can only reproduce it in the same workspace. I tried to add a new workspace, and it can not reproduce it. 
<img width="207" alt="image" src="https://github.com/toeverything/blocksuite/assets/134897361/ed2c15ed-14e9-4d9a-8e18-b4a636d4f4d5">

doouding:
Can you export your workspace data so that I can try to reproduce it.

Yanxi0202:
sure, will send it to your feishu account

doouding:
Reproduced after copy and paste operation.

PR #4320, opened 2023-08-23
fix(edgeless): rebuild connector path and relation after paste or undo operation

### LINKED ISSUE #4312: After deleting contents in Edgeless, the result of undo operation will temporarily loss some elements.
https://user-images.githubusercontent.com/142964904/262606385-e14c93a7-4124-490e-b3a8-0aa09ab66d78.mp4

doouding:
close #4312 #4308

### DIFF
--- a/packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
+++ b/packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
@@ -375,6 +375,15 @@ export class EdgelessPageBlockComponent
         const element = this.surface.pickById(id);
         assertExists(element);
         if (element instanceof ConnectorElement) {
+          const { source, target } = element;
+
+          if (
+            (source.id && !getEdgelessElement(this, source.id)) ||
+            (target.id && !getEdgelessElement(this, target.id))
+          ) {
+            return;
+          }
+
           this.connector.updatePath(element);
         } else if (element instanceof FrameElement) {
           this.frame.calculateFrameColor(element);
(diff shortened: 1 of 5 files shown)


doodlewind (approved):

packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
```diff
@@ -375,6 +375,10 @@ export class EdgelessPageBlockComponent
         const element = this.surface.pickById(id);
         assertExists(element);
         if (element instanceof ConnectorElement) {
+          if (!this.connector.hasRelatedElement(element)) {
+            return;
+          }
```
regischen, line 380: If just return, we need at some moment to update the path.

regischen: connector will not show if we delete elements connectored to connector then undo

doouding: The note element added afterward will trigger a connector path update.  Not sure if there's another scenario.

regischen: https://github.com/toeverything/blocksuite/assets/58546692/c4109cc2-2eff-441f-b608-de9e93279a80

I think connector path update may better not trigged by other element.

packages/phasor/src/elements/connector/connector-element.ts
```diff
@@ -120,6 +120,10 @@ export class ConnectorElement extends SurfaceElement<IConnector> {
       this.strokeStyle === StrokeStyle.Dashed
     );
 
+    // points might not be build yet in some senarios
+    // eg. undo/redo, copy/paste
+    if (!points.length) return;
```
regischen, line 125: I think we should make sure connector points must exists, if not , there must something wrong.

packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
doouding: It's kind of like calling `connector.syncConnectorPos([newElement])` when new element is added.

regischen: alright

PR #4321, opened 2023-08-23
style(edgeless): selected rect border radius

doouding:


doodlewind:
Note that the `style` type in commit message indicates changes related to coding styles, not UI styling. So this should be just a `fix`.

doouding:
> Note that the `style` type in commit message indicates changes related to coding styles, not UI styling. So this should be just a `fix`.

TIL

PR #4330, opened 2023-08-24
fix(edgeless): correctly display hover style on moving with arrow key

luke358:
https://github.com/toeverything/blocksuite/assets/48149577/59087554-5ecd-40a2-8384-dc456e646a61


doodlewind (approved):
Good catch, thanks!

Issue #4352, opened 2023-08-25
fix: the database double link is error after importing notion


zuoxiaodong0815:
show the right after click
![34499b89-f425-41b6-8daf-84be831f2d73](https://github.com/toeverything/blocksuite/assets/53252747/1d653c93-f6c9-4f31-866f-f2fb538d74d7)

PR #4353, opened 2023-08-25
fix(database): transform page link in title cell

zzj3720:
close https://github.com/toeverything/blocksuite/issues/4352


fourdim (approved):
The order matters. :)

PR #4372, opened 2023-08-25
feat(edgeless): fix wrong connector data

regischen:


packages/store/src/__tests__/migration.unit.spec.ts
```diff
@@ -93,6 +95,34 @@ describe('block migration', () => {
     assert.equal(shape.get('italic'), true);
   });
 
+  test.only('fix wrong connector data', async () => {
```
Saul-Mirone, line 98: test.only

Flrande (approved):

PR #4373, opened 2023-08-25
fix(page): unable to insert bookmark into list

BangKk:
Close: #3614


packages/blocks/src/bookmark-block/bookmark-model.ts
```diff
@@ -30,7 +30,7 @@ export const BookmarkBlockSchema = defineBlockSchema({
   metadata: {
     version: 1,
     role: 'content',
-    parent: ['affine:note'],
+    parent: ['affine:note', 'affine:list'],
```
doodlewind, line 33: Yes this indeed allows the nesting from model side. But here we may also require some effort to confirm if styling and editing functionalities could work well in this case.

BangKk:
This PR took so long, so I closed it.

PR #4409, opened 2023-08-29
feat(database): support copy&paste when editing kanban cell

zqran:
close: #4393


packages/blocks/src/database-block/table/clipboard.ts
```diff
@@ -459,7 +339,7 @@ function pasteToCells(
   const srcRowLength = srcColumns.length;
   const srcColumnLength = srcColumns[0].length;
 
-  model.page.captureSync();
+  data.captureSync();
```
zzj3720, line 342: Can we remove this deprecated function call

zzj3720 (approved):

PR #4416, opened 2023-08-29
fix(edgeless): hide overlay when pointer move on toolbar

donteatfriedrice:
Before:
When using shape or note tool in edgeless page, the overlay will follow pointer even the pointer move on toolbar, it is wired.

https://github.com/toeverything/blocksuite/assets/99816898/7a2e5c9a-4b0d-42f2-b9a0-0ebb07fd4b7a

After:
Hide overlay when pointer move on toolbar.

https://github.com/toeverything/blocksuite/assets/99816898/16e6736f-8fbe-44d5-a3b2-ebc5e8fa0581


doodlewind (approved):
Nice improvement in details!

PR #4439, opened 2023-08-29
perf: replace spread with Array.from

sxzz:
Benchmark: https://jsben.ch/5lKjg

### DIFF
--- a/packages/store/src/persistence/blob/indexeddb-storage.ts
+++ b/packages/store/src/persistence/blob/indexeddb-storage.ts
@@ -1,23 +1,21 @@
-import type { createStore } from 'idb-keyval';
+import type { UseStore } from 'idb-keyval';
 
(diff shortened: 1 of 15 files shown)


packages/store/src/persistence/blob/indexeddb-storage.ts
```diff
@@ -1,23 +1,21 @@
...
-    mimeTypeDbPromise = createStore(`${database}_blob_mime`, 'blob_mime');
+    db = createStore(`${database}_blob`, 'blob');
+    mimeTypeDb = createStore(`${database}_blob_mime`, 'blob_mime');
     return idb;
   });
   return {
     crud: {
       get: async (key: string) => {
         const get = (await idbPromise).get;
-        const db = await dbPromise;
```
sxzz, line 19: No more Promise here

PR #4457, opened 2023-08-30
fix(page): remove children when selecting list and delete

lawvs:
Revert part of https://github.com/toeverything/blocksuite/pull/4428

Related to https://github.com/toeverything/blocksuite/pull/4434


packages/blocks/src/page-block/keyborad/keyboard-manager.ts
```diff
@@ -104,13 +100,8 @@ export class PageKeyboardManager {
     const parent = this._page.getParent(first);
     const index = parent?.children.indexOf(first);
 
-    const children = this._deleteBlocksBySelection(selections);
-    const blockId = this._page.addBlock(
-      flavour,
-      { props, children },
```
lawvs, line 110: I think that the error occurred because the props were not correctly spread.

Saul-Mirone (approved):

donteatfriedrice (approved):

PR #4484, opened 2023-08-31
fix(edgeless): make tooltip with shortcut conform to design

CatsJuice:
- before
CleanShot 2023-08-31 at 08 53 54@2x
- after
CleanShot 2023-08-31 at 08 53 21@2x


doodlewind:
Thanks for the fix! Just note that the `style` commit type is normally used for coding style. UI styling fixes should also be marked as `fix`.

doodlewind (approved):

CatsJuice:
> Thanks for the fix! Just note that the `style` commit type is normally used for coding style. UI styling fixes should also be marked as `fix`.

Got it!

Issue #3924, opened 2023-08-09
The viewport will jilter when drag from block hub and move the block to the left bottom


donteatfriedrice:
https://github.com/toeverything/blocksuite/assets/99816898/0bde1618-b179-401a-b12d-fd407d7509ec

doodlewind:
Though the position of drag indicator can be confusing at times, but it's not reproduced:

https://github.com/toeverything/blocksuite/assets/7312949/37e3d0c4-2414-47af-8d73-c9ad3c7aa2fc

donteatfriedrice:
This issue can now be reproduced:
I guess it would be reproduced if drag the block to the left-bottom of the page, out of center area.
And issue #3684 maybe related with this problem. 

https://github.com/toeverything/blocksuite/assets/99816898/8f741d90-bd4d-4c51-8ad8-0cc5b11be846

doodlewind:
Closed by #4509

PR #4509, opened 2023-08-31
fix(page): viewport jilter when drag to left-bottom and cannot insert to page bottom from block-hub

### LINKED ISSUE #3684: Cannot drag and drop blank lines from blockhub to the bottom of the editor to insert blank lines
This feature was still working in the version a few days ago, and now the latest version is broken.

donteatfriedrice:
To close #3684 #3924 

This issue caused by legacy doc page check api.
Now the page tagName change to 'AFFINE-DOC-PAGE', but the code still check 'AFFINE-DEFAULT-PAGE', this would let  function findBlockElement() return pageBlockElement too, and cause calculate drop target position out of page and jilter, this is incorrect, the findBlockElement() function should only return blockElement.

Now we can insert blank lines without jilter:

https://github.com/toeverything/blocksuite/assets/99816898/a0042a7c-87de-48f0-a8f6-f853d9d8e675

### DIFF
--- a/packages/blocks/src/__internal__/utils/query.ts
+++ b/packages/blocks/src/__internal__/utils/query.ts
@@ -409,10 +409,10 @@ function hasBlockId(element: Element): element is BlockComponentElement {
 }
 
 /**
- * Returns `true` if element is default page.
+ * Returns `true` if element is doc page.
  */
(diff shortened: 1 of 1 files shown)


doodlewind (approved):

doodlewind:
Good catch!

PR #4514, opened 2023-09-01
fix: import notion embed block

### LINKED ISSUE #4180: Support importing file embed from notion
Because the blocksuite has file block type.  So when importing notion, the file embed should transform to a file block instead of a link.

zuoxiaodong0815:
close #4180


fourdim (commented):
image
Before state-transitions.png
After state-transitions.png.png

packages/blocks/src/__internal__/utils/types.ts
```diff
@@ -247,6 +247,8 @@ export type SerializedBlock = {
   children: SerializedBlock[];
   sourceId?: string;
   caption?: string;
+  name?: string;
```
doodlewind, line 250: Are we going to deprecate using this type? /cc @Saul-Mirone @fourdim

doodlewind (approved):

packages/blocks/src/__internal__/utils/types.ts
zuoxiaodong0815: Ok. Because the attachment block has the name property.

zuoxiaodong0815:
> image Before state-transitions.png After state-transitions.png.png

the fix pr is here. https://github.com/toeverything/blocksuite/pull/4587

packages/blocks/src/__internal__/utils/types.ts
Saul-Mirone: > Are we going to deprecate using this type? /cc @Saul-Mirone @fourdim

Yep

PR #4518, opened 2023-09-01
refactor(edgeless): remove phasor package

regischen:
remove phasor package


tests/utils/actions/edgeless.ts
```diff
@@ -4,13 +4,13 @@ import '../declare-test-window.js';
...
 
-import type {
-  CssVariableName,
-  IPoint,
-  NoteBlockModel,
+import {
+  type CssVariableName,
+  type IPoint,
+  type NoteBlockModel,
+  Vec,
```
fourdim, line 11: Remove this from tests.

pnpm-lock.yaml
```diff
@@ -1514,7 +1299,7 @@ packages:
       '@types/node': 18.17.11
       playwright-core: 1.37.1
     optionalDependencies:
-      fsevents: 2.3.2
+      fsevents: registry.npmmirror.com/fsevents@2.3.2
```
fourdim, line 1302: Remove the npmmirror

pnpm-lock.yaml
```diff
@@ -28,100 +28,100 @@ importers:
     devDependencies:
       '@commitlint/cli':
         specifier: ^17.7.1
-        version: 17.7.1
+        version: registry.npmmirror.com/@commitlint/cli@17.7.1
```
fourdim, line 31: Remove the npmmirror.

fourdim (changes requested):

pnpm-lock.yaml
fourdim: https://pnpm.io/npmrc#https-proxy
https://pnpm.io/npmrc#http-proxy

regischen: why should we modify pnpm-lock.yaml manuallly? it's generated

fourdim: You should not modify it manually. What you need to do is to use the proxy to pull from npmjs rather than the mirror.

fourdim: For oversea users, the npmmirror field in the pnpm lock file will force the pnpm to request the package from the npmmirror. That will slow down their installation process (including CI), and may violate their goverment or company policies.

pnpm-lock.yaml
```diff
@@ -162,6 +159,9 @@ importers:
       file-type:
         specifier: ^16.5.4
         version: 16.5.4
+      fractional-indexing:
+        specifier: ^3.2.0
+        version: registry.npmmirror.com/fractional-indexing@3.2.0
```
fourdim, line 164: still

pnpm-lock.yaml
regischen: okay, it's because in my machine I use npm mirror. 
It would be fine that I use the official npm source.

fourdim (approved):

Issue #4521, opened 2023-09-01
Lost content with duplicate note action in edgeless


BangKk:
What happened?

https://github.com/toeverything/blocksuite/assets/13979818/606e6be8-8abe-4ce0-b5e5-67b8eb61cc75

Distribution version
last version of playground

What browsers are you seeing the problem on if you're using web version?
Chrome 116.0.5845.110

Relevant log output
No response

Anything else?
No response

Are you willing to submit a PR?
- [x] Yes I'd like to help by submitting a PR!

doodlewind:
Thanks the report! This should be a regression.

PR #4529, opened 2023-09-01
fix(edgeless): lost content with duplicate note action in edgeless #4521

BangKk:
fix: #4521.

The `getService` import statement was deleted in [pr-4316 service.ts](https://github.com/toeverything/blocksuite/pull/4316/files#diff-ea130923a45d52986ac100629b83b0e49c6661ff6bf6e9718807b24655f08c3d) and [pr-4316 note-service.ts](https://github.com/toeverything/blocksuite/pull/4316/files#diff-4418768de534fcccdd14733aba3a21e49a6d879c8d0d9af87b45e0779211560c).

It was used but deleted. is this a mistake? And i add a unit test for this case.

### DIFF
--- a/packages/blocks/src/__internal__/service/legacy-services/note-service.ts
+++ b/packages/blocks/src/__internal__/service/legacy-services/note-service.ts
@@ -4,6 +4,7 @@ import type { NoteBlockModel } from '../../../note-block/note-model.js';
 import type { SerializedBlock } from '../../utils/index.js';
 import { addSerializedBlocks } from '../json2block.js';
 import { BaseService } from '../service.js';
+import { getService } from '../singleton.js';
 
 export class NoteBlockService extends BaseService<NoteBlockModel> {
   override async json2Block(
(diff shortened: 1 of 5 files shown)


BangKk:
I will see why the test failed.

BangKk:
It failed at https://github.com/toeverything/blocksuite/pull/4466/files#diff-41c1fdbb8d43aa2c5f7eeae3dd8901431c75b425e530765ca9e0c79c1d3f2ca2

packages/blocks/src/__internal__/service/legacy-services/note-service.ts
```diff
@@ -4,6 +4,7 @@ import type { NoteBlockModel } from '../../../note-block/note-model.js';
 import type { SerializedBlock } from '../../utils/index.js';
 import { addSerializedBlocks } from '../json2block.js';
 import { BaseService } from '../service.js';
+import { getService } from '../singleton.js';
```
doodlewind, line 7: @zuoxiaodong0815 the workaround here is sourced from your `// @ts-ignore`, the parameter doesn't match. This should be fixed.

PR #4540, opened 2023-09-01
feat: improve the type of command system

zzj3720:
Ensure type safety for context

PR #4553, opened 2023-09-02
feat(database): support using arrow keys to select kanban cells

zqran:
part of: #4494 

https://github.com/toeverything/blocksuite/assets/15389209/c57645af-520a-45bd-80f7-33ab634e6e64


Saul-Mirone:
Just a hint: we have [command](https://github.com/toeverything/blocksuite/pull/4540) now so you can reuse some duplicate logic with it.

packages/blocks/src/database-block/kanban/selection.ts
```diff
@@ -120,33 +121,260 @@ export class KanbanSelection {
...
+    );
+    const cardPos = nextCards
+      .map((card, index) => {
+        const targetRect = card.getBoundingClientRect();
+        return {
+          offsetY: getYOffset(rect, targetRect),
+          index,
+        };
+      })
+      .reduce((prev, curr) => {
```
zzj3720, line 204: Does this cause an error if there is only one cell?

packages/blocks/src/database-block/kanban/selection.ts
```diff
@@ -120,33 +121,260 @@ export class KanbanSelection {
...
+        if (prev.offsetY < curr.offsetY) {
+          return prev;
+        }
+        return curr;
+      });
+    const nextCell = cells[cellPos.index];
+
+    return {
+      cell: nextCell,
+      cardId: nextCard.getAttribute('data-card-id') as string,
```
zzj3720, line 233: Maybe nextCard.cardId would be more typesafe.

zzj3720 (approved):

Saul-Mirone:
We should add some tests for this feature.

doodlewind (approved):
Should be good to have tests in following PRs.

PR #4556, opened 2023-09-02
fix(edgeless): canvas text editor rotation

### LINKED ISSUE #3457: Text editor missing rotation when editing rotated shape text
https://github.com/toeverything/blocksuite/assets/50035259/7ec4a024-8a3e-4241-8705-3b200144a576

LittleSound:
resolve #3457

#### Demo:

[feat-text-editor-rotation.webm](https://github.com/toeverything/blocksuite/assets/19204772/1c5082cf-1eab-4a77-9ae9-168ebc324dfc)


packages/global/src/utils/math.ts
```diff
@@ -26,3 +26,45 @@ export const clamp = (value: number, min: number, max: number): number => {
...
   return value;
 };
+
+/**
+ * Converts an angle from degrees to radians.
+ *
+ * @param {number} angle - The angle in degrees.
+ * @returns {number} - The angle in radians.
+ */
+export function toRadians(angle: number) {
```
doodlewind, line 36: This global math module only exists for historical reasons. Adding the utils within the edgeless related dirs should be the better scope.

doodlewind (approved):
LGTM, thanks for the fix!

regischen:
there already has toRadian and rotwith
https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/surface-block/utils/vec.ts#L281C13-L281C13

LittleSound:
> there already has toRadian and rotwith https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/surface-block/utils/vec.ts#L281C13-L281C13

@regischen Thank you for telling me about this, I have reorganized these codes, removed the duplicate functions, the new code is in PR #4563

Issue #4221, opened 2023-08-18
Support hover animation for drag handle


doodlewind:

hwangdev97:
We should be done with this already? @donteatfriedrice cc @doodlewind

donteatfriedrice:
> We should be done with this already? @donteatfriedrice cc @doodlewind

I think this feature has been fulfilled if you agree and find it acceptable.

doodlewind:
This is considered as landed, thanks @donteatfriedrice @hwangdev97 @CatsJuice for the help involved!

PR #4566, opened 2023-09-03
fix(page): drag handle's animation is strange when multi-selected

### LINKED ISSUE #4565: Optimize the animation of expanding and retracting the drag handle
The animation of expanding & retracting the drag handle is kind of strange, the expand/retract center should be fixed to the hovered block.

https://github.com/toeverything/blocksuite/assets/39363750/58f9c875-eeca-4d17-9e18-8e3b482e23da

CatsJuice:
Closes #4565


donteatfriedrice:
It seems that when moving between blocks, the drag handle sometimes stays at the block before, maybe need to check with it...


https://github.com/toeverything/blocksuite/assets/99816898/ba4b7c7a-6bde-4867-8553-f5d8ad5852a7

donteatfriedrice:
Thanks for the quick update!.  

It looks kind of strange when the pointer move on multi-blocks selected after the second pr. Looks as below, the movement of drag handle also applied with the animation, and it will expanding firstly then shorten to origin height.

https://github.com/toeverything/blocksuite/assets/99816898/322770a7-11a2-4c27-9b0b-84717e432a79

Maybe can remove the second parameter of the function _show() based on the first pr when pointerleave. 

Screenshot 2023-09-03 at 21 58 03

The effect look like this:

https://github.com/toeverything/blocksuite/assets/99816898/1dd9daa9-0f2e-45f4-8a14-69eaffa76096

CatsJuice:
> Thanks for the quick update!.
> 
> It looks kind of strange when the pointer move on multi-blocks selected after the second pr. Looks as below, the movement of drag handle also applied with the animation, and it will expanding firstly then shorten to origin height.
> 
>  Screen.Recording.2023-09-03.at.21.47.46.mov 
> Maybe can remove the second parameter of the function _show() based on the first pr when pointerleave.
> 
> Screenshot 2023-09-03 at 21 58 03
> The effect look like this:
> 
>  Screen.Recording.2023-09-03.at.21.56.44.mov

Yes indeed, I'm still testing more edge cases that also include this one. Thank you very much for your help!

donteatfriedrice (approved):
Cool optimization!

hwangdev97 (approved):

donteatfriedrice:
related to #4221

PR #4583, opened 2023-09-04
fix: image filename in card view

adityash1:
Changes
-  shows image filename in card view when dragged and dropped & when uploaded via slash menu option

Fixes https://github.com/toeverything/blocksuite/issues/4603

### DIFF
--- a/packages/blocks/src/__internal__/utils/filesys.ts
+++ b/packages/blocks/src/__internal__/utils/filesys.ts
@@ -221,6 +221,10 @@ export const uploadImageFromLocal = async (storage: BlobManager) => {
     const sourceId = await storage.set(file);
     res.push({ file, sourceId });
   }
+  const { saveAttachmentData } = withTempBlobData();
+  res.forEach(({ file, sourceId }) => {
+    saveAttachmentData(sourceId, { name: file.name });
+  });
   return res;
 };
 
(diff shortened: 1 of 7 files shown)


lawvs (changes requested):
Thanks for your enthusiasm! However, The image block does not include the name property by design.

placing all miscellaneous data into a block scheme is not a good idea for several reasons:

- It will make type-checking unreliable. The growing amount of empty data will affect the development experience.
- They are mostly useless in most scenarios.
- It can also cause unexpected privacy leaks.

So we only save temporary data in memory. (If you refresh the page, the data carried over from the previous block will be lost.)

adityash1:
> Thanks for your enthusiasm! However, The image block does not include the name property by design.
> 
> placing all miscellaneous data into a block scheme is not a good idea for several reasons:
> 
> * It will make type-checking unreliable. The growing amount of empty data will affect the development experience.
> * They are mostly useless in most scenarios.
> * It can also cause unexpected privacy leaks.
> 
> So we only save temporary data in memory. (If you refresh the page, the data carried over from the previous block will be lost.)

Makes sense.  It was using blob to get the filename. And as I know, blobs doesn't contain file names. What should be my other approach here ? Can you explain ?

lawvs:
> Makes sense. It was using blob to get the filename. And as I know, blobs doesn't contain file names. What should be my other approach here ? Can you explain ?

Yes, It's a bug. I'll add commits to your pull request that use a default name when no name is available in `attachmentConvertData`.

lawvs (approved):

PR #4584, opened 2023-09-04
ci: size report

sxzz:


tsconfig.json
```diff
@@ -57,10 +57,18 @@
...
       "path": "./tests"
+    },
+    {
+      "path": "./scripts"
     }
   ],
   "ts-node": {
     "esm": true,
-    "experimentalSpecifierResolution": "node"
+    "experimentalSpecifierResolution": "node",
```
doodlewind, line 67: I thought it could be better to use the same rule for node scripts.

sxzz: I don't think so. For `ts-node` I think unused locals/params checking is unnecessary. In most cases using `ts-node` is for developing or debugger. However, for compile-time, it's somehow a linter (like ESLint)

scripts/size-data.ts
```diff
@@ -0,0 +1,118 @@
...
+    entry: 'dist/index.js',
+  },
+  {
+    name: 'editor',
+    imports: '*',
+    pkg: path.resolve(pkgDir, 'editor'),
+    entry: 'dist/index.js',
+  },
+  {
+    name: 'editor',
```
doodlewind, line 35: Is this duplicated with the list item before?

doodlewind (commented):
It's expected that for packages, we measure the `store` / `blocks` / `editor` / `virgo` packages.

scripts/size-generate.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+rm -fr ./temp/size &&
+  pnpm run build:editor &&
+  pnpm run size:data &&
+  pnpm -C ./packages/playground run build:size &&
+  pnpm run --silent size:report >./size-report.md &&
+  cat ./size-report.md
```
fourdim, line 8: Use `set -eu` instead.

sxzz: Where to add this and why?

lawvs, line 8:
```suggestion
set -eu

rm -fr ./temp/size
pnpm run build:editor 
pnpm run size:data 
pnpm -C ./packages/playground run build:size 
pnpm run --silent size:report >./size-report.md 
cat ./size-report.md
```

The `-e` option tells the shell to exit immediately if any command within the script exits with a non-zero status.

`-u` treats unset variables and parameters as an error when performing parameter expansion.

I also believe that this can help us write more reliable shell.

https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

doodlewind (approved):
LGTM, thanks for the feat!

sxzz:
Seems like E2E failed randomly. 🤔

PR #4589, opened 2023-09-05
feat: support LEDITS image editing

### LINKED ISSUE #4399: Support image editing addon
This is related to integration of [DragGAN](https://github.com/XingangPan/DragGAN)-based image editing.

doouding:
close #4399

### Notable change
- Add the `AffineImageToolbarWidget` widget (singleton, migrate from legacy `ImageOptionsTemplate`)
    - The custom tool can be registered through `widgetElements.imageToolbar.registerEntry` at runtime
- Add the `AffineCustomModal` widget (singleton)
    - Open the modal by calling `widgetElements.modal.open`, the content of the modal can be customized
- Add an LEDITS editor (feature flag: `enable_bultin_ledits`)


doodlewind (commented):
With the discussion with @Saul-Mirone, for now the extension and addon manager part may be better postponed, making this PR work without special loading mechanism. Instead, making this a standard feature inside BlockSuite (with feature flag disabled by default) should be a better fit for the moment, and let's just leave @Saul-Mirone to handle the further extension prototyping.

packages/blocks/src/widgets/image-toolbar/addon-manager.ts
```diff
@@ -0,0 +1,28 @@
...
+  addons: ImageAddonRegistion[] = [];
+
+  register(registion: ImageAddonRegistion) {
+    if (this.addons.find(addon => addon.name === registion.name)) {
+      return;
+    }
+
+    this.addons.push(registion);
+  }
+}
```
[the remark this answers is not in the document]
doouding: I think it's just a naming convention problem. The main purpose of this "AddonManager" is to manage custom-registered menu entries.  Maybe `AddonManager` is not a good name, but this kind of manager is necessary.

lawvs:
I also believe there is no need to introduce the new `extensions`. We can customize the toolbar by overriding the `affine-image-toolbar-widget`, which is the role of the Widget. By using this approach, we don't need to modify any code in other places.

You can find references here: https://github.com/toeverything/blocksuite/pull/3407 https://github.com/toeverything/AFFiNE/pull/3488

/cc @Flrande @Saul-Mirone 

However, there are still some drawbacks to override widgets (difficult to customize configurations) and a lack of best practices.

doouding:
Just removed the extension-related code.

packages/blocks/src/widgets/modal-widget/modal.ts
```diff
@@ -0,0 +1,147 @@
...
+  callback?: (div: HTMLDivElement) => void;
+
+  close() {
+    this.remove();
+  }
+
+  override firstUpdated(): void {
+    this.callback?.(
+      this.renderRoot.querySelector('.modal-main') as HTMLDivElement
+    );
```
lawvs, line 101: Try to use `ref` property instead callback.

See https://lit.dev/docs/templates/directives/#ref

packages/blocks/src/image-block/ledits/utils.ts
```diff
@@ -0,0 +1,132 @@
+export function createInlineIframe() {
+  const htmlContent = `<!DOCTYPE html>
```
lawvs, line 2:
```ts
const htmlContent = html`<html></html>`
const rawString = String.raw(html.strings, ...html.values);
```

Use the `html` tagged template literal from Lit to highlight text in the editor. To get the original string, use `String.raw(html.strings, ...html.values)`.

lawvs (approved):

PR #4596, opened 2023-09-05
Requesting HTTP Requests instead of WebSockettings

spacewalkingninja:
Hi!
We'd like to see this with a bit of more traditional http request flavor.
Is it possible?
:)

### DIFF
--- a/packages/playground/apps/default/components/quick-edgeless-menu.ts
+++ b/packages/playground/apps/default/components/quick-edgeless-menu.ts
@@ -34,10 +34,13 @@ import { css, html, nothing } from 'lit';
 import { customElement, property, query, state } from 'lit/decorators.js';
 import { Pane } from 'tweakpane';
 
-import {
+/*import {
   generateRoomId,
   initCollaborationSocket,
-} from '../providers/websocket-channel';
+} from '../providers/http-requests';
+*/
+import {   generateRoomId,
+  initCollaborationRequest } from '../providers/http-requests';
 import { notify } from '../utils/notify';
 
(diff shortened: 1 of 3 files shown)


spacewalkingninja:
Still haven't implemented the PHP server//side but from what I understand I would ((prety//much)) just have to store/send the json (EXAMPLE):
```{
  "blocks": {
    "gH97y19qys": {
      "sys:id": "gH97y19qys",
      "sys:flavour": "affine:page",
      "sys:children": [
        "ZmPUJxcAJR",
        "aQhKWxJlEH"
      ],
      "prop:title": []
    },
    "ZmPUJxcAJR": {
      "sys:id": "ZmPUJxcAJR",
      "sys:flavour": "affine:surface",
      "sys:children": [],
      "prop:elements": {
        "type": "$blocksuite:internal:native$",
        "value": {}
      }
    },
    "XUHWe3BSQq": {
      "sys:id": "XUHWe3BSQq",
      "sys:flavour": "affine:paragraph",
      "sys:children": [],
      "prop:type": "h6",
      "prop:text": [
        {
          "insert": "TESTINGGGGGGGGGGGGGGGg"
        }
      ]
    },
    "hhgXOfCI7M": {
      "sys:id": "hhgXOfCI7M",
      "sys:flavour": "affine:paragraph",
      "sys:children": [],```

doouding:
Hello, thank you for your interest in this project. It looks like you want to implement an HTTP based provider for just an experimental purpose. It would be better if you could just clone this project and test it locally (or just fork this project).
BTW, HTTP is theoretically possible but it might be a little cumbersome.

PR #4613, opened 2023-09-06
refactor: use currentcolor to set svg fill

### LINKED ISSUE #4391: Better way to implement dynamic `fill` property in svg icons
### TLDR
use [`currentColor`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword) instead!

---

### Using `currentColor` in SVG is more convenient and maintainable.

Instead of directly querying multiple SVG elements and setting inline styles for them, set `currentColor` for `fill` property, and then set `css color` directly to `SVG` element(or ancestor element), thanks to css inheritance, it should work as expected.

---

### Example

original:
```ts
@query('.brush-head-rect')
private _brushHeadRect!: SVGElement;

private _setBrushColor(color: string) {
    this._brushHeadRect.style.fill = `var(${color})`;
    this._brushMidlineRect.style.fill = `var(${color})`;
    this._brushMidlineStroke.style.stroke = `var(${color})`;
}
```
The above code can be completely removed, bind dynamic `color` style instead:

```ts
render() {
  const colorStyle = styleMap({ color: this.color })

  return html`
    <div .style=${colorStyle}>
       <!-- ..... -->
    </div>
  `
}
```

lzlme:
close #4391

### DIFF
--- a/packages/blocks/src/icons/edgeless.ts
+++ b/packages/blocks/src/icons/edgeless.ts
@@ -999,11 +999,17 @@ export const EdgelessPenIcon = html`<svg
       y="-1"
       width="11"
       height="11"
-      fill="none"
+      fill="currentColor"
       class="brush-head-rect"
     />
   </g>
-  <rect y="48" width="22" height="2" fill="none" class="brush-midline-rect" />
+  <rect
+    y="48"
+    width="22"
+    height="2"
+    fill="currentColor"
+    class="brush-midline-rect"
+  />
   <rect
     x="0.5"
     y="48.5"
@@ -1011,6 +1017,7 @@ export const EdgelessPenIcon = html`<svg
     height="1"
     stroke="none"
     stroke-opacity="0.1"
+    fill="currentColor"
     class="brush-midline-stroke"
   />
   <defs>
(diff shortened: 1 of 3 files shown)


CatsJuice (approved):

donteatfriedrice:
The color of brush and shape can be consistent with the chosen color after your work, it works well. But the arrow icon with shape seems to be invisible, maybe need to check with it. 

Screenshot 2023-09-06 at 13 35 35

I am not sure whether should keep the arrow icon for the right part of the edgeless toolbar, since the new design of the toolbar have removed all the arrow icon for it.  Maybe you need to check with the designer @hwangdev97

lzlme:
> The color of brush and shape can be consistent with the chosen color after your work, it works well. But the arrow icon with shape seems to be invisible, maybe need to check with it.
> 
> Screenshot 2023-09-06 at 13 35 35
> I am not sure whether should keep the arrow icon for the right part of the edgeless toolbar, since the new design of the toolbar have removed all the arrow icon for it. Maybe you need to check with the designer @hwangdev97

Yes, the new design of the toolbar has removed the arrow icon, so I deleted it.

donteatfriedrice (approved):

hwangdev97 (approved):

PR #4630, opened 2023-09-06
chore: delete database flag

zzj3720:


doodlewind (commented):
Are we still using the `__unstableSchema` for database? If so, guess it could be removed together.

Saul-Mirone (approved):

Issue #4538, opened 2023-09-01
Edgeless Toolbar optimization


hwangdev97:
![shot-2023-09-01-18 42 41@2x](https://github.com/toeverything/blocksuite/assets/24713927/0dc0f734-7d1c-4b16-bcba-1e1a99433709)

![shot-2023-09-01-19 55 20@2x](https://github.com/toeverything/blocksuite/assets/24713927/60444a46-8e75-4b4c-bfab-97fb3e95cac4)

[the design file link](https://www.figma.com/file/UTTtfVCcStw1dKUg0gCSPo/%F0%9F%9B%9E-Design-System?type=design&node-id=4454%3A31690&mode=design&t=2HhMqxlgFat698b8-1)

- [x] Modify the style of the pen and eraser. Keep the hover elongation effect.
- [x] Modify the style of the connector to be a separate icon.
- [x] Display icons and badges on the left side, but not on the right side.
- [ ] Enlarge the shape of the Shape, supporting dragging and animations.
- [x] The pen/shape/text supports color changes.
- [x] Secondary menus support scrolling, and the left and right buttons enlarge on hover.

doodlewind:
Thanks for the optimization! I would ask if @lzlme is capable of tweaking this in this iteration.

lzlme:
> Thanks for the optimization! I would ask if @lzlme is capable of tweaking this in this iteration.

I'll get it done !

PR #4637, opened 2023-09-06
feat: edgeless toolbar update

lzlme:
Part of #4538 

- [x] Modify the style of the connector to be a separate icon.
- [x] Display icons and badges on the left side, but not on the right side.
- [x] Enlarge slide menu button
- [x] Enlarge icon container padding


hwangdev97:
image

The details are still relatively poor, but considering that the UI will continue to be adjusted in the future, it is acceptable to do this for the time being

hwangdev97 (approved):

donteatfriedrice (approved):

PR #4645, opened 2023-09-07
feat(edgeless): merge select and hand

lzlme:
Part of #4538


packages/blocks/src/page-block/edgeless/edgeless-keyboard.ts
```diff
@@ -184,38 +178,46 @@ export class EdgelessPageKeyboardManager extends PageKeyboardManager {
...
       }
     });
+    this._bindToggleHand();
+  }
+
+  private _bindToggleHand() {
+    this.pageElement.handleEvent(
+      'keyDown',
+      ctx => {
+        const event = ctx.get('defaultState').event;
```
Flrande, line 188: You can use 'keyboardState' here and no need to check event type.

hwangdev97:
when switch hand and select, please add this animation:

shot-2023-09-07-16 11 25

and I think you need update the select icon for blocksuite 🤔

on designer's file
shot-2023-09-07-16 12 49@2x

on playground
shot-2023-09-07-16 13 01@2x

packages/blocks/src/page-block/edgeless/components/toolbar/default/default-tool-button.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  private _fadeIn() {
+    this.currentIcon.style.opacity = '1';
+    this.currentIcon.style.transform = `translateY(0px)`;
+  }
+
+  private _changeTool() {
+    this._fadeOut();
+    setTimeout(() => {
```
Flrande, line 52: Why we need `setTimeout` here?

hwangdev97 (approved):

Flrande (approved):

Issue #4676, opened 2023-09-08
Click in search input won't close popup in code-block's language select menu


CatsJuice:
https://github.com/toeverything/blocksuite/assets/39363750/5c78c5a4-adfe-410b-893f-58716a783443

PR #4677, opened 2023-09-08
feat: add `closeOnClickAway` option for `createSimplePortal`

### LINKED ISSUE #4675: Support auto close when click away for `createLitPortal`
For now, when it comes to create a new `portal`, we usually have to handle `click-outside` event in order to close `portal` when blurred. I think it will be more convenient to handle this in `createSimplePortal`.

### Expected solution

```ts
createLitPortal({
    // maybe some option like this
    closeOnClickAway: true
})
```

### Solution yet
- Use a global mask and capture client event on mask:
  ```ts
  // packages/blocks/src/attachment-block/components/rename-model.ts
  html` <div
    class="affine-attachment-rename-overlay-mask"
    @click="${() => abortController.abort()}"
  ></div>`
  ```
- Register `client` on `document` in `portal` template component

  ```ts
   // packages/blocks/src/code-block/components/lang-list.ts
   override async connectedCallback() {
      super.connectedCallback();
      // Avoid triggering click away listener on initial render
      setTimeout(() =>
        document.addEventListener('click', this._clickAwayListener)
      );
  
      setTimeout(() => {
        this.filterInput?.focus();
      }, 0);
    }
  
   override disconnectedCallback() {
      super.disconnectedCallback();
      document.removeEventListener('click', this._clickAwayListener);
    }
  
    private _clickAwayListener = (e: Event) => {
      if (this.renderRoot.parentElement?.contains(e.target as Node)) {
        return;
      }
      this.onClose?.();
    };
  ```

CatsJuice:
close #4675, #4676

Extend `PortalOptions`:
```ts
/**
 * Whether to close the portal when click away(click outside).
 * @default false
 */
closeOnClickAway?: boolean;
```

### DIFF
--- a/packages/blocks/src/components/portal.ts
+++ b/packages/blocks/src/components/portal.ts
@@ -88,7 +88,7 @@ type PortalOptions = {
   /**
    * The portal is removed when the AbortSignal is aborted.
    */
-  abortController?: AbortController;
+  signal?: AbortSignal;
   /**
    * Defaults to `true`.
    */
@@ -111,7 +111,7 @@ type PortalOptions = {
 export function createSimplePortal({
   template,
   container = document.body,
-  abortController = new AbortController(),
+  signal = new AbortController().signal,
   renderOptions,
   shadowDom = true,
   identifyWrapper = true,
@@ -123,7 +123,7 @@ export function createSimplePortal({
   if (shadowDom) {
     portalRoot.attachShadow({ mode: 'open' });
   }
-  abortController.signal.addEventListener('abort', () => {
+  signal.addEventListener('abort', () => {
     portalRoot.remove();
   });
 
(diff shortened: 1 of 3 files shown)


CatsJuice:
🤣 I'm not sure what happened, but yes, #4676  was fixed!

packages/blocks/src/components/portal.ts
```diff
@@ -150,6 +156,22 @@ export function createSimplePortal({
...
+
+  function _onClickAway(e: Event) {
+    if (portalRoot.contains(e.target as Node)) return;
+    dispose();
+  }
+
+  if (closeOnClickAway) {
+    // Avoid triggering click away listener on initial render
+    setTimeout(() => document.addEventListener('click', _onClickAway));
+  }
```
lawvs, line 168: Try using the following pattern; you no longer need to use `dispose`.

```ts
  function _onClickAway(e: Event) {
    if (portalRoot.contains(e.target as Node)) return;
    abortController.abort();
  }

document.addEventListener('click', _onClickAway, { signal: abortController.signal }));
```

By the way, I believe placing `closeOnClickAway` in `createLitPortal` may be more appropriate since I want to keep `createSimplePortal` as tidy as possible.

lawvs (commented):
Nice works!

lawvs (approved):

PR #4685, opened 2023-09-09
chore(page): set shape-rendering to auto for attachment-block svg icon

CatsJuice:
CleanShot 2023-09-10 at 00 04 09@2x


lawvs (approved):
This value is built-in when exporting from Figma. Is it possible to handle this case in Figma next time?

CatsJuice:
> This value is built-in when exporting from Figma. Is it possible to handle this case in Figma next time?

Yes, but I'm not sure if it's possible to configure the rendering of exported SVG in Figma.

PR #4689, opened 2023-09-10
feat: new whenHover API

lawvs:
## Motivation

It is very common to show a popover when hovering over something.

```ts
// Existing implementation

class WebComponent {
  private _onHover = () => {
    this._hovered = true;
  };

  private _onHoverLeave = () => {
    this._hovered = false;
    this._hideAfterDelay();
  };

  // Need to handle delay manually
  private _hideAfterDelay = () => {
    if (this._hideTimer) clearTimeout(this._hideTimer);

    this._hideTimer = setTimeout(() => {
      if (!this._hovered) {
        this.hide();
      }
      this._hideTimer = null;
    }, 150);
  };


    super.connectedCallback();
    this.pageElement.handleEvent('pointerMove', ctx => {
      this.show();
    });
  }

  render() {
    // Add event listeners
    return html`<div
      @mouseover=${this._onHover}
      @mouseout=${this._onHoverLeave}
    ></div>`;
  }
}
```

## New Design

```ts
class WebComponent {
  constructor() {
    // Some hover tips
    let hoverTip: HTMLElement | null = null;
    
    // 1. Call whenHover at first
    const { setReference, setFloating } = whenHover(
      (isHover) => {
        if (!isHover) {
          // Clean
          hoverTips?.remove();
          return;
        }
        // Show hover element
        hoverTip = document.createElement("div");
        document.body.append(hoverTip);

        // 2. Set ref
        setFloating(hoverTip);
      },
      { hoverDelay: 500 }
    );
    this._setReference = setReference;
  }

  render() {
    // 3. Set ref
    return html` <div ref=${this._setReference}></div> `;
  }
}
```

## API

```ts
type whenHover = (whenHoverChange: (isHover: boolean, event: Event) => void, { leaveDelay, }?: {
    // 300ms by default
    leaveDelay?: number | undefined;
}): {
    setReference: (el: Element | undefined) => void;
    setFloating: (el: Element | undefined) => void;
    dispose: () => void;
}
```

## Furthermore

We can implement the `safe polygon` in the future.

see https://floating-ui.com/docs/usehover#safepolygon

### DIFF
--- a/packages/blocks/src/attachment-block/attachment-block.ts
+++ b/packages/blocks/src/attachment-block/attachment-block.ts
@@ -1,7 +1,9 @@
+import { whenHover } from '@blocksuite/global/utils';
 import { BlockElement } from '@blocksuite/lit';
 import { flip, offset } from '@floating-ui/dom';
 import { html, type PropertyValues } from 'lit';
 import { customElement, query, state } from 'lit/decorators.js';
+import { ref, type RefOrCallback } from 'lit/directives/ref.js';
 
 import { stopPropagation } from '../__internal__/utils/event.js';
 import { humanFileSize } from '../__internal__/utils/math.js';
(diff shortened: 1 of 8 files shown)


Saul-Mirone:
Maybe use a shared [reactive controller](https://lit.dev/docs/composition/controllers/) to share logic?

Saul-Mirone (approved):

Issue #3663, opened 2023-07-22
Add getting started tutorial for react and next.js


neelhabib:
I have tried but unable to implement the editor in Nextjs. Is there any simple way to implement in nextjs? Thanks

himself65:
See our minimum example https://github.com/himself65/affine-minimum/blob/stable/src/components/Editor.tsx

neelhabib:
> See our minimum example https://github.com/himself65/affine-minimum/blob/stable/src/components/Editor.tsx

Thanks. But its not straight forward. Requires other packages. I want to implement it on my existing project with redux. Is there any official docs for nextjs implementation. Thanks

himself65:
I'm working on this.

yunnian:
Is there any progress in the tutorial? I am also unable to use it in next.js 

error & stack：
![image](https://github.com/toeverything/blocksuite/assets/5779513/033acfa5-f6ff-49ea-87ba-dfc28dc18033) ![image](https://github.com/toeverything/blocksuite/assets/5779513/a7d5d14c-dbff-4d42-85e9-7c4a371af6e4)![image](https://github.com/toeverything/blocksuite/assets/5779513/937f9a8d-69bf-4177-80b2-7048faad5cf7)

doodlewind:
@zuoxiaodong0815 seems that the error stack above is related to `ContentParser`, which would require you to take a look in Next.js environment if there are recent regressions. Thanks.

zuoxiaodong0815:
> @zuoxiaodong0815 seems that the error stack above is related to `ContentParser`, which would require you to take a look in Next.js environment if there are recent regressions. Thanks.
Ok, I will look at it today later.

yunnian:
![image](https://github.com/toeverything/blocksuite/assets/5779513/abd340d6-f28c-4046-9311-cb241c9b881c)
官方的这个get start 试了市面上的框架都没执行成功过...，不用框架也报错 ，心累

himself65:
> ![image](https://user-images.githubusercontent.com/5779513/265233061-abd340d6-f28c-4046-9311-cb241c9b881c.png) 官方的这个get start 试了市面上的框架都没执行成功过...，不用框架也报错 ，心累

See our official binding!
https://github.com/toeverything/AFFiNE/blob/master/packages/component/src/components/block-suite-editor/index.tsx

kaianuar:
[...]

PR #4709, opened 2023-09-11
fix: fix errors in next.js

yunnian:
fix  two errors in next.js:
1. Part of [#3663](https://github.com/toeverything/blocksuite/issues/3663)   document and window is not defined
2. Unable to build in next.js if throw error , because next.js is prerendered on the server and hydrated on the client.


packages/blocks/src/index.ts
```diff
@@ -60,7 +60,7 @@ if (env[importIdentifier] === true) {
 }
 
 if (typeof window === 'undefined') {
-  throw new Error(
+  console.error(
```
himself65: I think it is indeed an error, we shouldn't just ignore it.

packages/editor/src/index.ts
```diff
@@ -22,7 +22,7 @@ if (env[importIdentifier] === true) {
 }
 
 if (typeof window === 'undefined') {
-  throw new Error(
+  console.error(
```
himself65: NO, editor cannot be imported in SSR mode Whatever how

himself65 (changes requested):

packages/blocks/src/index.ts
yunnian: In the latest next.js, even with the use of client mode, it still have some server-side rendering's works, so throwing exceptions here will cause build(npm run build) errors. `npm run dev` can work, but there are also error reminders

Do you have a good way to solve the build errors  in next.js ? We can discuss it .

Blocksuite cannot be rendered on the server, I think this reminder in console is enough, We also can add a notice in the doc.  Or do you have any other good ideas？

image

himself65: Yeah, the error here is correct. Your code here is still not correct.

'use client' also means it has SSR.

Don't hide the error. Next.js intends to run the code in a node.js environment

yunnian: @himself65 
I don't understand Why my code is incorrect. 

In the next.js app router mode, the `use client` is used to represent the client : https://nextjs.org/docs/app/building-your-application/rendering/client-components

How to write the correct code   in next.js's app router mode ？

Screenshot 2023-09-19 at 02 22 59

himself65: https://nextjs.org/docs/app/building-your-application/optimizing/lazy-loading#skipping-ssr

doodlewind (approved):
Thanks for the help! Also /cc @sxzz to see if the code snippet provided here could be reused in the ecosystem-ci.

PR #4710, opened 2023-09-11
refactor: move image toolbar widget to image block

lawvs:
Related to https://github.com/toeverything/blocksuite/pull/4589 https://github.com/toeverything/blocksuite/pull/4635

- Fix the issue where the image toolbar remains even after deleting the image block.
- Move image toolbar widget to the image block
- Using the new whenHoverController


Saul-Mirone (approved):

doouding:
Those disabled buttons in read-only mode should hide according to https://github.com/toeverything/AFFiNE/issues/4222

PR #4713, opened 2023-09-12
fix: modal-widget style

doouding:
- Fix modal style
- Ledits demo bug

### DIFF
--- a/packages/blocks/src/image-block/ledits/utils.ts
+++ b/packages/blocks/src/image-block/ledits/utils.ts
@@ -1,3 +1,4 @@
+/* eslint-disable */
 import { html } from 'lit';
 
 export function createInlineIframe() {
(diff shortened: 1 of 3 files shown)


packages/blocks/src/image-block/ledits/utils.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
lawvs, line 1:
```suggestion
```

lawvs:
It seems like it doesn't automatically add the image to the input, is it expected?


Screenshot 2023-09-12 at 17 01 58

packages/blocks/src/image-block/ledits/utils.ts
Saul-Mirone: You should only disable some of the rules instead of disable whole eslint.

Saul-Mirone (changes requested):
Please don't disable the whole eslint. Just disable some of the rules would be enough.

doouding:
> It seems like it doesn't automatically add the image to the input, is it expected?
> 
> Screenshot 2023-09-12 at 17 01 58

The regular expression code was formatted incorrectly. I've fixed it.

Saul-Mirone (approved):

PR #4720, opened 2023-09-12
feat: edgeless toolbar secondary menu scrollable

lzlme:
Part of #4538 
Secondary menus support scrolling, and the left and right buttons enlarge on hover.


Saul-Mirone (approved):

hwangdev97:
The default scrolling direction should be horizontal.

shot-2023-09-12-17 55 22

PR #4734, opened 2023-09-13
refactor(database): replace pure virgo with `rich-text` in text cell

Flrande:


packages/blocks/src/database-block/common/columns/rich-text/cell-renderer.ts
```diff
@@ -202,76 +184,41 @@ export class RichTextCellEditing extends BaseCellRenderer<Y.Text> {
...
   override connectedCallback() {
     super.connectedCallback();
-    if (this._container) {
-      this.init();
+    if (!this.value || typeof this.value === 'string') {
+      this._initYText(this.value);
     }
   }
 
+  override updated() {
```
zzj3720, line 204: add event listener in updated? This may result in listening for keydown event more than once

zzj3720 (approved):

PR #4735, opened 2023-09-13
feat(edgeless): toolbar ui improvement

lzlme:


hwangdev97:
Please remember select/hand state.

hwangdev97 (approved):

hwangdev97:
Good

packages/blocks/src/page-block/edgeless/components/toolbar/connector/connector-tool-button.ts
```diff
@@ -91,7 +93,7 @@ export class EdgelessConnectorToolButton extends WithDisposable(LitElement) {
 
   override render() {
     const type = this.edgelessTool?.type;
-
+    const arrowColor = type === 'connector' ? 'currentColor' : '#77757D';
```
donteatfriedrice, line 96: Does the design team require to set it as #77757D in both dark/light mode? Maybe it is better to use var(--affine-icon-color) here instead of #77757D?

donteatfriedrice (approved):

packages/blocks/src/page-block/edgeless/components/toolbar/connector/connector-tool-button.ts
lzlme: Yes, it's by design.

PR #4738, opened 2023-09-13
fix(edgeless): isolate the impact of widgets on the layout

### LINKED ISSUE #4701: Edgeless and page switching buttons are missing
In the case of #4699, the kanban that exceeds the screen length causes the toolbar to move upward as a whole, causing the switch button to be lost.

https://github.com/toeverything/blocksuite/assets/142964904/332b3a72-821b-431a-b31a-078b12dd6398

doouding:
close #4701


doouding:
I found that it is the widgets that mainly cause the broken layout of the edgeless mode. I isolate the impact of the widgets on the layout, but not sure if this can solve the whole potential problem.

regischen (approved):

Issue #4722, opened 2023-09-12
Support readonly mode in database block


houjoe0829:
related with  https://github.com/toeverything/AFFiNE/issues/4222
For details: https://affine-pro.slack.com/docs/T03K3CD2015/F05S5P1FXGQ

houjoe0829:
1. Currently, modifying a Database View in read-only mode is not supported. This will be addressed in the future.
2. Some newly created entrances to content have not been hidden, even though they are unresponsive.
3. Currently, it is not possible to expand content in detail view. It is recommended to support this feature.
![image](https://github.com/toeverything/blocksuite/assets/22443345/33e33c18-8a00-43c4-9ef2-47c4cdd54aa1)

zqran:
> 2. Some newly created entrances to content have not been hidden, even though they are unresponsive.

In read-only mode, filtering operations cannot be displayed? @houjoe0829

houjoe0829:
> > 2. Some newly created entrances to content have not been hidden, even though they are unresponsive.
> 
> In read-only mode, filtering operations cannot be displayed? @houjoe0829

Based on our previous discussion, for now we can temporarily not support the option to modify views in read-only mode, including filtering, sorting, and grouping. We can consider supporting it later.

PR #4745, opened 2023-09-13
feat(database): update content displayed in readonly mode

zqran:
close: #4722


packages/blocks/src/database-block/table/components/column-header/column-header.ts
```diff
@@ -25,6 +25,8 @@ export class DatabaseColumnHeader extends WithDisposable(ShadowlessElement) {
   }
   private addColumnPositionRef = createRef();
   addColumnButton = renderTemplate(() => {
+    if (this.readonly) return null;
```
zzj3720, line 28: why not return "nothing"

zqran: done

zzj3720 (approved):

doodlewind (approved):

PR #4758, opened 2023-09-14
feat: edgeless text change color

lzlme:
Part of #4538


hwangdev97:
If the content is placed, there should be no scrolling
shot-2023-09-14-22 05 05@2x
the design file:
shot-2023-09-14-22 06 16@2x

donteatfriedrice:
Should we record the text color last selected like other tools, such as brush, shape. So that we can keep adding text with same color?


https://github.com/toeverything/blocksuite/assets/99816898/cf305fb8-c650-4611-a3f8-21c50c0e516e

hwangdev97:
> Should we record the text color last selected like other tools, such as brush, shape, So that we can keep adding text with same color?
> 
>  Screen.Recording.2023-09-15.at.09.55.10.mov

yes cc @lzlme

lzlme:
`color menu no scroll` and `color save` were done. @hwangdev97 @donteatfriedrice

hwangdev97 (approved):

donteatfriedrice (approved):

PR #4796, opened 2023-09-17
docs: keep consistent

himself65:
keep the same with https://github.com/toeverything/AFFiNE/blob/master/LICENSE-MIT


doodlewind (changes requested):
No, we don't have agreement on migrating license.

PR #4808, opened 2023-09-18
fix(edgeless): note does not update when other peers modified note

### LINKED ISSUE #4697: Note doesn't follow selection on collaboration in edgeless

doouding:
close #4697

The `BlockSuiteRoot` component has built a rerender mechanism inside. But, it does not work for edgeless mode cause edgeless wraps everything inside `SurfaceBlock` which stops the notes that are deep inside the components tree from updating.

I added an extra slot listener to track notes' updates individually.

### DIFF
--- a/packages/blocks/src/page-block/edgeless/components/auto-complete/edgeless-auto-complete.ts
+++ b/packages/blocks/src/page-block/edgeless/components/auto-complete/edgeless-auto-complete.ts
@@ -376,7 +376,7 @@ export class EdgelessAutoComplete extends WithDisposable(LitElement) {
     const bound = this._computeNextShape(type);
     const path = this._computeLine(type, this._current, bound);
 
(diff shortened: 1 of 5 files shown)


regischen:
Is it the content or the position that can't be synchronized? In render, there is root's render method, so theoretically, there should be event listeners.

doouding:
Both, like when you drag the note or press enter to add a new paragraph. In the previous version (before `SurfaceBlockComponent` was introduced), everything was placed directly under the `EdgelessPageBlockComponent`, the rerender works well.

I wonder if there's a better approach to solving this issue.

Saul-Mirone:
I think we should merge this pr. Although there may have better solutions, we can improve the implementation later. Anyway, it's a fatal bug. And I think we should add some e2e test for this case.

regischen (approved):

doouding:
Got it. I'll merge after pushing some related fixes.

PR #4809, opened 2023-09-18
fix(virgo): v-range change not trigger update when virgo has v-range provider

### LINKED ISSUE #4769: Linked page no selected style
image

Flrande:
Close #4769


Saul-Mirone:
Will `null` be triggered if selection is moving out of the virgo instance?

PR #4842, opened 2023-09-20
refactor(page): migrate to HoverController

lawvs:


packages/blocks/src/attachment-block/attachment-block.ts
```diff
@@ -39,7 +39,7 @@ export class AttachmentBlockComponent extends BlockElement<AttachmentBlockModel>
   @state()
   private _error = false;
 
-  private _whenHover = new WhenHoverController(this, ({ abortController }) => ({
+  private _whenHover = new HoverController(this, ({ abortController }) => ({
```
Flrande, line 42: The name of variable should be changed as well.

doodlewind (approved):

PR #4848, opened 2023-09-20
feat(page): add spoiler prototype

lawvs:
This is just a prototype, intended to demonstrate the potential of the blocksuite, and is not expected to be merged.

The spoiler feature is useful for preventing spoilers when sharing content, but it has no value in enterprise-level collaboration settings.

Before merging, careful evaluation is necessary.

https://github.com/toeverything/blocksuite/assets/18554747/43be5441-c20d-41b8-96bb-596d5d8c8144


Saul-Mirone:
This looks fun.

doodlewind:
Closing since there's not product plan on shipping this feature. Hopefully this could be a third-party work after the extension  system is designed.

PR #4850, opened 2023-09-20
test: handle block init edge case in collab

himself65:


packages/store/src/__tests__/workspace.unit.spec.ts
```diff
@@ -115,6 +118,35 @@ describe('basic', () => {
     });
   });
 
+  it('can blocks init in 10 rounds', async () => {
```
doodlewind, line 121: I’m not sure if this is a valid operation sequence. What I see here doing is:

1. Init ydoc `workspace1` and its subdoc `page1`, then fill block content into `page1`.
2. Init ydoc `workspace2` and its subdoc `page2`, but `page2` is empty.
3. Apply encoded `workspace1` binary into `workspace2`.
4. Apply encoded `page1` binary into `page2`.

Note that `workspace1`, `workspace2`, `page1`, `page2` all have their underlying ydoc.

In this sequence, seems that `workspace1` and `workspace2` have been diverged (in step 2) before applying binary updates (in step 3). @himself65 is this still a possible sequence in AFFiNE production?

/cc @Saul-Mirone

doodlewind:
Closing since this data flow is not aligned with the design of BlockSuite, cc @joooye34 for potential discussion.

PR #4851, opened 2023-09-20
refactor(store): remove API `waitForLoaded`

himself65:
`waitForLoaded` is a useless function. The loading state should be controlled by the downstream application, not the blocksuite editor.

store should only listen and wait for the block update, instead of extra waiting.

`createPage` is explicitly behavior to **add** data into yDoc, wait for the created page loaded is a weird behavior

Also fixes: https://github.com/toeverything/blocksuite/pull/4850


fourdim (commented):
Please update "packages/docs" as well.

doodlewind (changes requested):
- The current default playground entry http://localhost:5173/?init breaks with this update.
- The undo/redo support of the legacy entry http://localhost:5173/starter/?init also breaks.

doodlewind:
We are looking into #4850, but I'm afraid landing this breaking change is inappropriate. When it comes to subdoc, we still need to expose the underlying async `subdoc.load` to some extent.

Issue #4857, opened 2023-09-21
Should support keeping list toggle state


doodlewind:
Current implementation in #3770 doesn't persist the toggle state.

PR #4865, opened 2023-09-21
fix(page): keeping list toggle state

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/4857

## Changes

- Add `collapsed` to the list model and default to `false`

---

- [x] Keeping list toggle state
- [x] Should expand list after indent collapsed block
- [x] Update snapshot


doodlewind (approved):
@lawvs are you still willing to work on passing the remaining test cases? Would greatly appreciate if it's under your consideration, and please also let me know if not.

doodlewind (approved):
Should land this, thanks for the fix!

PR #4871, opened 2023-09-22
feat(database): add basic animation for database header drag handle

CatsJuice:


packages/blocks/src/database-block/table/table-view.ts
```diff
@@ -54,6 +54,17 @@ const styles = css`
...
+    padding-top: 1.5px;
+  }
+  .affine-database-block-table::before {
+    content: '';
+    position: absolute;
+    left: 0;
+    right: 0;
+    top: 0;
+    height: 1.5px;
+    background-color: var(--affine-border-color);
```
zzj3720, line 67: @zanwei Is this as expected?

zzj3720: 1.5px border at top of table view

zanwei: Based on what we discussed earlier, it should be 1px border at top of table view

CatsJuice: Got it!

zzj3720 (approved):

PR #4877, opened 2023-09-22
feat(edgeless): shape tool drag and animation

### LINKED ISSUE #4538: Edgeless Toolbar optimization

lzlme:
close #4538 

https://github.com/toeverything/blocksuite/assets/117659326/57a29f94-9b32-4539-ad9e-b0379d5e5abd


hwangdev97:
Good job 👍 

There are still some details: 

1. In the click-activated scenario, it is necessary to deal with the problem of super small shapes being created due to slight movement.

I think it would be better to start creating only when the displacement is greater than a certain number, maybe 20px or 50px, 
Depends on specific implementation

shot-2023-09-22-13 57 59@2x

2. The placement of the shape has shifted, which is not quite what we expected. A slight displacement will have the metaphor of letting go, but the current displacement is too large, which is wrong. 

My suggestion is to position it about 4px lower and right than the placeholder.


https://github.com/toeverything/blocksuite/assets/24713927/cb2c7cc2-c293-4737-9b2d-b8323b38628e


3. This bottom shape area needs to be larger

Current preview's width is `92px` design file is `102px`. And Inactive shapes can be moved further down.

shot-2023-09-22-14 08 43@2x

packages/blocks/src/page-block/edgeless/components/toolbar/shape/shape-element.ts
```diff
@@ -0,0 +1,319 @@
...
+  @state()
+  private startCoord: Coord = { x: 0, y: 0 };
+
+  @state()
+  private dragging: boolean = false;
+
+  @state()
+  private isOutside: boolean = false;
+
+  private onDragStart = (coord: Coord) => {
```
doodlewind, line 117: Please add  `_` prefix for private methods and variables.

packages/blocks/src/page-block/edgeless/components/toolbar/shape/shape-element.ts
```diff
@@ -0,0 +1,319 @@
...
+interface TransformMap {
+  [key: string]: {
+    x: number;
+    y: number;
+    scale: number;
+    origin: string;
+  };
+}
+
+@customElement('edgeless-shape-element')
```
doodlewind, line 47: This naming seems to be overkill.

packages/blocks/src/page-block/edgeless/components/toolbar/shape/shape-element.ts
lzlme: done

hwangdev97 (approved):

donteatfriedrice (approved):

doodlewind (approved):
Thanks for polishing this!

Issue #4880, opened 2023-09-22
Code block menu update


zanwei:
We would like to have an update for Code block menu.

Please check [Internal design file](https://www.figma.com/file/UTTtfVCcStw1dKUg0gCSPo/%F0%9F%9B%9E-Design-System?type=design&node-id=9949-30595&mode=design&t=luE3kOMFZX9yKmph-4)

<img width="1381" alt="image" src="https://github.com/toeverything/blocksuite/assets/40333372/96f2dbbd-0c4a-49dc-88e3-52bb2fcf3efe">

PR #4884, opened 2023-09-22
refactor(page): update lang menu styles on code block

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/4880

Screenshot 2023-09-22 at 17 15 40


zanwei:
Thank you for your update!

but there're serval things I need to mention

1.  Please refer to the design of the search result UI. the search results should affect the panel height
image

2. Please refer to the design of the search bar when it's in its default state
image

lawvs:
Because we will automatically focus the input field when clicking the button, I don't think there is a need to hide the search icon when the input is active.

doodlewind (approved):
Aside from the minor icon issue, this improvement already looks good to merge. And we should land this to avoid more potential hand over costs.

PR #4894, opened 2023-09-25
feat(edgeless): support top-level image block inside surface

### LINKED ISSUE #3594: Support top-level image block in whiteboard
The current images must appear in the Note component. It is hoped that images can be inserted independently on the whiteboard. In addition, based on this capability, it is supported to select from the sticker library provided by the design Team, or insert from some common external resources, such as Unsplash.

regischen:
close #3594


Flrande:
image

packages/blocks/src/image-block/image-block.ts
```diff
@@ -19,8 +23,157 @@ import { type ImageBlockModel, ImageBlockSchema } from './image-model.js';
...
+          this._fetchError(new Error('Cannot find blob'));
+        }
+      })
+      .catch(this._fetchError);
+  };
+
+  override connectedCallback() {
+    super.connectedCallback();
+    this._fetchImage();
+    this._disposables.add;
```
lawvs, line 148:
```suggestion
```

doodlewind (commented):
The image dragged from native OS window is still inserted with the note block wrapper.

https://github.com/toeverything/blocksuite/assets/7312949/e9a866e8-e863-4ef8-87de-b95e38d5d820

packages/blocks/src/page-block/edgeless/edgeless-blocks-container.ts
```diff
@@ -166,11 +175,20 @@ export class EdgelessBlockContainer extends WithDisposable(LitElement) {
...
-          <edgeless-notes-container
-            .edgeless=${edgeless}
-            .notes=${sortedNotes}
-            .renderer=${renderModel.bind(this)}
-          ></edgeless-notes-container>
+          ${repeat(
+            blocks,
+            block => block.id,
+            (block, index) => {
+              const tag = literal`${unsafeStatic(
```
doodlewind, line 182: The block flavour here comes from unreliable dynamic content, so using it transparently as HTML tag type seems to be unsafe.

packages/blocks/src/image-block/image-block.ts
```diff
@@ -19,8 +23,157 @@ import { type ImageBlockModel, ImageBlockSchema } from './image-model.js';
...
+        this._fetchImage();
+        // 1s, 2s, 3s
+      }, 1000 * this._retryCount);
+    } else {
+      console.error(e);
+      this._imageState = 'failed';
+    }
+  };
+
+  protected _fetchImage = () => {
```
doodlewind, line 119: Why `protected`? Are these protected methods designed to be inherited?

packages/blocks/src/image-block/image-block.ts
```diff
@@ -19,8 +23,157 @@ import { type ImageBlockModel, ImageBlockSchema } from './image-model.js';
...
+        .root=${this.root}
+        .widgets=${this.widgets}
+        .content=${this.content}
+      ></affine-edgeless-image>`;
+    }
+  }
+}
+
+class ImageBlock extends BlockElement<ImageBlockModel> {
+  static maxRetryCount = 1;
```
doodlewind, line 75: Define this as `MAX_RETRY_COUNT` at the top of this module.

packages/blocks/src/surface-block/surface-block.ts
```diff
@@ -265,6 +315,222 @@ export class SurfaceBlockComponent extends BlockElement<SurfaceBlockModel> {
...
+
+    this._initBlocks();
+  }
+
+  private _initBlocks() {
+    const { page } = this;
+    this.blocks.forEach(block => {
+      this._addToBatch(block);
+    });
+    page.slots.blockUpdated.on(e => {
```
doodlewind, line 327: Why this is not wrapped in `this._disposables.add()`?

packages/blocks/src/image-block/image-model.ts
```diff
@@ -24,6 +40,33 @@ export const ImageBlockSchema = defineBlockSchema({
     role: 'content',
   },
   transformer: () => new ImageBlockTransformer(),
+  toModel: () => {
+    return new ImageBlockModel();
+  },
 });
 
-export type ImageBlockModel = SchemaToModel<typeof ImageBlockSchema>;
+@RectElement
```
doodlewind, line 48: This decorator duplicates the concept of *element*. Is it indicating top-level content in edgeless mode? If so, it's also weird that shape, connector and frame are not decorated by it. Could you elaborate?

packages/blocks/src/__internal__/clipboard/edgeless-clipboard.ts
```diff
@@ -184,8 +238,29 @@ export class EdgelessClipboard implements Clipboard {
...
+      sourceIds: string[];
+      images: string[];
+    } = {
+      data: data.data,
+      sourceIds: data.sourceIds,
+      images: [],
+    };
+    if (data.imageBlobs.length > 0) {
+      copyData.images = await Promise.all(
+        data.imageBlobs.map(blob => blobToBase64(blob))
```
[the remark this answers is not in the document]
regischen: this was intended to fix #4967 , but never mind, I am intented leave it to another pr.
In short, I find this way to copy multiple images. 
no, the base64 doesn't exist in the block tree.

packages/blocks/src/image-block/image-model.ts
regischen: It is Mixin, some utils of rect-like. so changing a name would be great.

packages/blocks/src/page-block/edgeless/edgeless-blocks-container.ts
regischen: in this case, the flavour is one of note or image, so I think it's safe.

packages/blocks/src/page-block/edgeless/components/image/edgeless-image.ts
```diff
@@ -0,0 +1,57 @@
...
+import { LitElement } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
+import { styleMap } from 'lit/directives/style-map.js';
+import { html } from 'lit/static-html.js';
+
+import type { ImageBlockModel } from '../../../../image-block/index.js';
+import { Bound } from '../../../../surface-block/index.js';
+import type { SurfaceBlockComponent } from '../../../../surface-block/surface-block.js';
+
+@customElement('edgeless-image')
```
doodlewind, line 11: Now we have  both `edgeless-image` and `affine-edgeless-image` in the codebase, which is duplicate. How to differentiate them?

doodlewind (approved):
Should land this after the cleanups, thanks for the feat.

Issue #4903, opened 2023-09-26
Code block unexpected horizontal scrolling when click blank area on the side


zanwei:
Code has incorrect indentation when closing code selection menu

https://github.com/toeverything/blocksuite/assets/40333372/bfc139d2-f999-4172-b7fa-ab71783e565d

Flrande:
Here is a more simple reproduction.

https://github.com/toeverything/blocksuite/assets/50035259/f2dbe958-8b79-4eb9-ac59-8c7b4216179b

zqran:
Just click outside the code-block to reproduce:

https://github.com/toeverything/blocksuite/assets/15389209/1a8579be-fdea-430d-892d-848a476c0bc7

Issue #4403, opened 2023-08-28
Support range provider in Virgo


Flrande:
At present, Virgo autonomously synchronizes the browser's native range with its v-range. Yet, there are instances when we prefer to synchronize them outside of Virgo to maintain a single data flow.

Currently, Blocksuite incorporates a three-way synchronization involving SelectionManager, Virgo, and the native range. (You can see it in [range-synchronizer](https://github.com/toeverything/blocksuite/blob/274405a6d09ba2a4e80454ebb0fef460b37f3da1/packages/lit/src/utils/range-synchronizer.ts#L17)) This setup not only complicates code maintenance but also poses challenges in implementing certain features, such as #4377.

Flrande:
This issue block #4376

PR #4905, opened 2023-09-26
fix: `code-block` wrong cursor position

zqran:
close: #4903


Flrande (changes requested):
I don't think it is a good fix. You can see #4403 to know why we add `vRangeProvider`.

You should know where this bug from and fix it from the source.

https://github.com/toeverything/blocksuite/assets/50035259/2d0a742a-66ae-4eec-baf8-5e80791d3a87

Flrande:
see #4915

Issue #4907, opened 2023-09-26
When hovering over the list content, the Toggle icon for each row is displayed


houjoe0829:
This is not as expected.

![CleanShot 2023-09-26 at 11 14 14@2x](https://github.com/toeverything/blocksuite/assets/22443345/806f7328-297a-4ec4-8107-12acab639120)

Expected effect(an sample)

https://github.com/toeverything/blocksuite/assets/22443345/2f4046a3-c362-4930-a95e-4cb77bbd2eed

PR #4913, opened 2023-09-26
fix(page): display of toggle icon of `list-block`

zqran:
close: #4907

https://github.com/toeverything/blocksuite/assets/15389209/9a341a50-d631-4b48-9b35-a119577653eb

### DIFF
--- a/packages/blocks/src/list-block/styles.ts
+++ b/packages/blocks/src/list-block/styles.ts
@@ -47,16 +47,16 @@ const toggleStyles = css`
   .toggle-icon:hover {
     background: var(--affine-hover-color);
   }
-  .affine-list-block-container:hover .toggle-icon {
+  .affine-list-rich-text-wrapper:hover .toggle-icon {
     opacity: 1;
   }
   .toggle-icon__collapsed {
     opacity: 1;
   }
 
-  .with-drag-handle .toggle-icon {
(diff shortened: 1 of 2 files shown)


lawvs:
When hovering over the parent, the Toggle icon for children should not be shown. However, I suppose that when hovering over the children, the parent's toggle icon should be shown.

@houjoe0829 What do you think?

doodlewind (approved):
Good catch!

donteatfriedrice:
It seems that the toggle icon should be visible when hover on drag handle. See #3770.

https://github.com/toeverything/blocksuite/assets/99816898/f44bc935-1bdf-4b84-9ff7-5c102221327c

zqran:
> It seems that the toggle icon should be visible when hover on drag handle. See #3770.
> 
>  Screen.Recording.2023-09-27.at.10.52.03.mov

@houjoe0829 In this case, should the child's icon be displayed?

lawvs:
> > It seems that the toggle icon should be visible when hover on drag handle. See #3770.
> > Screen.Recording.2023-09-27.at.10.52.03.mov
> 
> @houjoe0829 In this case, should the child's icon be displayed?

@hwangdev97 believes that this is necessary https://github.com/toeverything/blocksuite/pull/3770#issuecomment-1717508924

lawvs (approved):
Great works!

Issue #4904, opened 2023-09-26
Unable to use enter key in image caption


houjoe0829:
Currently, the Enter key cannot be used in the caption of an image. Pressing the key will not have any response. 

The expected result is that if a line break is needed in the middle of the caption, a new text block will be created after the image, and the content after the cursor will be moved to the new text block. If a line break is needed at the end of the caption, a new blank text block will be created directly after the image.

https://github.com/toeverything/blocksuite/assets/22443345/92de8a74-9438-4273-b6f5-d1349b5f0013

Flrande:
cc? @zqran

PR #4928, opened 2023-09-27
feat(page): supports the `enter` key for image caption

zqran:
close: #4904

https://github.com/toeverything/blocksuite/assets/15389209/4cf6e1f5-eef2-4f42-aa0d-e93a1bc1a709

### DIFF
--- a/packages/blocks/src/image-block/image-block.ts
+++ b/packages/blocks/src/image-block/image-block.ts
@@ -3,10 +3,12 @@ import './image/placeholder/loading-card.js';
 
 import { PathFinder } from '@blocksuite/block-std';
 import { BlockElement } from '@blocksuite/lit';
+import { Text } from '@blocksuite/store';
 import { css, html, type PropertyValues } from 'lit';
 import { customElement, query, state } from 'lit/decorators.js';
 import { styleMap } from 'lit/directives/style-map.js';
 
+import { asyncFocusRichText } from '../__internal__/utils/common-operations.js';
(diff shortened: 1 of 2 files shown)


packages/blocks/src/image-block/image-block.ts
```diff
@@ -411,6 +413,37 @@ export class ImageBlockComponent extends BlockElement<ImageBlockModel> {
     };
   }
 
+  private _onCaptionKeydown(e: KeyboardEvent) {
+    if (e.key === 'Enter') {
```
Flrande, line 417: Need to check `e.isComposing`

packages/blocks/src/image-block/image-block.ts
```diff
@@ -411,6 +413,37 @@ export class ImageBlockComponent extends BlockElement<ImageBlockModel> {
...
+      if (!parent) return;
+      const index = parent.children.indexOf(model);
+      const id = page.addBlock(
+        'affine:paragraph',
+        { text: new Text(nextBlockText) },
+        parent,
+        index + 1
+      );
+      asyncFocusRichText(model.page, id, {
+        index: nextBlockText.length,
```
Flrande, line 441: I think it will be more reasonable to use `0` to keep index when we split the text.

Flrande:
Can you add a test?

packages/blocks/src/image-block/image-block.ts
zqran: done

PR #4933, opened 2023-09-28
fix(page): delete nested item in the middle should not change the hierarchy of its siblings below

### LINKED ISSUE #4864: Deleting nested list item in the middle should not change the hierarchy of its siblings below
This is mainly related to indented lists and their impact on other content.

The main purpose is to avoid changing the hierarchy of same-level content below a blank line when deleting it in the middle, which is inconsistent with common editing behavior.

How to reproduce:
1. When there is same-level content below indented content, such as "Child 2" in the screenshot.
2. Starting from the end, click the delete key to delete until it becomes a blank line.
3. Deleting again will reduce the indentation level and affect the indentation of "Child 3" below.

CleanShot 2023-09-21 at 17 47 25@2x

The expected effects are:

In step 2, after deleting to a blank line, clicking the delete key again will directly delete it.
In step 2, if deleting starts from the beginning, the indentation level will not be reduced, but the content will be merged into the previous line.

This is the effect I recorded in Notion：

https://github.com/toeverything/blocksuite/assets/22443345/a51068b7-2ad3-4d72-a965-2f15e1b14df3

### LINKED ISSUE #4944: Can not delete forward in the start of first paragraph
https://github.com/toeverything/blocksuite/assets/50035259/65cc20c9-8ad0-45bc-9b3f-9786ba2014e9

Flrande:
Close #4944
Close #4864

### DIFF
--- a/packages/blocks/src/components/rich-text/rich-text-operations.ts
+++ b/packages/blocks/src/components/rich-text/rich-text-operations.ts
@@ -1,5 +1,6 @@
 import { assertExists } from '@blocksuite/global/utils';
-import type { BaseBlockModel, Page } from '@blocksuite/store';
+import type { Page } from '@blocksuite/store';
+import { type BaseBlockModel } from '@blocksuite/store';
 import { Text } from '@blocksuite/store';
 
 import { supportsChildren } from '../../__internal__/utils/common.js';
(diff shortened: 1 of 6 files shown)


doodlewind (approved):
Great improvement!

PR #4934, opened 2023-09-28
fix: block hub z-index

### LINKED ISSUE #4820: Incorrect block hub entry tooltip z-index in edgeless
image

lzlme:
fix #4820

### DIFF
--- a/packages/blocks/src/components/block-hub.ts
+++ b/packages/blocks/src/components/block-hub.ts
@@ -192,7 +192,7 @@ export const BLOCKHUB_FILE_ITEMS = [
 const styles = css`
   affine-block-hub {
     position: absolute;
-    z-index: 1;
+    z-index: 3;
     user-select: none;
   }
 


doodlewind (approved):

packages/blocks/src/components/block-hub.ts
```diff
@@ -192,7 +192,7 @@ export const BLOCKHUB_FILE_ITEMS = [
 const styles = css`
   affine-block-hub {
     position: absolute;
-    z-index: 1;
+    z-index: 3;
```
doodlewind, line 195: Do we have sementic css variable for this? Maybe it could be better to use the variable together with `calc`.

lzlme: We have some variables like`--affine-z-index-xx`, but I used the magic number `3` here because I found that the z-index of &lt;edgeless-toolbar&gt; was written as 3, but because it was nested too deeply, I didn’t dare to modify it easily. After that, we should change all the z-index index is unified into variables.

PR #4939, opened 2023-09-28
perf(edgeless): remove edgeless CSS variable to increase performance

doouding:
Zoom in/out or change the viewport of the edgeless will update the CSS variables like `--affine-edgeless-x`, `--affine-edgeless-y`, `--affine-zoom`, and so on. Each time those variables are updated, the browser needs to recalculate all its descendants' styles even if those descendants does not use any of the variable.

The solution is simple, remove the variable and just use those values in JavaScript directly.

**Before**
Screenshot 2023-09-28 at 4 48 10 PM

**After**
Screenshot 2023-09-28 at 4 50 05 PM

**Change**
- remove edgeless viewport CSS variables, using `translateX`, `translateY`, and `zoom` from `edgeless.surface.viewport` along with the `edgeless.slots.viewportUpdated` should be sufficient for most scenarios.
- slicer bug after refactoring of surface block


doodlewind (approved):
Thanks for the optimization!

PR #4941, opened 2023-09-28
feat(edgeless): optimize hollow shape elements hit test

### LINKED ISSUE #4841: Support hit test optimization for hollow shape elements
For shape elements without fill color, it would be better to mark its center area as selectable, so that the drag experience can be optimized.

Would be great if @SydianAndrewChen can take a look.

donteatfriedrice:
To close #4841 

- [ ] Need to confirm what ratio of the center area is selectable.

https://github.com/toeverything/blocksuite/assets/99816898/75e28efa-74f7-48c5-8072-899421137dcc


doodlewind (approved):
Nice improvement, thanks!

doodlewind (approved):

PR #4948, opened 2023-10-01
fix(page): surface element break path calculate in page mode

Flrande:


Flrande:
I will add test realated `ArrowUp` and `ArrowDown` in #4949, which will cover this case.

doodlewind (approved):

Issue #3744, opened 2023-07-13
​The tooltips in the editor are blocked


JimmFly:
### What happened?

upstream: https://github.com/toeverything/AFFiNE/issues/2406
![image](https://github.com/toeverything/AFFiNE/assets/102217452/1ae81028-59e4-44f3-88b5-ad52d6def582)

### Distribution version

macOS x64

### What browsers are you seeing the problem on if you're using web version?

Chrome, Other

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

Yipei-Operation:
update: under {full width layout},the code block has some uiux bug,sidebar could not be shown:
<img width="735" alt="截屏2023-07-28 13 58 57" src="https://github.com/toeverything/blocksuite/assets/79373028/ef44abcb-0a1b-4b6a-b88e-7c4148e261f3">
correct case should be:
<img width="931" alt="截屏2023-07-28 13 59 55" src="https://github.com/toeverything/blocksuite/assets/79373028/1eaad6db-94b4-4734-947e-1f6aa8b307c8">
(version: 0.7.0 stable)

doodlewind:
@xell @hwangdev97 what's the expected behavior in this case? We're working on related UI fixes.

xell:
The tooltip should display in the opposite, i.e. left, side if there is no full room in the right side.

doodlewind:
LGTM, so I guess @lawvs could take higher priority fixing this.

PR #4950, opened 2023-10-01
refactor(page): switch to `affine-tooltip`

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/3744

Sunsetting `tool-tip`. The new `affine-tooltip` is a portal, so it will not be affected by overflow, and it uses the floating API, ensuring that it won't exceed the window.

https://github.com/toeverything/blocksuite/assets/18554747/b928943a-94ee-403e-b688-8b0ce51e7df3

## Motivation

The legacy `tool-tip` was introduced at https://github.com/toeverything/blocksuite/pull/178 last year. The `tool-tip` is so simple, it only has a few styles. It works well but can be blocked or extend beyond the window.

## Migrate guide

```diff
- <icon-button class="has-tool-tip" style="${tooltipStyle}">
-   <tool-tip inert tip-position="top" role="tooltip">Preview</tool-tip>
+ <icon-button>
+   <affine-tooltip tip-position="top">Preview</affine-tooltip>
</icon-button>
```

## API

```ts
 // Simple usage
 html`
 <blocksuite-tooltip>Content</blocksuite-tooltip>
 `
 // With placement
 html`
 <affine-tooltip tip-position="top">
   Content
 </affine-tooltip>
 `

 // With custom properties
 html`
 <affine-tooltip
   .zIndex=${0}
   .offset=${4}
   .autoFlip=${true}
   .arrow=${true}
   .tooltipStyle=${css`:host { z-index: 0; --affine-tooltip: #fff; }`}
   .allowInteractive=${false}
 >
   Content
 </affine-tooltip>
 `
```

## Drawbacks

In order to make the tooltip usage simpler, the content in the `slot` will be **cloned** to the portal. This causes the tooltip to be unable to show dynamic content and cannot add event listeners.

```html
// Need to clone the static slot content
<affine-tooltip>slot content</affine-tooltip>

// No above-mentioned drawbacks, but it decreases developer-friendliness.
<affine-tooltip .template=${html`content`}></affine-tooltip>
```

### DIFF
--- a/packages/blocks/src/attachment-block/components/options.ts
+++ b/packages/blocks/src/attachment-block/components/options.ts
@@ -62,15 +62,13 @@ export function AttachmentOptionsTemplate({
     >
       <icon-button size="24px" ?hidden=${true}>
         ${ViewIcon}
-        <tool-tip inert tip-position="top" role="tooltip">Preview</tool-tip>
+        <blocksuite-tooltip>Preview</blocksuite-tooltip>
       </icon-button>
       <div class="divider" ?hidden=${true}></div>
 
       <icon-button size="24px" ?hidden=${true || readonly}>
         ${LinkIcon}
-        <tool-tip inert tip-position="top" role="tooltip"
-          >Turn into Link view</tool-tip
-        >
+        <blocksuite-tooltip>Turn into Link view</blocksuite-tooltip>
       </icon-button>
       <icon-button
         size="24px"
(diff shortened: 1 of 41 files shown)


lawvs:
> Regarding the naming, I think it's a mistake to have `affine-foo` and `blocksuite-foo` components coexisting, which is confusing. IMO it would be better to stick to `affine-foo` naming, since BlockSuite is providing the official implementation for the affine scope.

The proposal about prefixes starts from https://github.com/toeverything/blocksuite/pull/3830#discussion_r1284248140, I think @Saul-Mirone will be able to help.

Saul-Mirone (approved):

doodlewind (approved):
Let's land this, thanks!

PR #4956, opened 2023-10-03
refactor(page): enhance `rich-text` and replace `virgo-input` with `rich-text`

### LINKED ISSUE #4858: Copy is not working
https://github.com/toeverything/blocksuite/assets/3468483/4f9b0f37-da4c-4280-8ddb-61b2dc3960b8

Flrande:
With the provision of `rich-text` capabilities, basic logic operations such as undo-redo, copy, paste, and cut are inherently handled, obviating the need for higher-level intervention. They can cover all text input scenarios and make developers no need to mount Virgo manually.

Close #4858


packages/blocks/src/components/plain-text/plain-text.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  override render() {
+    return html`<div class="virgo-container"></div>`;
+  }
+}
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'plain-text': PlainText;
```
lawvs, line 183: Since all custom elements share the same namespace, it is suggested to add a prefix to prevent unexpected name conflicts with application developers.

Flrande: It may need to add prefix for all blocksuite components. (worth creating a separate issue)

doodlewind (commented):
Do you think the current `rich-text` component deserves a standalone introduction section in the virgo README documentation? It will be great to do this as long as you consider it has well-defined API surface.

doodlewind (approved):

Flrande:
> Do you think the current `rich-text` component deserves a standalone introduction section in the virgo README documentation? It will be great to do this as long as you consider it has well-defined API surface.

Sure, comparing with Virgo, directly using `rich-text` may be more developer-friendly. I will update doc and do some name changes in this PR.

PR #4965, opened 2023-10-08
feat(page): safe triangle for `whenHover` API

lawvs:
Implements the safe triangle feature that was mentioned in https://github.com/toeverything/blocksuite/pull/4689

It is worth noting that this new feature has almost no significance for blocksuite now, so this feature is closed by default. It can be enabled by adding `safeTriangle` options in the `whenHover` API.

https://github.com/toeverything/blocksuite/assets/18554747/580cdec4-b4f4-4dd1-bcd5-066f6dd47472

Additionally, this PR added a virtual safe bridge between the reference element and the floating element. This feature is enabled by default and can be disabled by add`safeBridge` options in the `whenHover` API.

https://github.com/toeverything/blocksuite/assets/18554747/6869f7ee-e703-4a0d-9fac-25cf734f3f7f

## References

- https://www.smashingmagazine.com/2023/08/better-context-menus-safe-triangles/
- https://floating-ui.com/docs/usehover#safepolygon
- https://github.com/shoelace-style/shoelace/pull/1600


packages/global/src/utils/index.ts
```diff
@@ -1,6 +1,7 @@
 export * from './assert.js';
 export * from './disposable.js';
 export * from './function.js';
+export * from './hover/index.js';
```
doodlewind, line 4: I don't think it's appropriate to put this in the `global` package, which should only contain essential non-UI related primitives and utils.

lawvs: Now the API moved to `blocks/components/hover`

```
components
├── controller.ts <- HoverController
├── index.ts
├── middlewares
│   ├── basic.ts
│   └── safe-area.ts
├── types.ts
└── when-hover.ts
```

Issue #4925, opened 2023-09-27
Standardize the `ESC` and Shortcut operation behavior


zanwei:
1. Pressing ESC should interrupt the process of `Pen/Eraser/Shape/Text` tools and let Edgeless return to its default mouse state

https://github.com/toeverything/blocksuite/assets/40333372/6166a19d-5637-4371-983c-250b12c98236

2. When using the shortcut key `S` to evoke Shape, the secondary panel must also appear at the same time

https://github.com/toeverything/blocksuite/assets/40333372/3beca2c5-51e4-4889-a6ce-8b6ef62b9ed5

PR #4980, opened 2023-10-10
feat(edgeless): open secondary menu when evoke shape tool

lzlme:
Part of #4925 
When using the shortcut key S to evoke Shape, the secondary panel must also appear at the same time.


donteatfriedrice (approved):

doodlewind (approved):
Thanks for the improvement!

PR #4982, opened 2023-10-10
feat: convert xywh to native array

### LINKED ISSUE #4167: Switch to native plain object for note size and phasor elements
To make efficient use of Yjs, switching to native type would make a lot of sense. Since the upstream API is supported in #4020, there are 2 parts that can be planned to be done recently:

1. For page block, switch from `xywh: '[0,0,100,100]'` to `xywh: number[]`.
2. For phasor elements, switch to plain JSON object per element, instead of using their Y-alternatives. We don't have to keep `Y.Text` inside the canvas text element, or so on.

/cc @Saul-Mirone @Flrande @regischen @doouding for potential discussion.

doouding:
close #4167 

###  yDoc encoded size 

Measure size of encoded yDoc using methods `encodeStateAsUpdate` and `encodeStateAsUpdateV2`  (tested on heave-whiteboard preset):

| Conversion strategy   | encodeStateAsUpdate | encodeStateAsUpdateV2
|:----------|:-------------|:-------------|
|No conversion |598|459|
|`xywh` property |629|484|
|`xywh` property and convert surface element to plain object|629|484|

### update patch size

Move 100 blocks and 100 shapes 1px simultaneously and measure the update patch size receiving from yDoc `update` event:

| Conversion strategy   | Patch size (bytes) |
|:----------|:-------------|
| No conversion |15268|
| `xywh` property |16884|


doouding:
It seems this would increase the update size a little bit @doodlewind

doodlewind:
Another important metrics is about measuring the binary size after batch dragging, trying on that should also help.

doouding:
Just updated patch size metrics.

doodlewind:
Seems that (most possibly) due to the existence of `NativeWrapper` (cc @Saul-Mirone), switching to the plain object here doesn't worth the desired effect. Also, the rich text support in TextElement also limits the potential switch to element-level plain object. So we should not move forward landing this PR.

But the metrics is after all very important to the project, thanks for the trying! The automated performance testing suite should also be on our plan, like [what figma built](https://www.figma.com/blog/keeping-figma-fast/).

doouding:
While I was working on the metrics of update patch size, I found moving a bunch of blocks simultaneously is pretty laggy. This might be related to the serialization and deserialization of the xywh property.
I wrote some benchmark tests. Rather than using JSON directly, I wrote two functions to serialize and deserialize the xywh property manually. Here's the result:

**Serialization**
Screenshot 2023-10-19 at 12 39 11 PM

**Deserialization**
Screenshot 2023-10-19 at 12 40 07 PM

Do you think it's worth continuing with this matter?

PR #4994, opened 2023-10-12
feat(edgeless): standardize edgeless toolbar esc behavior

### LINKED ISSUE #4925: Standardize the `ESC` and Shortcut operation behavior

lzlme:
close #4925

### DIFF
--- a/packages/blocks/src/page-block/edgeless/components/toolbar/brush/brush-tool-button.ts
+++ b/packages/blocks/src/page-block/edgeless/components/toolbar/brush/brush-tool-button.ts
@@ -125,6 +125,14 @@ export class EdgelessBrushToolButton extends WithDisposable(LitElement) {
     super.disconnectedCallback();
   }
 
+  override firstUpdated() {
+    this._disposables.addFromEvent(document, 'keydown', event => {
+      if (event.key === 'Escape' && this.edgelessTool.type === 'brush') {
+        this.setEdgelessTool({ type: 'default' });
+      }
+    });
+  }
+
   override render() {
     const type = this.edgelessTool?.type;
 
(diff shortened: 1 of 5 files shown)


packages/blocks/src/page-block/edgeless/components/toolbar/brush/brush-tool-button.ts
```diff
@@ -125,6 +125,14 @@ export class EdgelessBrushToolButton extends WithDisposable(LitElement) {
     super.disconnectedCallback();
   }
 
+  override firstUpdated() {
+    this._disposables.addFromEvent(document, 'keydown', event => {
```
doodlewind, line 129: Listening events from document may be overkill, which doesn't work well handling multiple editor instanes. @Saul-Mirone do we have API example for subscribing events within editor scope?

Saul-Mirone: Yep, we have the `bindHotKey` method on block element.

lzlme: I modified~

doodlewind (approved):

PR #5002, opened 2023-10-13
fix: enter key should as expected after setting heading by shortkey

### LINKED ISSUE #4987: After using shortcut keys to set headings, using the Enter key will clear the content
The following screen recording is the effect after using the setting Heading 1 in the Internal version.

https://github.com/toeverything/blocksuite/assets/22443345/bdf80e43-f7cf-4858-8c6b-c8c839d49203

lzlme:
fix #4987


packages/blocks/src/page-block/utils/callback.ts
```diff
@@ -17,10 +17,13 @@ export async function onModelTextUpdated(
   await richText.updateComplete;
   const vEditor = richText.vEditor;
   assertExists(vEditor, 'VEditor is not ready yet.');
-  vEditor.slots.updated.once(() => {
-    if (callback) {
-      callback(richText);
-    }
+  await new Promise<void>(resolve => {
```
doodlewind, line 20: This seems to be simply a `queueMicrotask` alternative. But semantically here shall we wait for an extra tick here? This may lose emitted events and potentially break other behavior.

lzlme: Yes, I'll look into it further.

Flrande (approved):

PR #5008, opened 2023-10-13
feat(edgeless): frame image background

### LINKED ISSUE #4707: Frame background settings
[the figma link](https://www.figma.com/file/rfH3P9Wbz4IEp2Ai18O8dV/%E2%96%B6%EF%B8%8F-Affine.2023.Late?type=design&node-id=8345%3A5605&mode=design&t=K8ZO7ckxVIUbiLdd-1)

new edgeless item's toolbar  for frame's background settings.

regischen:
close #4707 

image


doodlewind:
It's good to have this as prototype, but I don't think this is expected to be shipped in next release. Actually, having none styled frame matches the new AFFiNE usage better.

doodlewind (changes requested):
The current design introduces 3 different fields for frame background:

1. `background` for background color.
2. `source` for background image.
3. `gradient` for background gradient.

These fields overlaps, so when setting a frame with image background and then switch to background color, all these fields co-exist without clarified priority:

image

Another major problem is that the `source` doesn't follow the pattern in image block, having a blob id with `localhost` doesn't make sense to collaborate.

Also, the UX using background looks not intuitive - for example, why only supports adding gradient for image?

We should not land this until the design is clarified.

doodlewind:
As as starting point, having a new PR only having background color is more acceptable, which is not necessarily to land in this iteration scope. I can see this current implementation looks too perfunctory and may bring unexpected debts.

regischen:
> The current design introduces 3 different fields for frame background:
> 
> 1. `background` for background color.
> 2. `source` for background image.
> 3. `gradient` for background gradient.
> 
> These fields overlaps, so when setting a frame with image background and then switch to background color, all these fields co-exist without clarified priority:
> 
> image
> Another major problem is that the `source` doesn't follow the pattern in image block, having a blob id with `localhost` doesn't make sense to collaborate.
> 
> Also, the UX using background looks not intuitive - for example, why only supports adding gradient for image?
> 
> We should not land this until the design is clarified.

image

image

actually `source` is not in the crdt of frame-model.
frame-model is a custom model.  `source` is a runtime-property which is useful to semantically  develop

PR #5013, opened 2023-10-13
feat(edgeless): add surface ref block

### LINKED ISSUE #5010: Support partially insert group and frame into doc mode

doouding:
Close #5010

### Change
- support edgeless content in page mode
- add a new type of block `affine:surface-ref` to hold edgeless content in doc mode

### Design

The `affine:surface-ref` block is designed to be the direct child of the page, which means it has the same block level as the `affine:note` block. The `reference` property is the id of the surface element/children that are expected to be displayed in doc mode. 
In practice, any edgeless content can be displayed in doc mode through this block including shape, canvas text, brush, etc.

```typescript
export type SurfaceRefProps = {
  reference: string;
};

export const SurfaceRefBlockSchema = defineBlockSchema({
  flavour: 'affine:surface-ref',
  props: () => ({
    reference: '',
  }),
  metadata: {
    version: 1,
    role: 'hub',
    parent: ['affine:page'],
  },
});

export type SurfaceRefBlockModel = SchemaToModel<typeof SurfaceRefBlockSchema>;
```

Regarding the rendering implementation, the `affine:surface-ref` block holds a surface renderer to render surface elements and creates a portal-like renderer for the `affine:note` block to render notes inside a frame.
It will interpret the reference property and set the viewport according to the referenced element's `XYWH` property every time the frame changes.

```typescript
export class SurfaceSyncBlockComponent extends BlockElement<SurfaceRefBlockModel> {
  private refreshViewport() {
    if (!this._referenceModel) {
      return;
    }

    this._surfaceRenderer.onResize();
    this._surfaceRenderer.setViewportByBound(
      Bound.fromXYWH(deserializeXYWH(this._referenceModel.xywh))
    );
  }

  override render() {
    return html`
      <div class="affine-surface-ref">
        <div class="surface-viewport">
          <div class="surface-block-portal">
            ${this._renderEdgelessNotes(notes)}
          </div>
          <div class="surface-canvas-container">
            <!-- attach surface canvas here -->
          </div>
        </div>
      </div>`;
  }
}
```

There are still some things need to be discuss:
- The role of the `affine:surface-ref` is set to be `hub`, cause the `content` block cannot be the children of the `root` block. But the `hub` seems a little inappropriate.

### Todo
- [x] Toolbar in doc mode
    - [x] Caption support
    - [x] Download/Copy as an image
- [ ] Popup window (like image)
- [ ] Paste surface content in doc mode
- [x] Toolbar in edgeless mode
    - [x] Visible button in the toolbar
- [x] Fallback when content has been deleted

There are some remains:
- Popup window should be implemented after bumping into AFFiNE
- Copy-paste functionality will be tracked separately because the code is a bit confused
- The `surface-ref` block can't be dragged using the drag-handler so far, it is necessary to consult with the designers.


doouding:
I just updated the design doc. Feel free to post any comments. @doodlewind

doodlewind (commented):
> The role of the `affine:surface-ref` is set to be hub, cause the content block cannot be the children of the root block. But the hub seems a little inappropriate.

@Saul-Mirone do you think we can break the previous assumption? In this case, `affine:page` could have both `affine:note` or `affine:surface-ref` as its direct children.

packages/lit/src/utils/range-synchronizer.ts
```diff
@@ -140,7 +140,9 @@ export class RangeSynchronizer {
     if (!range) return;
 
     const blocks = this._rangeManager.getSelectedBlockElementsByRange(range, {
-      match: element => element.model.role === 'content',
+      match: element =>
+        element.model.role === 'content' ||
+        element.model.flavour === 'affine:surface-ref',
```
doodlewind, line 145: I think it's anti-pattern to hard-code `affine` related logic in the lit package. @Saul-Mirone any advice?

packages/lit/src/element/block-element.ts
```diff
@@ -145,10 +145,14 @@ export class BlockElement<
     return result;
   }
 
+  calculatePath = () => {
```
Saul-Mirone, line 148: Why you need to expose this method?

packages/lit/src/utils/range-synchronizer.ts
Saul-Mirone, line 145: We shouldn't sync this block for native selection. Consider handle the edge case in `Delete` keymap.

Saul-Mirone (changes requested):
We shouldn't hard code logic in lit and std package.

Saul-Mirone (approved):

Saul-Mirone:
It's okay to refactor later.

doouding:
FIXME comment has been added.

doodlewind:
Thanks for the feat, looking forward to the upcoming housekeeping updates!

doouding:
I will add some tests later on.

PR #5019, opened 2023-10-13
fix(page): `remote-selection` make other users can not select text by mouse click

### LINKED ISSUE #4998: When collaborating in page mode, when no one performs a select-all operation, all content is selected
As shown in the video, the text content appears to be all selected, and others cannot perform any operations on the text, but in fact no one operates to select all. This situation often disappears when the screen recording starts, so only the first few seconds are captured.

Image

Flrande:
Close #4998


lawvs (changes requested):
You can set `pointer-events: none;` in `remote-selection` directly

lawvs (approved):

PR #5048, opened 2023-10-18
fix(edgeless): fix frame title when inputting Chinese use Enter key

luke358:
https://github.com/toeverything/blocksuite/assets/48149577/a54f6227-2536-43d5-b23d-77f6c72f9ba0


Flrande (changes requested):
We should keep characters from IME and not unmount editor when we press Enter.

https://github.com/toeverything/blocksuite/assets/50035259/a6c0ad6b-5c47-40d1-8a96-b736e6ff7d21

Flrande (approved):

PR #5050, opened 2023-10-18
fix: fix and refactor page-level migration associated with surface block

Flrande:


packages/blocks/src/api/transformer/zip.ts
```diff
@@ -78,6 +78,7 @@ export async function importPages(workspace: Workspace, imported: Blob) {
     slots.afterImport.on(payload => {
       if (payload.type === 'page') {
         workspace.schema.upgradePage(
+          0,
```
Saul-Mirone, line 81: This seems not right.

Saul-Mirone (changes requested):

Saul-Mirone (approved):

doodlewind (approved):

PR #5053, opened 2023-10-18
docs: add editor state and improve unidirectional data flow

Saul-Mirone:


packages/docs/.vitepress/config.ts
```diff
@@ -47,12 +47,18 @@ export default defineConfig({
         ],
       },
       {
-        text: 'AFFiNE Editor Overview',
+        text: 'BlockSuite Foundation',
```
doodlewind, line 50: Love this outline design 😄

PR #5062, opened 2023-10-19
fix(edgeless): fix `copyAsPng` not work

zqran:


packages/blocks/src/page-block/edgeless/controllers/clipboard.ts
```diff
@@ -178,7 +178,7 @@ export class EdgelessClipboardController implements ReactiveController {
       return;
     }
 
-    const json = await this.std.clipboard.readFromClipboard(data);
+    const json = this.std.clipboard.readFromClipboard(data);
```
Saul-Mirone, line 181: Why remove the await here?

zqran: Because this method is not async function:

https://github.com/toeverything/blocksuite/blob/master/packages/block-std/src/clipboard/index.ts#L181-L193

doodlewind (approved):

PR #5064, opened 2023-10-19
feat(edgeless): support more connector arrow style

donteatfriedrice:
- [x] Support changing start point arrow style.
- [x] Support changing end point arrow style.
- [x] Support switching start and end point style.  

https://github.com/toeverything/blocksuite/assets/99816898/96a0f9e5-f610-4b66-910d-9091f15b0cfb


packages/blocks/src/page-block/edgeless/connector-manager.ts
```diff
@@ -969,7 +969,7 @@ export class EdgelessConnectorManager {
       const start = this._getConnectorEndElement(connector, 'source');
       const end = this._getConnectorEndElement(connector, 'target');
 
-      const [startPoint, endPoint] = this._computeStartEndPoint(connector);
+      const [startPoint, endPoint] = this._computeStartEndpoint(connector);
```
doodlewind, line 972: Note that "endpoint" itself is a non-synthetic word, which can be confused with the `endPoint` here (within the context of `startPoint` / `endPoint`). Could you try to avoid the `startPoint` / `endPoint` naming to 
reduce the potential cost of understanding?

doodlewind (approved):

packages/blocks/src/page-block/edgeless/connector-manager.ts
donteatfriedrice: Sure, I'll try to avoid it.

donteatfriedrice:
Need to update diamond, triangle, arrow to add rounded corners.

shot-2023-10-24-17 30 29@2x

hwangdev97:
shot-2023-10-25-15 58 38@2x

Please use new design of triangle

PR #5066, opened 2023-10-19
fix(page): user should be able to expand toggle list in readonly mode

lawvs:
Users should be able to expand the toggle list even in readonly mode.

Blame to https://github.com/toeverything/blocksuite/pull/4865


doodlewind (approved):
This looks intuitive, thanks.

PR #5069, opened 2023-10-20
feat(edgeless): group

### LINKED ISSUE #4999: Support group in edgeless
Design file：[figma link](https://www.figma.com/file/rfH3P9Wbz4IEp2Ai18O8dV/%E2%96%B6%EF%B8%8F-Affine.2023.Late?type=design&node-id=11396-21225&mode=design)

regischen:
close #4999 
### notes
1. group is designed as surface-element, and have `children` of Y.Map field.
2. `surfaceBlock` add private `groupMap` property for tracking child to it's group. And provide `getGroup` and `setGroup` method.
3. SelectionManager now has a `_activeGroup ` property, which is to track current activated group for interaction
4. `groupRootId` constant, virtual root group id. 

Because Y.Map doesn't support Set value, so here using array
just make a update size test
100 shape elements
| children type  | group patch size | release one patch size | total size|
|:----------|:----------|:-------------|:---------|
| Y.Map |2338| 11 |  49715|
| Array | 1340| 1248| 48720

50 shape elements
| children type  | group patch size | release one patch size |
|:----------|:----------|:-------------|
| Y.Map |1238| 11 |
| Array |740|618|

10 shape Elements
| children type  | group patch size | release one patch size |
|:----------|:----------|:-------------|
| Y.Map |358| 11 |
| Array |259|138|

2 shape Elements
| children type  | group patch size | release one patch size |
|:----------|:----------|:-------------|
| Y.Map |181| 11 |
| Array |164|42|

we can see that when group size is becoming larger, the Group operation size of Y.Map is becoming larger. But it's updating patch size is constantly small. 
However the updating patch size of Array is also becoming larger.
Besides if type is array, the searching would also be inefficient.
So maybe Y.Map is good choice?
@doodlewind


doodlewind (commented):
> All EdgelessElement all have `group` property which is updated in runtime. for searching from child to parent.

This should not be persisted into CRDT.

regischen:
> > All EdgelessElement all have `group` property which is updated in runtime. for searching from child to parent.
> 
> This should not be persisted into CRDT.

of course

packages/blocks/src/page-block/edgeless/components/component-toolbar/release-from-group-button.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  protected override render() {
+    return html`<edgeless-tool-icon-button
+      .iconContainerPadding=${2}
+      @click=${() => {
+        this.surface.group.releaseFromGroup(
+          this.surface.edgeless.selectionManager.firstElement
+        );
+      }}
+      .tooltip=${'Release From Group'}
```
doodlewind, line 23: Note the toolbar button font size:

image

packages/blocks/src/image-block/image-model.ts
```diff
@@ -2,6 +2,7 @@ import { BaseBlockModel, defineBlockSchema } from '@blocksuite/store';
 
 import { BLOCK_BATCH } from '../surface-block/batch.js';
 import type { EdgelessBlockType } from '../surface-block/edgeless-types.js';
+import { groupRootId } from '../surface-block/elements/group/contants.js';
```
doodlewind, line 5: I would argue that the block implementation should not depend on group related modules.

packages/blocks/src/note-block/note-model.ts
```diff
@@ -70,6 +71,15 @@ export class NoteBlockModel
   implements IEdgelessElement
 {
   override flavour!: EdgelessBlockType.NOTE;
+  gridBound!: Bound;
```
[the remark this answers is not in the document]
regischen: okay, that's fair

packages/blocks/src/image-block/image-model.ts
```diff
@@ -50,8 +50,10 @@ export class ImageBlockModel
   extends BaseBlockModel<ImageBlockProps>
   implements IEdgelessElement
 {
+  gridBound!: Bound;
```
doodlewind, line 53: I'd suggest that the `gridBound` also should not exist on the model, neither does the `EdgelessSelectableMixin` related methods. But currently since we've landed #4894, removing the debt of these existing methods could be postponed in future refactoring. But at least we should not add more intrusive properties like `gridBound` for the moment.

packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
```diff
@@ -264,8 +264,12 @@ export class EdgelessPageBlockComponent extends BlockElement<
     _disposables.add(
       page.slots.yBlockUpdated.on(({ id, props }) => {
         const block = page.getBlockById(id);
-        if (block && block.flavour === 'affine:note') {
-          if ('prop:xywh' in props) {
+        if (
+          block &&
+          (block.flavour === EdgelessBlockType.NOTE ||
+            block.flavour === EdgelessBlockType.IMAGE)
```
doodlewind, line 270: Such kind of nesting `if` should be simplified.

tests/utils/constants.ts
```diff
@@ -0,0 +1 @@
+export const groupRootId = 'group-root';
```
doodlewind, line 1: This is unnecessary, and please use `consts.ts` for naming instead. Do you really care about being aligned with other team mates' work?

packages/blocks/src/page-block/edgeless/components/component-toolbar/component-toolbar.ts
```diff
@@ -234,18 +264,32 @@ export class EdgelessComponentToolbar extends WithDisposable(LitElement) {
...
         html`<component-toolbar-menu-divider></component-toolbar-menu-divider>`
       );
+      buttons.unshift(this._getCreateGroupButton());
+      buttons.unshift(
+        html`<component-toolbar-menu-divider></component-toolbar-menu-divider>`
+      );
       buttons.unshift(this._getCreateFrameButton());
     }
+    if (elements.length === 1) {
+      if (this.surface.getGroup(selection.firstElement) != groupRootId) {
```
doodlewind, line 285: Don't use `!=` or `==`, always use `!==` and `===`.

packages/blocks/src/page-block/edgeless/components/text/edgeless-group-title-editor.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  override firstUpdated(): void {
+    const dispatcher = this.edgeless.dispatcher;
+    const { surface } = this.edgeless;
+    assertExists(dispatcher);
+
+    this.updateComplete.then(() => {
+      this.vEditor.selectAll();
+
+      surface.updateElementLocalRecord(this.group.id, { titleHide: true });
```
doodlewind, line 46: Use `showTitle` should be more straightforward. Avoid negative naming (and default field values) should be better, since it's better to use `if (show)` than `if (!hide)`, especially in complex condition nesting.

Also, the field naming grammar should be `adj. + n.`, instead of `n. v.`, which is semantically incorrect.

packages/blocks/src/surface-block/elements/group/group-element.ts
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  override get gridBound() {
+    const bound = Bound.deserialize(this.xywh);
+    bound.y -= this._titleHeight;
+    bound.h += this._titleHeight;
+    return bound;
+  }
+
+  addChild(id: string) {
```
doodlewind, line 108: Why put the method on the group element instance side, instead of adding method in `GroupManager`?

packages/blocks/src/surface-block/elements/group/group-element.ts
```diff
@@ -0,0 +1,217 @@
...
+import {
+  getFontString,
+  getLineHeight,
+  getLineWidth,
+  truncateTextByWidth,
+} from '../text/utils.js';
+import type { IGroup, IGroupLocalRecord } from './types.js';
+
+export class GroupElement extends SurfaceElement<IGroup, IGroupLocalRecord> {
+  private _cacheChildren: string[] = [];
```
doodlewind: Should use `cachedChildren` for naming. Beware that "cached" is adjective but "cache" is a verb. Don't be too careless about basic grammar.

packages/blocks/src/surface-block/elements/group/group-element.ts
```diff
@@ -0,0 +1,217 @@
...
+  private _cacheId: string = '';
```
doodlewind: When will the cached id be different with the `this.id`?

packages/blocks/src/surface-block/managers/group-manager.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    const parent = this.surface.pickById(
+      this.surface.getGroup(group)
+    ) as GroupElement;
+    if (parent) {
+      parent.addChild(element.id);
+    }
+  }
+
+  unGroup(group: GroupElement) {
```
doodlewind, line 108: Don't leave all the typos to me.

packages/blocks/src/surface-block/elements/group/group-element.ts
regischen: it's used when the YMap is deleted from the doc. Because at that time, we can't retreive id property from YMap

packages/blocks/src/surface-block/elements/group/group-element.ts
regischen: 1. the interface is simple.
2. groupElement updated its internal cache state.

packages/blocks/src/image-block/image-model.ts
regischen: okay

packages/blocks/src/note-block/note-model.ts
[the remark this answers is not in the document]
regischen: will refactor these property in following iteration,

doodlewind (approved):

PR #5073, opened 2023-10-20
fix(edgeless): frame to image

doouding:


packages/blocks/src/page-block/edgeless/frame-manager.ts
```diff
@@ -127,3 +125,24 @@ export class EdgelessFrameManager {
...
+
+export function getBlocksInFrame(
+  page: Page,
+  model: FrameBlockModel,
+  fullyContained: boolean = true
+) {
+  const bound = Bound.deserialize(model.xywh);
+
+  return (
+    page.getBlockByFlavour([
```
regischen, line 137: how about using `surface.getblocks(EdgelessBlockType)`

packages/blocks/src/__internal__/content-parser/index.ts
```diff
@@ -254,6 +257,35 @@ export class ContentParser {
         blockBound.w,
         blockBound.h
       );
+
+      if (block.flavour === 'affine:frame') {
```
regischen, line 261: how about not hard code? using `EdgelessBlockType.FRAME`

packages/blocks/src/page-block/edgeless/frame-manager.ts
doouding: I try not to rely on the surface cause' it does not exist when rendering edgeless content in doc mode.

doodlewind (approved):

PR #5104, opened 2023-10-24
fix: unselect database cell when click blank area

### LINKED ISSUE #5070: Can't deselect database cell in Edgeless Mode(it also breaks copying outside of the database on edgeless)​
### What happened?

Here is the gif:

https://jmp.sh/S6rqZrDh

After selecting a cell on datatable, I'm not able to copy-paste anymore with keyboard (able to copy with right click and paste )

The issues fixes if you exit from edgeless mode and re-enter

### Distribution version

Windows x64

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

donteatfriedrice:
To close #5070 

https://github.com/toeverything/blocksuite/assets/99816898/f73fd9e8-a9d5-4278-ac61-631889c8a716

### DIFF
--- a/packages/blocks/src/page-block/edgeless/services/selection-manager.ts
+++ b/packages/blocks/src/page-block/edgeless/services/selection-manager.ts
@@ -206,7 +206,10 @@ export class EdgelessSelectionManager {
     this._selection.setGroup('edgeless', [this.state, instance]);
   }
 
-  clear() {
+  clear(force = false) {
+    // if force is true, clear all selections, including all groups and types
+    if (force) this._selection.clear();
+
     this.setSelection({
       blockId: '',
       elements: [],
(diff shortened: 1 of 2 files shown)


packages/blocks/src/page-block/edgeless/services/selection-manager.ts
```diff
@@ -206,7 +206,10 @@ export class EdgelessSelectionManager {
     this._selection.setGroup('edgeless', [this.state, instance]);
   }
 
-  clear() {
+  clear(force = false) {
```
doodlewind, line 209: Instead of adding new option, can we switch the default behavior to `force: true`?

donteatfriedrice: @doouding @Saul-Mirone Can we directly always clear all the selections in edgeless mode when call selection manager clear function?

doodlewind (approved):

Issue #5068, opened 2023-10-20
Unexpected offset for shapes added by dragging


hwangdev97:
![shot-2023-10-20-09 52 47](https://github.com/toeverything/blocksuite/assets/24713927/26b39fd8-fead-4e21-9e3a-6d1eb1eb4949)

Using AFFiNE canary `0.10.2`,  macOS

hwangdev97:
I think @lzlme  can 👀

PR #5106, opened 2023-10-24
fix(edgeless): position of dragged shape

lzlme:
fix: #5068 

https://github.com/toeverything/blocksuite/assets/117659326/358f8199-4385-4119-bab5-dfea20a5d15d


doodlewind (approved):
Thanks, good catch!

PR #5109, opened 2023-10-25
feat(edgeless): support custom note background rgba color for adaptive theme

spacewalkingninja:
Using this one liner we can use any color with `rgba(*,*,*,0.5)` where 0.5 is an example transparency, thus we can easily render custom colors outside the default palette that are enjoyable both on the dark theme and light theme

### DIFF
--- a/packages/blocks/src/page-block/edgeless/components/block-portal/note/edgeless-note.ts
+++ b/packages/blocks/src/page-block/edgeless/components/block-portal/note/edgeless-note.ts
@@ -120,7 +120,7 @@ export class EdgelessBlockPortalNote extends WithDisposable(LitElement) {
       boxSizing: 'border-box',
       background: isHiddenNote
         ? 'transparent'
-        : `var(${background ?? DEFAULT_NOTE_COLOR})`,
+        : ( background.startsWith("rgba")  ? background : `var(${background ?? DEFAULT_NOTE_COLOR})`),
       boxShadow: isHiddenNote ? undefined : 'var(--affine-shadow-3)',
       pointerEvents: 'all',
       overflow: 'hidden',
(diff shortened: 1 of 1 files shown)


spacewalkingninja:
I am currently using it with the following additional function:
`// @ts-ignore
String.prototype.toHSL = function(opts) {
  // @ts-ignore
  var h, s, l;
  opts = opts || {};
  opts.hue = opts.hue || [0, 360];
  opts.sat = opts.sat || [75, 100];
  opts.lit = opts.lit || [40, 60];

  // @ts-ignore
  var range = function(hash, min, max) {
      var diff = max - min;
      var x = ((hash % diff) + diff) % diff;
      return x + min;
  }

  var hash = 0;
  if (this.length === 0) return hash;
  for (var i = 0; i < this.length; i++) {
      hash = this.charCodeAt(i) + ((hash << 5) - hash);
      hash = hash & hash;
  }

  h = range(hash, opts.hue[0], opts.hue[1]);
  s = range(hash, opts.sat[0], opts.sat[1]);
  l = range(hash, opts.lit[0], opts.lit[1]);

  s /= 100;
  l /= 100;
  // @ts-ignore
  const k = n => (n + h / 30) % 12;
  // @ts-ignore
  const a = s * Math.min(l, 1 - l);
  // @ts-ignore
  const f = n => // @ts-ignore
    l - a * Math.max(-1, Math.min(k(n) - 3, Math.min(9 - k(n), 1)));
  return 'rgba('+ (255 * f(0)) +','+ (255 * f(8)) + ','+ (255 * f(4)) + ', 0.5)';
`

It converts any text to a repeatable color. 
Here is how I add notes:
`
var bgcolor = cTask.task_type.toHSL({hue: [180, 360],sat: [75, 95],lit: [55, 65]})
...
const noteId = window.page.addBlock('affine:note', { xywh:dxywh, background:bgcolor }, pageId, );
`

doodlewind (commented):
Interesting work! Could you edit the `CLA.md` in our repo so that we could process the review process? Thanks!

doodlewind:
Still I'd close this since the request for CLA was not responded yet, and this one-liner requires more default color setup modification to be available for users. Thanks for the effort anyway!

PR #5110, opened 2023-10-25
fix(edgeless): frame container zindex

### LINKED ISSUE #5086: Frame layers are messed up
https://github.com/toeverything/blocksuite/assets/144991896/d29d43c7-694e-46f8-975f-fb45b84bc653

Image

regischen:
close #5086

### DIFF
--- a/packages/blocks/src/page-block/edgeless/components/block-portal/frame/edgeless-frame.ts
+++ b/packages/blocks/src/page-block/edgeless/components/block-portal/frame/edgeless-frame.ts
@@ -5,8 +5,11 @@ import { repeat } from 'lit/directives/repeat.js';
 import { styleMap } from 'lit/directives/style-map.js';
 
 import type { FrameBlockModel } from '../../../../../frame-block/index.js';
-import { Bound, compare } from '../../../../../surface-block/index.js';
+import { EdgelessBlockType } from '../../../../../surface-block/edgeless-types.js';
+import { Bound } from '../../../../../surface-block/index.js';
 import type { SurfaceBlockComponent } from '../../../../../surface-block/surface-block.js';
+
+const { FRAME } = EdgelessBlockType;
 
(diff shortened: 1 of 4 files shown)


packages/blocks/src/page-block/edgeless/components/block-portal/frame/edgeless-frame.ts
```diff
@@ -54,6 +54,12 @@ export class EdgelessFramesContainer extends WithDisposable(LitElement) {
   @property({ attribute: false })
   surface!: SurfaceBlockComponent;
 
+  override connectedCallback(): void {
+    super.connectedCallback();
+    this.style.position = 'absolute';
```
doodlewind, line 59: Can this be declared as static CSS?

regischen: okay, let me try whether removing `createRenderRoot` will bring impact.
https://github.com/toeverything/blocksuite/blob/6ba0419a97e5977093cf22e985aa47a332de6047/packages/blocks/src/page-block/edgeless/components/block-portal/frame/edgeless-frame.ts#L72C1-L74C4

regischen: because this
https://github.com/toeverything/blocksuite/blob/6ba0419a97e5977093cf22e985aa47a332de6047/packages/blocks/src/frame-block/frame-block.ts#L39-L41
in order to access parentBlockElement, we can't put blockelement in shadow-root. so static styles will not work.

doodlewind: In this case, why not use `ShadowlessElement`?

doodlewind (approved):

PR #5123, opened 2023-10-25
fix(edgeless): edgeless toolbar submenu tooltip position

donteatfriedrice:
Edgeless toolbar submenu tooltips are no longer affected by overflow since affine-tooltip is a portal currently with pr #4950.
So we can set all the submenu tooltip position as  default ('top').


doodlewind (approved):
Good catch, thanks!

Issue #5101, opened 2023-10-24
In the new clipboard mode, the copy list block does not work properly


zqran:
How to reproduce?

https://github.com/toeverything/blocksuite/assets/15389209/8e64fdbf-99a2-4203-b524-2846d8d328af

PR #5124, opened 2023-10-25
fix: new clipboard will auto merge type for empty paragraph

Saul-Mirone:
Closes: #5101


zqran:
After pasting, error occurs when new element gets focus:


https://github.com/toeverything/blocksuite/assets/15389209/d09a9da6-71fc-458c-80b9-5eaa6858d737

zqran (approved):

Issue #5112, opened 2023-10-25
Label flickering when switching shape style


zanwei:
there's a label appears when switch the shape style

https://github.com/toeverything/blocksuite/assets/40333372/62e71faa-e4ed-4123-98c5-9a2c36ab3090

PR #5125, opened 2023-10-25
fix(edgeless): label flickering when switching shape style

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/5112

When the `edgeless-tool-icon-button` is clicked, the container of the edgeless button is hidden by the following styles. However, it is still present in the DOM, which causes the tooltip not to hide correctly.

https://github.com/toeverything/blocksuite/blob/6ba0419a97e5977093cf22e985aa47a332de6047/packages/blocks/src/page-block/edgeless/components/component-toolbar/change-shape-button.ts#L210-L226

### DIFF
--- a/packages/blocks/src/_common/components/portal.ts
+++ b/packages/blocks/src/_common/components/portal.ts
@@ -203,6 +203,27 @@ export type AdvancedPortalOptions = Omit<
 };
 
 /**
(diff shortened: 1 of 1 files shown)


doodlewind (approved):
LGTM

PR #5126, opened 2023-10-26
feat(page): support ctrl-y shortcut for redo

### LINKED ISSUE #5120: Support Ctrl + Y as a shortcut key for Redo in Windows
### Describe the bug

The Ctrl+Y shortcut is not working for undo actions in desktop client

### To Reproduce

1.  Write some text and use the Ctrl+Z shortcut
2.  Then try using Ctrl+Y shortcut

### Platform
Windows 10

### Distribution version

Windows x64

_No response_

### Relevant log output

_No response_

### Anything else?

Also the zoom-in shortcut is not working on the desktop client Ctrl+ +

lzlme:
resolve #5120

### DIFF
--- a/packages/blocks/src/page-block/keyboard/keyboard-manager.ts
+++ b/packages/blocks/src/page-block/keyboard/keyboard-manager.ts
@@ -9,6 +9,7 @@ export class PageKeyboardManager {
       {
         'Mod-z': ctx => {
           ctx.get('defaultState').event.preventDefault();
+
           if (this._page.canUndo) {
             this._page.undo();
           }
@@ -20,11 +21,11 @@ export class PageKeyboardManager {
           }
         },
         'Control-y': ctx => {
-          if (IS_WINDOWS) {
-            ctx.get('defaultState').event.preventDefault();
-            if (this._page.canRedo) {
-              this._page.redo();
-            }
+          if (!IS_WINDOWS) return;
+
+          ctx.get('defaultState').event.preventDefault();
+          if (this._page.canRedo) {
+            this._page.redo();
           }
         },
         'Mod-Backspace': () => true,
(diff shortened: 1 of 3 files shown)


tests/utils/actions/keyboard.ts
```diff
@@ -97,7 +97,11 @@ export async function formatType(page: Page) {
   });
 }
 
-export async function redoByKeyboard(page: Page) {
+export async function redoByKeyboard(page: Page, onlyWin = false) {
+  if (onlyWin && IS_WINDOWS) {
```
doodlewind, line 101: This is not a valid test, `IS_WINDOWS` should be always false on CI. Given the situation that cross-OS testing suite is not set up, not testing this PR is just fine.

packages/blocks/src/page-block/keyboard/keyboard-manager.ts
```diff
@@ -18,6 +18,12 @@ export class PageKeyboardManager {
             this._page.redo();
           }
         },
+        'Control-y': ctx => {
```
doodlewind, line 21: Should make this to be triggered windows only.

packages/blocks/src/page-block/keyboard/keyboard-manager.ts
```diff
@@ -18,6 +19,14 @@ export class PageKeyboardManager {
...
+          if (IS_WINDOWS) {
```
doodlewind, line 23: We prefer the `if (!IS_WINDOWS) return` guard pattern, this lowers the code nesting and makes condition more straightforward to understand.

lzlme: I just thought about whether to write it like this haha, I will change it now

doodlewind (approved):

PR #5129, opened 2023-10-26
chore(std): increase the attachment size limit

Brooooooklyn:


doodlewind (approved):
Thanks for the 100x optimization! 😈

PR #5130, opened 2023-10-26
fix(edgeless): limit initial zoom ratio

### LINKED ISSUE #5107: Limit the initial zoom ratio on edgeless
image

Currently when only few content exists in the page (e.g., only a piece of canvas text), opening the edgeless editor will set the content as large as possible (600% as the maximum). But the intuitive way should be centering the content but limit the zoom ratio to 100%.

On the other hand, if too much content exists in the page, the initial zoom ratio could be less than 100%, which is also intuitive (and has been implemented for now).

/cc @donteatfriedrice

donteatfriedrice:
To close #5107 

https://github.com/toeverything/blocksuite/assets/99816898/69c67cd3-e224-4e0e-a3d9-f31b803d9c80


regischen (approved):
LGTM

doodlewind (approved):
Thanks for the improvement!

PR #5131, opened 2023-10-26
fix(page): `getPreviousBlock` not works as expected when there are several notes

### LINKED ISSUE #5100: Delete at the start of second note will jump to title
https://github.com/toeverything/blocksuite/assets/50035259/e0674a6d-dcd3-43e2-938c-ed12190413dc

Flrande:
Close #5100

### DIFF
--- a/packages/blocks/src/_common/utils/query.ts
+++ b/packages/blocks/src/_common/utils/query.ts
@@ -90,46 +90,48 @@ export function getNextBlock(
  *  - paragraph <- when invoked here, the traverse order will be above
  * ```
  *
- * NOTE: this method will skip the `affine:note` and `affine:page` block
+ * NOTE: this method will just return blocks with `content` role
  */
-export function getPreviousBlock(
-  model: BaseBlockModel,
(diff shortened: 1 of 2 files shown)


packages/blocks/src/_common/utils/query.ts
```diff
@@ -90,46 +90,48 @@ export function getNextBlock(
...
-  model: BaseBlockModel,
-  map: Record<string, true> = {}
-): BaseBlockModel | null {
-  if (model.id in map) {
-    throw new Error(
-      "Can't get previous block! There's a loop in the block tree!"
-    );
-  }
-  map[model.id] = true;
+export function getPreviousBlock(model: BaseBlockModel): BaseBlockModel | null {
```
Saul-Mirone, line 95: I think we should use the `getNextBlockCommand` instead of this one.

Flrande: This relates to the refactoring of `rich-text-operations.ts`.

doodlewind (approved):

PR #5140, opened 2023-10-27
feat(store): add config manager in workspace

Flrande:
Support dynamic adjustment of file size limits.

```ts
// set max size to 100MB
workspace.config.set('attachmentMaxFileSize', 100 * 1000 * 1000)
```


Saul-Mirone (changes requested):
I think config should not be a workspace level property. I'll finish the context module soon and config should be set on context.

doodlewind:
Closing as waiting for upstream infra implementation.

PR #5141, opened 2023-10-27
fix(edgeless): keep component toolbar in viewport

donteatfriedrice:
https://github.com/toeverything/blocksuite/assets/99816898/5f800689-a67a-4f37-89e5-d04c515c08a8


doodlewind (approved):
Thanks for keep shipping detailed improvements!

PR #5158, opened 2023-10-31
feat: support config attachment max file size

Flrande:
Pass a custom service through the preset to add logic that can modify the maximum upload file size of attachments.

```ts
import {
  AttachmentService,
  EdgelessPreset,
  PagePreset,
} from '@blocksuite/blocks';

const params = new URLSearchParams(location.search);

class CustomAttachmentService extends AttachmentService {
  override mounted(): void {
    const userType = params.get('userType');
    if (userType === 'pro') {
      this.maxFileSize = 100 * 1000 * 1000; // 100MB
    } else {
      this.maxFileSize = 10 * 1000 * 1000; // 10MB
    }
  }
}

const getPlaygroundPresets = () => {
  const pageModePreset = PagePreset.map(preset => {
    if (preset.schema.model.flavour === 'affine:attachment') {
      return {
        ...preset,
        service: CustomAttachmentService,
      };
    }
    return preset;
  });
  const edgelessModePreset = EdgelessPreset.map(preset => {
    if (preset.schema.model.flavour === 'affine:attachment') {
      return {
        ...preset,
        service: CustomAttachmentService,
      };
    }
    return preset;
  });

  return {
    pageModePreset,
    edgelessModePreset,
  };
};

const presets = getPlaygroundPresets();
const editor = new EditorContainer();
editor.pagePreset = presets.pageModePreset;
editor.edgelessPreset = presets.edgelessModePreset;
editor.page = page;
document.append(editor);
```

https://github.com/toeverything/blocksuite/assets/50035259/70d7e3a1-57d4-401f-b4b7-868c17e3c02d


packages/playground/apps/default/preset.ts
```diff
@@ -0,0 +1,44 @@
...
+    const userType = params.get('userType');
+    if (userType === 'pro') {
+      this.maxFileSize = 100 * 1000 * 1000; // 100MB
+    } else {
+      this.maxFileSize = 10 * 1000 * 1000; // 10MB
+    }
+  }
+}
+
+export const getPlaygroundPresets = () => {
```
doodlewind, line 20: Just saying that using `export function getPlaygroundPresets() {}` should be more preferable in our codebase.

doodlewind (approved):

Ygs0x00:
May I ask what the problem is that modifying the attachment size through this method is invalid and keeps reporting an error of 10mb

Flrande:
> May I ask what the problem is that modifying the attachment size through this method is invalid and keeps reporting an error of 10mb

How do you upload files? A screen record will be helpful.

Ygs0x00:
> > May I ask what the problem is that modifying the attachment size through this method is invalid and keeps reporting an error of 10mb
> 
> How do you upload files? A screen record will be helpful.

This is my screen record.
https://github.com/toeverything/blocksuite/assets/18319599/2ca1e5fc-3caa-40fc-a955-30b90f11173a

lawvs:
> May I ask what the problem is that modifying the attachment size through this method is invalid and keeps reporting an error of 10mb

You should override the default configuration in your custom service because the `mount` method will be called after the attachment block is inserted.

Here is an example:

```ts
export class CustomAttachmentService extends AttachmentService {
  maxFileSize = bytes.parse('100MB');

  // ...
}

```

Ygs0x00:
> > May I ask what the problem is that modifying the attachment size through this method is invalid and keeps reporting an error of 10mb
> 
> Could you try to override the default configuration in your custom service?
> 
> Here is an example:
> 
> ```ts
> export class CustomAttachmentService extends AttachmentService {
>   override maxFileSize = bytes.parse('100MB');
> 
>   // ...
> }
> ```

After trying, the situation remained the same.

Flrande:
> After trying, the situation remained the same.

It is because `affine-editor-container` not load specs from property to `doc-editor` and `edgeless-editor`. I will fix it.

image

Issue #4978, opened 2023-10-10
Connector UX update


Svaney-ssman:
[TOV-201](https://linear.app/affine-design/issue/TOV-201/connector-optimization)

[Here is the design ](https://www.figma.com/file/rfH3P9Wbz4IEp2Ai18O8dV/%E2%96%B6%EF%B8%8F-Affine.2023.Late?type=design&node-id=11018-34521&mode=design&t=Rg768nPCeFlpYPwn-4)

Changes:
1. The connector adjust indicator was redesigned

![Image](https://github.com/toeverything/blocksuite/assets/110808979/000321a5-b202-41a2-8aef-abbcfddd046a)

2. Connector with Text added
3. A new way of adding a connector

![Image](https://github.com/toeverything/blocksuite/assets/110808979/44b5ffd6-7be1-4d56-97a0-ba613fcbd871)

hwangdev97:
![shot-2023-10-24-17 37 26@2x](https://github.com/toeverything/blocksuite/assets/24713927/c21d91a4-b778-4631-9d1c-5199483bbf5c)

Update some details: Please place the top of the connect on the border. When there is no border, gaps can be left.

PR #5161, opened 2023-10-31
feat(edgeless): new way to add connectors

donteatfriedrice:
Part of #4978 : a new way of adding a connector

- [x] Drag on shape auto-complete button can add connector and select connector target type, including 5 shape types, canvas text, note, frame
- [x] Add placeholder for the different target types
- [x] Make frame connectable


packages/blocks/src/_common/utils/types.ts
```diff
@@ -137,9 +137,10 @@ export type Selectable = EdgelessElement;
 
 export type Erasable = EdgelessElement;
 
-export type Connectable =
-  | Exclude<TopLevelBlockModel, FrameBlockModel>
-  | Exclude<PhasorElement, ConnectorElement | BrushElement>;
+export type Connectable = Exclude<
```
regischen: This maybe includes `TopLevelBlockModel`?

regischen (approved):

packages/blocks/src/_common/utils/types.ts
donteatfriedrice: I will add it back. Thanks!

doodlewind (approved):
This generally looks good, thanks for the feat!

Issue #1290, opened 2023-02-20
Recognize the wrap boundary inside multi-line text


doodlewind:
Due to the limitation of `contenteditable`, inside a `<p>` tag, it's difficult to distinguish between the last position of first line and the start of next line. This could affects the edge case of ArrowUp support when cursor locates at the wrapped position of the line.

One approach is to use `contenteditabe` in page container level (#1272) but this may be overkill.

doodlewind:
This affects #2805 #2834 (and more cursor related issues). And this remains after #3734 since that's still how we manage cross `contenteditable` areas. But we are open to new options.

Flrande:
We don't need to make a distinction since we can't place the cursor at the end of the first line without causing a line break.

https://github.com/toeverything/blocksuite/assets/50035259/fd946529-6a87-4764-8a1e-9d6648f91a7c

I've reviewed issues #2834 and #4435, and I believe there are potential solutions for them. I'll allocate some time to work on it.

doodlewind:
@Flrande I would suggest making this pending, since we have the plan migrating to one single contenteditable.

PR #5166, opened 2023-11-01
fix(page): undo will be triggered outside of the editor

Flrande:
The root cause of this bug is that KeyboardControl will listen for keydown events on the document. After we switch to global contenteditable, we can avoid this workaround.


doodlewind (approved):

Flrande:
This bug is related to event target and the internal mechanisms of event dispatcher, and fixing it might require modifications to a lot of existing code. After discussing with @Saul-Mirone we suggest addressing this issue when moving to global contenteditable. cc @doodlewind @joooye34

doodlewind:
@Flrande makes sense, closing this and wait for the refactoring for #1290

PR #5196, opened 2023-11-03
feat(store): add gc in blob manager

### LINKED ISSUE #5170: Deleting images will never trigger `storage.crud.delete`
## Step to reproduce

open the playground and add a custom blob storage in console by:

```js
const map = new Map();

// how we know this script works
// if you delete the `get` method in crud, and paste a picture in the doc
// the pic will be shown `broken`.
workspace._storages.push({
  crud: {
    get: (k) => Promise.resolve(map.get(k)),
    set: (k, v) => { map.set(k, v); return Promise.resolve() },
    delete: (k) => { console.log('delete') },
  }
})
```

and whenever you paste/delete a picture, you will only see get/set methods get called, and no 'delete' printed in the console.

https://github.com/toeverything/blocksuite/assets/8281226/573e1213-8e85-49ce-8330-1067274291e7

Flrande:
Close #5170

### DIFF
--- a/packages/store/src/workspace/addon/blob.ts
+++ b/packages/store/src/workspace/addon/blob.ts
@@ -87,9 +87,16 @@ export const blob = addOnFactory<keyof BlobAddon>(
           },
         };
 
-        window.addEventListener('beforeunload', () => {
-          this.blob.gc();
-        });
+        if (typeof window !== 'undefined') {
+          window.addEventListener('beforeunload', () => {
+            this.blob.gc();
+          });
+        }
+        if (typeof process !== 'undefined') {
+          process.on('exit', () => {
+            this.blob.gc();
+          });
+        }
       }
     }
 );
--- a/packages/store/src/workspace/addon/blob.ts
+++ b/packages/store/src/workspace/addon/blob.ts
@@ -83,11 +83,13 @@ export const blob = addOnFactory<keyof BlobAddon>(
           },
           decreaseRef: async blobId => {
             const ref = this._blobsRef.get(blobId) ?? 0;
-            this._blobsRef.set(blobId, ref - 1 < 0 ? 0 : ref - 1);
-
-            this.blob.gc();
+            this._blobsRef.set(blobId, Math.max(ref - 1, 0));
           },
         };
+
+        window.addEventListener('beforeunload', () => {
+          this.blob.gc();
+        });
       }
     }
 );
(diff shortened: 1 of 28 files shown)


packages/store/src/workspace/addon/blob.ts
```diff
@@ -69,6 +67,26 @@ export const blob = addOnFactory<keyof BlobAddon>(
...
+          },
+          increaseRef: async blobId => {
+            const ref = this._blobsRef.get(blobId) ?? 0;
+            this._blobsRef.set(blobId, ref + 1);
+          },
+          decreaseRef: async blobId => {
+            const ref = this._blobsRef.get(blobId) ?? 0;
+            this._blobsRef.set(blobId, ref - 1 < 0 ? 0 : ref - 1);
+
+            this.blob.gc();
```
lawvs, line 88: We should delay deleting the blob, otherwise, it will break the cut or undo/redo operation.

https://github.com/toeverything/blocksuite/assets/18554747/bf77f623-9368-4d30-a5c3-c58ce403c6b2

lawvs (changes requested):

packages/store/src/workspace/addon/blob.ts
```diff
@@ -69,6 +67,26 @@ export const blob = addOnFactory<keyof BlobAddon>(
...
+              }
+            });
+          },
+          increaseRef: async blobId => {
+            const ref = this._blobsRef.get(blobId) ?? 0;
+            this._blobsRef.set(blobId, ref + 1);
+          },
+          decreaseRef: async blobId => {
+            const ref = this._blobsRef.get(blobId) ?? 0;
+            this._blobsRef.set(blobId, ref - 1 < 0 ? 0 : ref - 1);
```
lawvs, line 86: A little trick

```suggestion
            this._blobsRef.set(blobId, Math.max(ref - 1, 0));
```

packages/store/src/workspace/page.ts
```diff
@@ -510,7 +513,12 @@ export class Page extends Space<FlatBlockMap> {
       syncBlockProps(schema, yBlock, props);
     });
 
-    model.propsUpdated.emit();
+    const newProps = toBlockProps(yBlock, this.doc.proxy);
+
+    model.propsUpdated.emit({
+      oldProps: oldProps,
+      newProps: newProps,
```
lawvs, line 520:
```suggestion
      oldProps,
      newProps,
```

lawvs (approved):
LGTM, this solution can handle most scenarios.

However, I still think that relying solely on reference counting is not reliable enough. For example, it may incorrectly count the blobs of old versions, which could result in potential data loss.

I would recommend considering the addition of a new resource manager based on an indexer in the future.

doodlewind (approved):

PR #5200, opened 2023-11-06
feat(edgeless): support group indexing

### LINKED ISSUE #5127: Support adjusting layer sequence of edgeless group
Part of #5009 

Because group has name, so its better to make group have index too.

# Solution 1:
group itself have index too, and the children's index is relative to its group.

## illustrate

Like: if we have following order, D on the top
A --- B --- C  ---- D

Group A and C:

group --- B ----- D
|        |
A     C
we just assign group a index. that's enough. the relative order of A and C is not changed.

Adjust A's index is to compare to it's same level elements like C. The principle applies to group.

## ungroup
group --- B----D
|       |
A      C

recalculate the index of A and C, ensure the index is less then B.

## release from group
group ---- B------D
|     |
A    C

release A from group.
make sure A is between group and B

## Render order
group-1 -------  E ------- F
|       |      |
group2 ---- A -----    B    
|    |
C  D

it's Post-order Traversal。
C , D , group2 , A , B，Group1, E, F

# Solution 2:
Group also have index. But the index of the children of group is not dependent to the index of group.
All indexes are at the same level. For example:
group1 has children of C and D.
the index order is:
A B group1 C D E  F

## Group
A --- B--- C ---D 
after grouping A and C, we get:
B --- D --- group --- A --- C

## Ungroup
B --- D --- group --- A --- C
after ungrouping group, we get:
B --- D --- A ----C

regischen:
close #5127 

# Notes:
1. refactoring old index adjustment logic.
2. extract `getGroupParent`, `setGroupParent` logic from surface to global scope. because now render order is dependent of group. and surface ref doesn't access surface block instance.
3. GroupMap type changed from `Map<id,id>` to `Map<id, GroupElement>`

# Solution :
group itself have index too, and the children's index is relative to its group.

## illustrate

Like: if we have following order, D on the top
```
A --- B --- C  --- D
```
Group A and C:
```
group --- B --- D
|   |
A   C
```
we just assign group a index. that's enough. the relative order of A and C is not changed.

Adjust A's index is to compare to it's same level elements like C. The principle applies to group.

## ungroup
```
group --- B --- D
|   |
A   C
```
recalculate the index of A and C, ensure the index is less then B.
```
A --- C --- B --- D
```

## release from group
```
group --- B ----- D
|   |
A   C
```
release A from group.
make sure A is between group and B
```
group --- A --- B ----- D
   |
   C
```

## Render order
```
---- group-1 -------  E ------- F
    |  | |
group2 A B    
|  |
C  D
```

it's Post-order Traversal。
C , D , group2 , A , B，group1 , E , F


packages/blocks/src/surface-block/managers/group-manager.ts
```diff
@@ -3,10 +3,92 @@ import { Workspace } from '@blocksuite/store';
 
 import type { EdgelessElement } from '../../_common/utils/index.js';
 import { PhasorElementType } from '../elements/edgeless-element.js';
-import { GROUP_ROOT_ID } from '../elements/group/consts.js';
+import { GROUP_ROOT } from '../elements/group/consts.js';
 import { GroupElement } from '../elements/group/group-element.js';
 import type { SurfaceBlockComponent } from '../surface-block.js';
 
+export const GroupMap = new Map<string, GroupElement>(); // { blockId: groupId }
```
doodlewind, line 10: This may not work for multiple editor instances.

doodlewind (approved):
I can see some anti pattern related to how we share code between page block and surface block, but it's best to do the housekeeping in a standalone PR.

Issue #5042, opened 2023-10-17
Support recognizing inline-code on pressing enter


fundon:
https://github.com/toeverything/blocksuite/assets/27926/e766241d-894c-4077-a331-6017577056cb

doodlewind:
How about pressing space at the end of first line? This should trigger markdown recognition.

fundon:
> How about pressing space at the end of first line? This should trigger markdown recognition.

Yes, it works.

PR #5226, opened 2023-11-07
feat(page): enter key also triggers inline style formatting

AyushAgrawal-A2:
Closes: #5042 

https://github.com/toeverything/blocksuite/assets/54364088/86591e48-d302-4135-9b2a-e6dcc7c938ff


doodlewind (approved):
Good catch!

PR #5233, opened 2023-11-07
fix(edgeless): auto connect for ref block

### LINKED ISSUE #5223: Support auto connector for surface-ref block

regischen:
close #5223 

https://github.com/toeverything/blocksuite/assets/58546692/cefa700c-8406-40c5-bb1e-84e474eec320


packages/blocks/src/page-block/edgeless/components/component-toolbar/component-toolbar.ts
```diff
@@ -220,33 +212,83 @@ export class EdgelessComponentToolbar extends WithDisposable(LitElement) {
...
+
+    const left = clamp(x, 10, width - rect.width - 10);
+    top = clamp(top, 10, height - rect.height - 100);
+
+    this.style.left = left + 'px';
+    this.style.top = top + 'px';
+  }
+
   override render() {
+    this._updatePosition();
```
doodlewind, line 291: I think it's anti-pattern to perform this kind of side effect in render function.

packages/blocks/src/page-block/edgeless/components/component-toolbar/component-toolbar.ts
```diff
@@ -254,8 +265,17 @@ export class EdgelessComponentToolbar extends WithDisposable(LitElement) {
     return html`<component-toolbar-menu-divider></component-toolbar-menu-divider>`;
   }
 
-  private async _updatePosition() {
-    // await this.updateComplete;
+  protected override updated(_changedProperties: PropertyValues): void {
+    // if (_changedProperties.has('left') || _changedProperties.has('top')) {
+    const [left, top] = this._updatePosition();
```
doodlewind, line 261: Generally an `updateSomething` API indicates performing side effect, instead of returning values. Using naming like `computePosition` could be better. Also note not having the redundant comment.

regischen: ok, I forgot to change the name

packages/blocks/src/page-block/edgeless/components/auto-connect/edgeless-auto-connect-line.ts
```diff
@@ -106,6 +118,9 @@ export class EdgelessAutoConnectLine extends WithDisposable(LitElement) {
           EXPAND_OFFSET / 2,
         ]);
 
+        if (newStart.every(a => isNaN(a)) || newEnd.every(a => isNaN(a)))
```
doodlewind, line 121: I think we should avoid using `isNaN` that works _after_ the invalid data is generated, but to add testing condition for the root cause and avoid `NaN`. The `isNaN` should not used across the codebase, unless for generic data conversion testing.

packages/blocks/src/page-block/edgeless/components/auto-connect/edgeless-index-label.ts
```diff
@@ -196,30 +206,112 @@ export class EdgelessIndexLabel extends WithDisposable(ShadowlessElement) {
...
         });
-        return html`
-          <div index=${index} class="edgeless-index-label" style=${style}>
-            ${index + 1}
-          </div>
-        `;
+        const components = [];
+        const count = counts[i];
+
+        function calculatePosition(gap: number) {
```
doodlewind, line 235: This `render` function is already very verbose, should extract this as util.

packages/blocks/src/page-block/edgeless/components/auto-connect/edgeless-index-label.ts
```diff
@@ -13,6 +13,39 @@ import type { SurfaceBlockComponent } from '../../../../surface-block/surface-bl
...
+      }
+    } else {
+      const offset = gap + iconWidth;
+      left = 10 + Math.ceil(Math.abs(j - middleIndex)) * offset * getSign(j);
+    }
+    positions.push([0, left]);
+  }
+
+  return positions;
+}
```
doodlewind, line 48: Mind adding blank space with classes :/

doodlewind (approved):

Issue #5014, opened 2023-10-13
The visual display is incorrect when trying to create a link in multiple lines


lawvs:
The `mock-selection` can only display rectangular shapes, which is not precise enough.

<img width="796" alt="Screenshot 2023-10-13 at 14 39 40" src="https://github.com/toeverything/blocksuite/assets/18554747/ed29f3d0-1ea3-4c6f-a38d-fdede33163e1">

https://github.com/toeverything/blocksuite/assets/18554747/42794201-d90e-48d1-b0ab-e9c327151ce6

PR #5241, opened 2023-11-07
fix(page): mock selection of multiline text on create/edit link

AyushAgrawal-A2:
Closes: #5014 

Changes:
* Mulitple mock selection divs, derived using `getClientRects`, for representing text selection accurately.
* Check for `isValidUrl` in the `_onConfirm` method before `normalizeUrl`. Previously, an invalid URL was getting accepted on pressing the Enter key as this check was after normalizeUrl.

https://github.com/toeverything/blocksuite/assets/54364088/2a39ac1d-923e-4f7c-ba00-1b0fc3a49545

### DIFF
--- a/packages/blocks/src/_common/components/rich-text/virgo/nodes/link-node/link-popup/link-popup.ts
+++ b/packages/blocks/src/_common/components/rich-text/virgo/nodes/link-node/link-popup/link-popup.ts
@@ -165,8 +165,10 @@ export class LinkPopup extends WithDisposable(LitElement) {
     if (!this.vEditor.isVRangeValid(this.goalVRange)) return;
 
     assertExists(this.linkInput);
-    const link = normalizeUrl(this.linkInput.value);
-    if (!isValidUrl(link)) return;
+    const linkInputValue = this.linkInput.value;
+    if (!linkInputValue || !isValidUrl(linkInputValue)) return;
+
+    const link = normalizeUrl(linkInputValue);
 
     if (this.type === 'create') {
       this.vEditor.formatText(this.goalVRange, {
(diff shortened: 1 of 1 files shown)


packages/blocks/src/_common/components/rich-text/virgo/nodes/link-node/link-popup/link-popup.ts
```diff
@@ -87,12 +87,19 @@ export class LinkPopup extends WithDisposable(LitElement) {
...
+      Object.values(domRects).forEach(domRect => {
+        const mockSelection = document.createElement('div');
+        mockSelection.classList.add('mock-selection');
+        mockSelection.style.left = `${domRect.left}px`;
+        mockSelection.style.top = `${domRect.top}px`;
+        mockSelection.style.width = `${domRect.width}px`;
+        mockSelection.style.height = `${domRect.height}px`;
+
+        assertExists(this.mockSelectionContainer);
+        this.mockSelectionContainer.appendChild(mockSelection);
```
doodlewind, line 101: Should the container be cleaned up before appending?

AyushAgrawal-A2: The link-popup is a fresh element added to dom every time create / edit link called, with an empty container.
It does not update during its lifecycle. Clearing it seems not nessesary in the current implementation.

AyushAgrawal-A2:
Hi @doodlewind,
While solving this issue i noticed that while creating a new link, if the selection has some text with different text attributes multiple link elements are created :-

https://github.com/toeverything/blocksuite/assets/54364088/ff86a1cb-8759-40f2-8c2e-1953e2aa75ac


Shouldn't it be a single link with all text attributes removed ?

doodlewind:
@AyushAgrawal-A2 thanks for the finding, removing all attributes here makes sense - new PR welcome :)

doodlewind (approved):

PR #5242, opened 2023-11-07
fix: surface-ref interaction && failed test in AFFiNE

doouding:
- No need to slice the note when inserting the `surface-ref` block
- Add "Insert into page" button


doodlewind (approved):
Thanks for the feat!

PR #5243, opened 2023-11-07
feat(page): support embed tweet in bookmark

lawvs:
https://github.com/toeverything/blocksuite/assets/18554747/7e224fad-53e7-4e41-8598-2f7dff95095d

### DIFF
--- a/packages/blocks/src/bookmark-block/embed.ts
+++ b/packages/blocks/src/bookmark-block/embed.ts
@@ -38,6 +38,70 @@ const embedConfig: EmbedConfig[] = [
       return `https://www.youtube.com/embed/${videoId}`;
     },
   },
+  {
+    // See also https://publish.twitter.com/
+    name: 'X(Twitter) ',
+    format: url => {
+      if (!['www.twitter.com', 'twitter.com', 'x.com'].includes(url.hostname))
+        return null;
+      // https://twitter.com/[username]/status/[tweetId]
+      const tweetId = url.pathname.match(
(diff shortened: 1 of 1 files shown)


doodlewind (commented):
This is really interesting! The card width looks unaligned, any change for a quick workaround?

image

lawvs:
> This is really interesting! The card width looks unaligned, any change for a quick workaround?

Based on the Twitter developer platform document, the maximum width of the embedded tweet is 550px, so it's not advisable to modify its width. but now it will align the center correctly.

Screenshot 2023-11-07 at 19 17 50

Besides, since the design file is used for embedding YouTube or Figma, it will crop any content that exceeds 480px when displaying a tweet.


> ## Will embedded Tweets adapt to my responsive site?
> An embedded Tweet will adjust to the width of its containing element when inserted into a page. An embedded Tweet requires a minimum width of 220 pixels and will fill up to 550 horizontal pixels.
> https://developer.twitter.com/en/docs/twitter-for-websites/embedded-tweets/faq#:~:text=An%20embedded%20Tweet%20requires%20a,up%20to%20550%20horizontal%20pixels.

doodlewind (approved):
Thanks for the effort, and it's just cool to land this as dev prototype. But note that since the way extending bookmark is not yet scalable (by having one block flavour to fit all third-party sites), and the UX of bookmark may be subject to change, the process exposing this feature to end users and developers still need to be pending for the product level decision.

PR #5254, opened 2023-11-08
fix: format-bar custom elements

doouding:
Fixed the bug that custom elements will disappear after the `format-bar` rerendering.


doodlewind (approved):
LGTM

Issue #5103, opened 2023-10-24
The database cells should support the operation of cutting ⌘/Ctrl + X


houjoe0829:
- For Link, Number and Text fields, cutting will be handled the same as in a plain text editor: it will delete the selected content and copy it to the clipboard.
- For single-select, multi-select, and date fields, when the entire cell is selected and operated on, it will clear the data in the cell and copy the content to the clipboard.
- In the edit state of single-select or multi-select fields, the behavior should be the same as a regular text input field, supporting cut operation.
- The cut operation is currently not supported for the Progress and Checkbox fields.

zqran:
> The cut operation is currently not supported for the Progress and Checkbox fields.

@houjoe0829 
I know that when performing a cut on the `progress` column, we don't need to clear its data. 
But do I need to copy the progress data to the clipboard?

houjoe0829:
> > The cut operation is currently not supported for the Progress and Checkbox fields.
> 
> @houjoe0829 I know that when performing a cut on the `progress` column, we don't need to clear its data. But do I need to copy the progress data to the clipboard?

Got it, there is no need to handle it temporarily, unless there is specific feedback from users.

houjoe0829:
Add:
For Title Column, cutting will be handled the same as in a plain text editor: it will delete the selected content and copy it to the clipboard.

PR #5255, opened 2023-11-08
feat(database): cells support cut operation

zqran:
close: #5103


doodlewind (approved):
Thanks for the continuous effort!

PR #5272, opened 2023-11-09
feat(edgeless): optimize batch drag update size

### LINKED ISSUE #5190: Optimize edgeless performance bottlenecks
Before landing the perf testing suite in #5095, we already have located several major perf bottlenecks:

* Too much update size in batch dragging.
* Too much update size in shape drawing.

The optimization should be on plan.

doouding:
close #5190 

### Change
- Add a new class to encapsulate the logic related to the local record
- Add a wrapper to prioritize retrieving the property of block from the local record


doouding:
Although all the tests have passed. It would be better to have a period of trial use before landing in AFFiNE.

doodlewind (approved):
Thanks for the optimization!

Issue #5273, opened 2023-11-09
Unexpected shape auto complete indicator appears when dragging


lawvs:
https://github.com/toeverything/blocksuite/assets/18554747/a09d8d29-e04a-4c4a-8851-f25aba61d8cc

PR #5275, opened 2023-11-09
fix(edgeless): remove shape overlay when element disconnect

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/5273


doodlewind (approved):

PR #5278, opened 2023-11-10
fix(std): should not re-execute previous commands in the chain before `try`

### LINKED ISSUE #5271: Command executed multiple times
Sample code to reproduce the problem:

```ts
  public onPageCut: UIEventHandler = ctx => {
    const e = ctx.get('clipboardState').raw;
    e.preventDefault();

    this._copySelected(e)
      .getTextSelection()
      .try(cmd => [cmd.deleteText(), cmd.deleteSelectedModels()])
      .run();
  };
```

Maybe it has something to do with `try()`.

Flrande:
Close #5271

### DIFF
--- a/packages/block-std/package.json
+++ b/packages/block-std/package.json
@@ -6,7 +6,10 @@
   "repository": "toeverything/blocksuite",
   "scripts": {
     "build": "tsc",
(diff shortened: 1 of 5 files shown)


doodlewind (approved):
Feel like being AI-powered!

Issue #5281, opened 2023-11-10
Cursor jumps lines during Arrow Up/Down Navigation


AyushAgrawal-A2:
If a block has multiple span elements, due to having text with different attributes, cursor jumps the entire block on navigating with arrow up/down

https://github.com/toeverything/blocksuite/assets/54364088/9cc49637-a1b2-46ad-9c14-0364e4aa2cd1

Issue #5155, opened 2023-10-30
Incorrect cursor jump behavior occurs when using the keyboard to navigate rich text


lawvs:
https://github.com/toeverything/blocksuite/assets/18554747/206d4782-519b-497e-831c-071d579928bf

PR #5282, opened 2023-11-10
fix(page): arrow up/down navigation

AyushAgrawal-A2:
Closes: #5281 (duplicate of #5155)
Closes: #5155 
Closes: #2834 

Currently, the app only handles inter-block up/down navigation. Within the same block is handled by the browser. After many iterations, it felt to have consistent behavior and handle all edge cases either app or browser should completely handle the navigation.

This may be removed when migration to a single root content editable container is complete and the browser is able to handle navigation completely.

### DIFF
--- a/packages/blocks/src/note-block/keymap.ts
+++ b/packages/blocks/src/note-block/keymap.ts
@@ -15,8 +15,8 @@ import {
 import {
   ensureBlockInContainer,
   getBlockSelectionBySide,
-  moveCursorToNextBlockElement,
-  moveCursorToPrevBlockElement,
+  moveCursorToBlock,
+  moveCursorVertically,
   pathToBlock,
   selectBetween,
   setBlockSelection,
@@ -43,6 +43,7 @@ export function bindHotKey(blockElement: BlockElement) {
   });
   blockElement.bindHotKey({
     ArrowDown: () => {
+      let currentBlock: BlockElement | null = null;
       return root.std.command
         .pipe()
         .inline((_, next) => {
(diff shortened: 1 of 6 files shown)


AyushAgrawal-A2:
Hey @Flrande ,
Some tests are failing because of code changes made in this PR, I am fixing them and I will push more commits here.
Marking this as a draft PR for now. I will re-request a review when complete.

AyushAgrawal-A2:
@Flrande 
Submitting for review.

Flrande:
> @Flrande Submitting for review.

Thank you for your fix! I will review your code a bit later. If you're willing, you could add corresponding tests, so as to prevent others' changes from causing the bug to reappear.

AyushAgrawal-A2:
> > @Flrande Submitting for review.
> 
> Thank you for your fix! I will review your code a bit later. If you're willing, you could add corresponding tests, so as to prevent others' changes from causing the bug to reappear.

sure, will do

AyushAgrawal-A2:
@Flrande
tests added

packages/blocks/src/note-block/keymap.ts
```diff
@@ -41,13 +41,37 @@ export function bindHotKey(blockElement: BlockElement) {
...
         .pipe()
         .inline((_, next) => {
           reset();
           next();
         })
         .try(cmd => [
+          // text selection - case 1: move cursor down within the same block
+          cmd
+            .getTextSelection()
+            .inline<'currentSelectionPath'>((ctx, next) => {
```
Flrande, line 55: There seems to be a repetitive `inline` command in the logic of `try`, and I think we can add it before `try` to avoid repetition.

```ts
root.std.command
        .pipe()
        .inline((_, next) => {
          reset();
          next();
        })
        .inline<'currentSelectionPath'>((ctx, next) => {
              const textSelection = ctx.currentTextSelection;
              assertExists(textSelection);
              const end = textSelection.to ?? textSelection.from;
              next({ currentSelectionPath: end.path });
        })
        .try(...)
```

AyushAgrawal-A2: @Flrande to move this part out of `try` block, I also have to move `getTextSelection` with it as it need `ctx.currentTextSelection`
```ts
root.std.command
        .pipe()
        .inline((_, next) => {
          reset();
          next();
        })
 -->   .getTextSelection()  
        .inline<'currentSelectionPath'>((ctx, next) => {
              const textSelection = ctx.currentTextSelection;
              assertExists(textSelection);
              const end = textSelection.to ?? textSelection.from;
              next({ currentSelectionPath: end.path });
        })
        .try(...)
```

This was failing in case selection is a block and does not reach the last condition of moving arrow for block selection present in the same try block.

I have now switched to a nested try block, where I am able to extract the repetitive code out of the inner try block.

Flrande (approved):

Flrande:
LGTM, thanks again for your contribution!

PR #5287, opened 2023-11-10
fix(edgeless): layout shift when switching editor mode

doouding:
Before:

https://github.com/toeverything/blocksuite/assets/9301743/b756ba0b-3bab-4d14-822f-7e4de1465f78

After:

https://github.com/toeverything/blocksuite/assets/9301743/1aa989f5-2e65-4ae8-abd7-c583535ce329


doodlewind (approved):
Looks good but seems that this would break some tests?

Issue #5291, opened 2023-11-11
Should correct the missing  image address in "package/docs"


water-in-stone:
1. The picture named "unidirectional-data-flow", required in "https://github.com/toeverything/blocksuite/blob/master/packages/docs/selection-api.md", is missing correct address.
2. The picture named "unidirectional data flow", required in https://github.com/toeverything/blocksuite/blob/master/packages/docs/rich-text-orchestration.md, is missing correct address.

doodlewind:
Closing since these are doc indeed.

PR #5292, opened 2023-11-11
docs: add correct image address (#5291)

water-in-stone:


doodlewind (commented):
Sorry, this is actually link to another doc, not images. Since we are preparing for refactoring on selection, this doc is not public in the doc site - but it still shows how current edit engine work.

Issue #5204, opened 2023-11-06
Unexpected placeholder


Flrande:
![image](https://github.com/toeverything/blocksuite/assets/50035259/0378fdba-1fa8-424a-84ae-53a369bb9d90)

Flrande:
We need a stable reproduction path.

PR #5295, opened 2023-11-12
fix(page): unexpected placeholder

zqran:
close: #5204

before:

https://github.com/toeverything/blocksuite/assets/15389209/dfbdc1ff-a42d-4d4f-a989-04044547921d

after:

https://github.com/toeverything/blocksuite/assets/15389209/623685ae-beed-4276-b37e-dcfc3dc3d427

### DIFF
--- a/packages/blocks/src/paragraph-block/paragraph-block.ts
+++ b/packages/blocks/src/paragraph-block/paragraph-block.ts
@@ -231,6 +231,16 @@ export class ParagraphBlockComponent extends BlockElement<ParagraphBlockModel> {
       this.requestUpdate();
     });
     this.model.childrenUpdated.on(() => this.requestUpdate());
+
+    this.page.awarenessStore.slots.update.on(v => {
+      const remoteSelections = this.std.selection.remoteSelections.get(v.id);
+      const textSelection = remoteSelections?.find(
+        selection => selection.type === 'text'
+      );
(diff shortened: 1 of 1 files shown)


lawvs (approved):

packages/blocks/src/paragraph-block/paragraph-block.ts
```diff
@@ -231,6 +231,16 @@ export class ParagraphBlockComponent extends BlockElement<ParagraphBlockModel> {
       this.requestUpdate();
     });
     this.model.childrenUpdated.on(() => this.requestUpdate());
+
+    this.page.awarenessStore.slots.update.on(v => {
+      const remoteSelections = this.std.selection.remoteSelections.get(v.id);
+      const textSelection = remoteSelections?.find(
+        selection => selection.type === 'text'
+      );
+      if (textSelection) {
```
Flrande, line 240: Why we need to check `textSelection` here?

zqran: Suppose there are two users A and B. User A creates `#`. At this time, user B only needs to update the placehoder when it senses data changes. Here, `remoteSelections` is mainly used to determine whether it is user B.

Only when it is a text selection, it means that the user is in editing mode. This can avoid unnecessary updates.

PR #5304, opened 2023-11-13
fix(edgeless): frames update in navigation

### LINKED ISSUE #5284: After deleting a frame, the presentation will still display the deleted frame
version：insider /  0.10.0 beta(11.10)

In the following video, four frames are selected in sequence, and then one of them (frame3) is deleted. However, during the presentation, the deleted frame3 will still be displayed.

Image

regischen:
close #5284


packages/blocks/src/page-block/edgeless/components/toolbar/edgeless-toolbar.ts
```diff
@@ -267,23 +270,22 @@ export class EdgelessToolbar extends WithDisposable(LitElement) {
...
+          requestAnimationFrame(() => {
             this._updateFrames();
-          }
-        } else if (e.flavour === 'affine:frame') {
+          });
+        }
+      })
+    );
+    _disposables.add(
+      page.slots.yBlockUpdated.on(e => {
```
doodlewind, line 282: Better mark this as deprecated, but let's fix this currently.

doodlewind (approved):

Issue #5305, opened 2023-11-13
Missing slot event for local `xywh` update


regischen:
Due to the current updates of `xywh` being processed through local update and not updating the `YModel`, the previous implementation that relied on `yBlockupdate` slots has become ineffective.

currently found:
1. The component toolbar's position is not updating promptly.
2. The auto line's position is not updating promptly.

https://github.com/toeverything/blocksuite/assets/58546692/a8f84e1c-faca-4bb7-8145-b0bed245b405

doouding:
The `edgeless-page-component` already has a slot called `elementSizeUpdated`; modifying the local record will also trigger the slot. This should be sufficient for most scenarios.

PR #5311, opened 2023-11-13
fix(edgeless): local record issue

### LINKED ISSUE #5307: Incorrect `pickById` result related to local record manager
now 
```
surface.pickById(id) === surface.pickById(id) // false when id is block's id
```
it may cause some cache not work.

https://github.com/toeverything/blocksuite/blob/dfc91fc330669baf786e086493edb8aab21f0557/packages/blocks/src/surface-block/surface-block.ts#L942-L948

doouding:
Close #5307 
Related #5305 

### Change
- remove redundant slots `elementSizeUpdated` and `yBlockUpdated`
- move `elementUpdated`, `elementAdded`, `elementDeleted` slots to edgeless component
- modifying properties in the local record will also trigger slots mentioned in the second item of the list
- add cache for the wrapped block model
- wrapped model may not be usable on the page API, add `unwrap` method to get the non-proxied model
- save data to local record during resize

### DIFF
--- a/packages/blocks/src/_common/utils/iterable.ts
+++ b/packages/blocks/src/_common/utils/iterable.ts
@@ -135,17 +135,20 @@ export function pickArray<T>(target: Array<T>, keys: number[]): Array<T> {
   }, [] as T[]);
 }
 
-export function pick<T, K extends Partial<keyof T>>(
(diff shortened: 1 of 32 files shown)


doouding:
Tests related to drag and resize need to be supplemented here.

Issue #4435, opened 2023-08-29
The format bar should not break cursor jumping


lawvs:
https://github.com/toeverything/blocksuite/assets/18554747/4b12c7f4-baa3-44d5-8621-292b5e369547

https://github.com/toeverything/blocksuite/blob/86653c592e02d8f7384e20f6bceab45b78a4dc4d/tests/format-bar.spec.ts#L1404-L1424

Saul-Mirone:
The behavior will be fixed once we migrate the root content editable container. \cc @doodlewind

Issue #2684, opened 2023-05-21
Support holding Shift+Down to select next line


SolarDesalination:
### What happened?

On the linux version, when my text cursor is on a line, and I have text on another line below it, the expected use is that when hold the shift key and push down, that the next line of text is also highlighted. The current editor doesn't enable a shift-down text cursor highlight function.

Using Linux version.

### Version

app.affine.pro

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

himself65:
This is a common issue on both platforms

doodlewind:
This is partly supported if you are using block level selection:

https://github.com/toeverything/blocksuite/assets/7312949/1e5f4498-5227-4278-a838-34a78e8db4bf

But for the shift+down support under native text selection,  /cc @Saul-Mirone to tell if this is a low-hanging fruit.

Saul-Mirone:
> This is partly supported if you are using block level selection:
> 
>  Screen.Recording.2023-08-16.at.13.06.39.mov 
> But for the shift+down support under native text selection, /cc @Saul-Mirone to tell if this is a low-hanging fruit.

I don't think there're any low-hanging fruit for now. We can roughly support the feature  since we already have the <kbd>ArrowDown/ArrowUp</kbd> supported. However there're a lot of edge cases for keyboard navigation, we still need to polish and test every edge cases.

PR #5315, opened 2023-11-13
fix(page): shift + arrow text selection

AyushAgrawal-A2:
Closes: #2684 
Closes: #4435 
Closes: #1290

Changes:
- Text Selection: Implemented app logic for changing text selection using arrow keys inside the note block.
- Reverse Text Selection: If text selction has been made using mouse and user continues to modify the selection using keyboard, the selection should change from the last cursor position. For this to work for reverse text selection (right to left / bottom to top) I had to modify:
   - `_mergeRanges` to calculate if reverse selection is made.
   - `_renderRange` to render reverse selection correctly.
   - add `isReverse` property to `TextSelection` to store reverse selection state and calculate the start / end of selection.
- Format Bar: Format bar appears on text selection and hide the text, blocking it from caret position calculation. To prevent this I opted to hide all widgets before caret position calculation and then restore them to their original state.

https://github.com/toeverything/blocksuite/assets/54364088/ef57f25d-3f23-49cd-8f48-6d8bb9847584

### DIFF
--- a/packages/block-std/src/selection/variants/text.ts
+++ b/packages/block-std/src/selection/variants/text.ts
@@ -12,6 +12,7 @@ export type TextRangePoint = {
 export type TextSelectionProps = {
   from: TextRangePoint;
   to: TextRangePoint | null;
+  isReverse?: boolean;
 };
 
 const TextSelectionSchema = z.object({
@@ -27,6 +28,7 @@ const TextSelectionSchema = z.object({
       length: z.number(),
     })
     .nullable(),
+  isReverse: z.boolean().nullable(),
 });
 
 export class TextSelection extends BaseSelection {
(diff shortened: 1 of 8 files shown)


Saul-Mirone:
Great work! However, I think it won't work for multiple notes in doc mode?

doodlewind:
> Great work! However, I think it won't work for multiple notes in doc mode?

No worries, I'd say that could be an edge case for further PRs.

AyushAgrawal-A2:
> Great work! However, I think it won't work for multiple notes in doc mode?

Currently it works within one note only. It can be extended for multi note as well. I have made a lot of changes here. Can I add that in another PR ?

Saul-Mirone:
> > Great work! However, I think it won't work for multiple notes in doc mode?
> 
> Currently it works within one note only. It can be extended for multi note as well. I have made a lot of changes here. Can I add that in another PR ?

Sure.

Saul-Mirone (approved):

Issue #5285, opened 2023-11-10
Stale number list index state


Yuhang1015:
When the number list in page mode is destroyed, the numbers cannot be modified automatically, nor can they be modified by entering numbers + dots + spaces.

![Image](https://github.com/toeverything/blocksuite/assets/142964904/6f2a7208-d4bd-462f-8b1d-dd154549eb9d)

PR #5326, opened 2023-11-14
fix(page): stale number list index state

zqran:
close: https://github.com/toeverything/blocksuite/issues/5285

before:

https://github.com/toeverything/blocksuite/assets/15389209/33852ebb-13ff-418f-816d-833051d7e2a3

after:

https://github.com/toeverything/blocksuite/assets/15389209/351c46f7-6515-45cf-b885-20de4834d5be

### DIFF
--- a/packages/blocks/src/_common/components/rich-text/rich-text-operations.ts
+++ b/packages/blocks/src/_common/components/rich-text/rich-text-operations.ts
@@ -389,6 +389,7 @@ function handleListBlockBackspace(page: Page, model: ExtendedModel) {
   const parent = page.getParent(model);
   if (!parent) return false;
 
+  const nextSiblings = page.getNextSiblings(model);
   const index = parent.children.indexOf(model);
   const blockProps = {
     type: 'text' as const,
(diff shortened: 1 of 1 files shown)


doodlewind (commented):
This looks very clever since it won't dispatch redundant updates over network, but triggering `model.propsUpdates` using empty `page.updateBlock`.

However, this won't work for real time collaboration, since remote client won't receive the event to update:

https://github.com/toeverything/blocksuite/assets/7312949/67814134-a76f-4ec2-8171-39be7c3bd194

Any idea @Flrande @Saul-Mirone ?

zqran:
By the way, when using `Ctrl+z` to undo, the second sequence number `1.` does not change to `3.`. After clicking, it will change to 3.


https://github.com/toeverything/blocksuite/assets/15389209/f79b0056-3ef7-42d0-b8a7-246b2f2fa4e2

zqran:
This seems to be a common problem, for example, the same problem occurs when unindent.

https://github.com/toeverything/blocksuite/assets/15389209/e33a484d-cda8-48a4-bc99-480b05da47b0

Flrande:
> Any idea @Flrande @Saul-Mirone ?

I think we need to make list index into CRDT.

doodlewind:
> I think we need to make list index into CRDT.

@Flrande I have concern about the correctness of state in this case. If we store index state into CRDT, concurrent edits will be more error-prone, and it's harder to cleanup potential dirty state - currently refreshing the page should reset the indexes correctly.

Flrande:
> @Flrande I have concern about the correctness of state in this case. If we store index state into CRDT, concurrent edits will be more error-prone, and it's harder to cleanup potential dirty state - currently refreshing the page should reset the indexes correctly.

If we want to give users more flexible control over list numbering (such as allowing them to control where the numbering starts in Feishu documents, or to change the numbering directly), we will eventually have to incorporate numbering into the CRDT. However, our current control over list numbering is still quite rudimentary, and there is no immediate product demand for such features. The way we handle it now is still barely acceptable.

doodlewind (approved):
Should land this for now, but require best practice afterwards.

Issue #4622, opened 2023-09-06
There is some delay in inserting attachments via drag and drop


houjoe0829:
In the AFFiNE Canary version, there will be a significant delay when inserting attachments by dragging and dropping, and the screen recording has been attached.

https://github.com/toeverything/blocksuite/assets/22443345/1f0aa02c-d867-4a0d-a524-04336d035440

doodlewind:
Closing as not reproduced in current internal.

PR #5332, opened 2023-11-14
fix(page): delay when inserting attachments via dnd

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/4622

Now the attachment can correctly show a loading state when inserted via drag and drop.

https://github.com/toeverything/blocksuite/assets/18554747/00c04e29-77b7-40cd-8ae9-1103eb159b78

This PR also sunsetting the `loadingKey` since I found the `page.addBlock` function supports the `id` parameter which is a better alternative to identify the block.

No need to perform the migration action because all legacy `loadingKey` will be ignored.

Note: other collaborators will see an error attachment block when the blob has not finished uploading. This issue can be resolve by sync the upload status through the awareness system in the future.

### DIFF
--- a/packages/blocks/src/attachment-block/attachment-model.ts
+++ b/packages/blocks/src/attachment-block/attachment-model.ts
@@ -9,11 +9,15 @@ export type AttachmentBlockProps = {
   type: string;
   caption?: string;
 
-  // `loadingKey` is used to indicate whether the attachment is loading
-  // You can query the loading state by `isAttachmentLoading(loadingKey)`
-  // We can not use `loading: true` directly because the state will be stored in the model
-  //
-  // The `loadingKey` and `sourceId` should not be existed at the same time
+  /**
+   * `loadingKey` is used to indicate whether the attachment is uploading.
+   * You can query the loading state by calling `isAttachmentLoading(loadingKey)`
+   *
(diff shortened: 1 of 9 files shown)


doodlewind (approved):
@Flrande I'd consider for this kind of changes, the migration would be overkill since it would bring HUGE maintaining costs for the whole team (not just our editor maintainers). As long as there is no breaking change, we should not perform migration, which is a tradeoff we intentionally chose.

packages/blocks/src/attachment-block/attachment-model.ts
```diff
@@ -21,6 +21,10 @@ export type AttachmentBlockProps = {
    */
   type: string;
   caption?: string;
+  // `loadingKey` was used to indicate whether the attachment is loading,
+  // which is currently unused but no breaking change is needed.
+  // The `loadingKey` and `sourceId` should not be existed at the same time.
+  // loadingKey?: string | null;
```
doodlewind, line 27: I'd consider removing an unused model key as non breaking change, but we should list the definition of that field in the codebase comment, which could be good for potential debugging in the future.

cc @Flrande @Saul-Mirone @doouding @regischen @donteatfriedrice

Issue #4972, opened 2023-10-09
Database hover animation curve


zanwei:
We would like an update on the hover animation for the database

In principle, hover animations have both "**start animations**" and "**end animations**." We aim to keep them consistent here.

The parameters would be `0.42, 0, 1, 1 ` in `150ms`

![image 27](https://github.com/toeverything/blocksuite/assets/40333372/e977eb7b-384d-41c1-92a8-34f4fe6664d9)

**Use Case:**

This animation will be applied to

<img width="681" alt="image" src="https://github.com/toeverything/blocksuite/assets/40333372/420da9e1-1ce9-429f-9406-901c517d76a9">

AyushAgrawal-A2:
Hi @zanwei ,
Please clarify if animation is required on row operation icons as well ? These also show up on hover.

![scrnli_11_15_2023_1-21-44 PM](https://github.com/toeverything/blocksuite/assets/54364088/1cd05dc6-f7bf-4ef0-8699-2797ebdc6c21)

zanwei:
Yes

AyushAgrawal-A2:
@zanwei no problem, I have reverted hover animation on operation icons.

PR #5336, opened 2023-11-15
feat(database): database hover animation

AyushAgrawal-A2:
Closes: #4972 

Changes:
- Added hover animation on the mentioned elements in database block.


AyushAgrawal-A2:
Waiting to reply to https://github.com/toeverything/blocksuite/issues/4972#issuecomment-1811964690 before submitting for review.

doodlewind (commented):
Thanks for the improvement! Just want to know if the following 2 places could support the hovering animation together?

https://github.com/toeverything/blocksuite/assets/7312949/24e4d919-a4f4-4bf0-aeb1-33786fb20c4d

AyushAgrawal-A2:
> Thanks for the improvement! Just want to know if the following 2 places could support the hovering animation together?
> 
>  Screen.Recording.2023-11-16.at.08.11.32.mov

As these two elements are part of different parent elements, I can think of two ways of doing this:
1. Using js, add eventlistener for hover on both and then add class to all linked components.
2. Using css :has() selector. This is a new selector and not present in current release of firefox. It is available on firefox 120 nightly build though. (ref: https://stackoverflow.com/questions/76133179/cross-browser-alternative-to-css-has-pseudo-class). As affine using chromium under the hood, this should work for our use case.

Screencast using css :has() selector on chrome.

https://github.com/toeverything/blocksuite/assets/54364088/11a621ad-042b-4c58-9455-389624114faf

Also, please clarify if 
1. only hover state is to be linked b/w `+` svg and `Add` button, or
2. same animation `150ms (0.42, 0, 1, 1)` also needs to be added to background color change on hover.

currently background changes on hover with no animation. I am confused because of the term `hovering animation together`

doodlewind:
The effect looks great! Using `has` should be adequate since it's a styling enhancement, and let's make the animation added on background color change :D

doodlewind (approved):
Your work is impressive! If you are looking for GitHub sponsor, feel free to DM me at doodlewind@toeverything.info :D

PR #5339, opened 2023-11-15
feat(edgeless): support more fonts in canvas text

### LINKED ISSUE #5313: Edgeless Toolbar Update
We would like to have a feature update for Edgeless Toolbar, please refer [design](https://www.figma.com/file/UTTtfVCcStw1dKUg0gCSPo/%F0%9F%93%9A-Design-System?type=design&node-id=13008-29020&mode=design&t=si6rvpKa7FN3MaFQ-4)
image

### Update Details

**1. Font Style**
We plan to add 6 font families to enrich the diversity of Edgeless. Additionally, the fonts of the current versions of **General** and **Scribble** will use their own names, which are **Inter(General)** and **Kalam(Scribble)**

It should be noted that when users switch fonts, the `Aa` in the Toolbar will change according to the style of the font.

image

**2. Font Size**
We have decided to adopt a method of directly selecting numerical values for adjusting font size. Please note that in the current version, the font size will not change when users input a value greater than `200`. However, the value will increase with the extent of dragging. Therefore, please remove this restriction, but limit the maximum font size input by users to not exceed `9999`.
image
image

**3. Font Weight**
In the new version, we will allow users to choose different `font-weight`. However, please note that we will limit the number of `font-weight`. Currently, each font will have a maximum of three weights and a minimum of one weight. 

- Please note that if the font has only one font-weight, the dropdown menu will be **disabled**.
- Please be aware that the names of the font-weights for each font may vary (some might be **Light, Regular, Medium**, while others could be **Regular, Medium, SemiBold**). **The specific names should be based on the font files provided by the designer.**

image
image

[Font Resources.zip](https://github.com/toeverything/blocksuite/files/13333180/Font.Resources.zip)

Flrande:
Close #5313

Breaking change in `surface-model`: `bold` -> `fontWeight`, `italic` -> `fontStyle`

By default, [font files are loaded through CDN](https://github.com/toeverything/blocksuite/pull/5339/files#diff-136f7c6594a5c5166fc603df6127280639cfcb55cc1f6309e5efaa3bb63f57c3R28-R126). If you want to reference local font files, you can modify this behaviour by overriding `page-service` (similar to #5158).


zanwei:
Looks good! @Flrande. However, there're some issues of UI

1. The text content needs to be all left-aligned, and the font needs to use the font's own style to ensure that users have an expectation of what the font will look like before choosing it.
image

2. If the font has only one font-weight, please disable the entry for selecting font-weight.
image
image

3. font-size issue still needs to fix
image

doodlewind (commented):
What will happen when opening existing canvas text element? I'd consider losing stying info is acceptable as long as no content lost or migration required.

Flrande:
> What will happen when opening existing canvas text element? I'd consider losing stying info is acceptable as long as no content lost or migration required.

If migration can run normally, the original style will be preserved.

https://github.com/toeverything/blocksuite/pull/5339/files#diff-a66b1dd0ce968e0f79ba989797b3636ba818af20f3a69664115006a9d33d9aa3R119-R138

Do we need to add a fallback logic for unexecuted migration? (to maintain normal rendering at the cost of losing the style)

doodlewind:
I don't think we should do migration for this kind of UX upgrade. Given the status in AFFiNE, simply wiping away existing style sounds better as a tradeoff.

Flrande:
> I don't think we should do migration for this kind of UX upgrade. Given the status in AFFiNE, simply wiping away existing style sounds better as a tradeoff.

If we simply remove the original styles, it means that `fontWeight` and `fontStyle` would need to become optional fields. However, this would lead to a wider range of code needing to adapt to it, which is difficult to say is a best practice. Canvas text is just one of the drawing elements under the surface, and even if the migration fails, its impact will only be limited to itself. The worst consequence is just that users may need to click the `fontWeight` and `fontStyle` settings buttons again, which isn't so severe that it becomes unusable. However, if we simply ignore the users' previous styles because of this update, it might lead to a worse user experience. For example, if a user has dozens of styled canvas texts, they would need to adjust them all back.

In short, I understand your concerns about the breaking changes to the data structure, but I believe it is different from the breaking changes of the AFFiNE GUID. Its impact is limited, and it will not lead to widespread rendering failures. Moreover, we will undergo multiple rounds of testing before releasing it. I think we don’t need to be so conservative at this point.

doodlewind:
I believe in this case, having the application layer compatible with legacy docs missing these new fields should be the expected tradeoff.

Given the cost scheduling and validating the migration, I am sure that text style does NOT deserve a breaking change. Migration should NOT happen more often than half a year.

And after all, due to the fragmented nature of local-first app, I even have the feeling that we will have to stop introducing ANY breaking changes - there is a famous slogan called "don't break the web", and the whole web APIs doesn't introduce breaking changes over DECADES.

Flrande:
>  after all, due to the fragmented nature of local-first app, I even have the feeling that we will have to stop introducing ANY breaking changes - there is a famous slogan called "don't break the web", and the whole web APIs doesn't introduce breaking changes over DECADES.

It sounds like taking a cold breath in.🥶

doodlewind:
@Flrande no worries, we'll figure out alternatives preventing breaking changes :)

Current breaking changes will lead to a new CRDT workspace, which could be still be the escape hatch, but I'd consider this to be an exclusive feature for AFFiNE Cloud.

doodlewind (approved):

Issue #5303, opened 2023-11-12
Incorrectly keyboard navigator in code block


lawvs:
When the line is empty, the ⬆️ and ⬇️ behavior is incorrect in the code block.

This issue seems related to the size of the editor width.

https://github.com/toeverything/blocksuite/assets/18554747/ef7d69f4-f1c5-4d14-b7ed-e1ac02f8b273

PR #5341, opened 2023-11-15
fix(page): arrow up/down navigation in code block

AyushAgrawal-A2:
Closes: #5303 

Changes:
- Using exact x position of cursor rect in an empty line is causing incorrect next caret calculation. Empty line has no padding. Here X position becomes exactly the boundary of the container. It skips the node above / below if the text there has some padding. Added 1 px to cursor x pos avoid resulting in exact container boundary.
- Adding just 1 px to x position cause some test cases to fail. This happened where text has narrow width letters such as letter `l` in hello world. Here the cursor would land 1 index right to the expected result, just because of 1 px extra gap. I changed the text in these test cases to all letter `a`, as letter `a` has full width and hence results in consistent cursor landing.


doodlewind (approved):
Thanks for your help maintaining the selection details!

PR #5342, opened 2023-11-15
refactor: remove mode in lit root

Saul-Mirone:


packages/blocks/src/surface-block/surface-block.ts
```diff
@@ -170,7 +170,7 @@ export class SurfaceBlockComponent extends BlockElement<SurfaceBlockModel> {
   }
 
   private get _isEdgeless() {
-    return this.root.mode === 'edgeless';
```
doodlewind, line 173: @regischen please request review from @Saul-Mirone for further modifications in `store`, `block-std`, and `lit`.

PR #5343, opened 2023-11-15
feat(edgeless): optimize connector auto align

### LINKED ISSUE #5183: Optimize the auto align offset of connector
Currently the auto align offset of connector is computed under model coords, so this only works well in large zoom level:

image

But the best practice should be using the viewport coords in computing the auto alignment, so that this feature could work in small zoom level.

lzlme:
resolve #5183 

before:

https://github.com/toeverything/blocksuite/assets/117659326/7bd2e385-03ac-481a-a0f7-3fe51da12ad7

after:

https://github.com/toeverything/blocksuite/assets/117659326/bb94b72e-2d93-47cc-8f77-f7f85d72f7d9


doodlewind (approved):
Thanks for the improvement!

PR #5347, opened 2023-11-16
feat(edgeless): allow canvas content interleaving with DOM based on `LayerManager`

doouding:
For now, it's just a prototype. Technical details will be supplemented later and any feedback is welcomed.

Todo:
- [x] Move the edgeless portal to surface
- [x] Add support of multiple canvases rendering for surface block
- [x] Bug & Test


doodlewind (approved):
Awesome work, thanks!

PR #5349, opened 2023-11-16
fix(page): support setting index of ordered list

### LINKED ISSUE #3008: List rendering is not as expected
I have a markdown like this
```
FAQs:

1. list1

list1 content

2. list2

list2 content

3. list3

list3 content

- list3-item
- list3-item
- list3-item
- list3-item
- list3-item

4. list4

list4 content

5. list5

list5 content

```

the result i want
image

The result I actually get
image

Is it my usage problem or maybe it's a bug?

AliceLanniste:
closed #3008


doodlewind (approved):
Thank for the quick fix!

PR #5354, opened 2023-11-16
refactor(page): use keymap controller for note block

Saul-Mirone:


packages/blocks/src/note-block/keymap-controller.ts
```diff
@@ -0,0 +1,789 @@
...
+  getBlockSelectionBySide,
+  moveCursorToBlock,
+  moveCursorVertically,
+  pathToBlock,
+  selectBetween,
+  setBlockSelection,
+  setTextSelectionBySide,
+} from './utils.js';
+
+export class KeymapController implements ReactiveController {
```
doodlewind, line 30: How do you distinguish the concept between "controller" and "service"? Is this also capable of being defined as a service?

packages/blocks/src/note-block/note-block.ts
```diff
@@ -17,10 +17,7 @@ export class NoteBlockComponent extends BlockElement<NoteBlockModel> {
     }
   `;
 
-  override connectedCallback() {
-    super.connectedCallback();
-    bindHotKey(this);
-  }
+  keymapController = new KeymapController(this);
```
doodlewind, line 20: Should we further promote this pattern of removing setups in `connectedCallback`?

doodlewind (approved):
Great to see legacy operation to be removed!

packages/blocks/src/note-block/keymap-controller.ts
Saul-Mirone: The controller here refers to the [reactive controller](https://lit.dev/docs/composition/controllers/) in lit.

packages/blocks/src/note-block/note-block.ts
Saul-Mirone: I think it's a good pattern for sharing logic in lit. I think it's better than maintain lifecycle in `xxxManager` because it can share the life cycle with lit element.

Issue #5351, opened 2023-11-16
Paste not work when trying to create a link in edgeless


lawvs:
https://github.com/toeverything/blocksuite/assets/18554747/552f6bb5-937e-4148-aba3-fc4d596e84aa

PR #5355, opened 2023-11-16
fix(page): paste not work when trying to create a link in edgeless

zqran:
close: #5351

### DIFF
--- a/packages/blocks/src/_common/components/rich-text/virgo/nodes/link-node/link-popup/link-popup.ts
+++ b/packages/blocks/src/_common/components/rich-text/virgo/nodes/link-node/link-popup/link-popup.ts
@@ -81,6 +81,20 @@ export class LinkPopup extends WithDisposable(LitElement) {
     );
   }
 
+  protected override firstUpdated() {
+    if (!this.linkInput) return;
+
+    this._disposables.addFromEvent(this.linkInput, 'copy', e => {
+      e.stopPropagation();
+    });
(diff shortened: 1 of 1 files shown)


packages/blocks/src/_common/components/rich-text/virgo/nodes/link-node/link-popup/link-popup.ts
```diff
@@ -81,6 +81,20 @@ export class LinkPopup extends WithDisposable(LitElement) {
     );
   }
 
+  protected override firstUpdated() {
+    if (!this.linkInput) return;
+
+    this._disposables.addFromEvent(this.linkInput, 'copy', e => {
```
doodlewind, line 87: @Saul-Mirone do we have some framework encapsulation for this kind of logic?

lawvs: To resolve this issue, migrating the link popup to a portal-style format can be a good solution.

Saul-Mirone: We don't have it for now. Maybe we can have something like `prevent events` config in the future. But it will be a low priority task to me because this pattern is rare for now. We can give it a consider if logic like this grows rapidly.

doodlewind (approved):

PR #5357, opened 2023-11-16
fix(page): tweak attachment styles

lawvs:
- Fix icon styles when the name of the attachment exceeds one line

Screenshot 2023-11-17 at 00 24 08

- Fix line break when the name of the attachment is lengthy
- Fix the height of the banner layer incorrectly

Screenshot 2023-11-17 at 00 26 30


doodlewind (approved):
Thanks for the effort optimizing UI details!

PR #5364, opened 2023-11-17
fix: backwards compatible with remote selection

lawvs:
Fix the error log in https://github.com/toeverything/AFFiNE/issues/4980

Screenshot 2023-11-17 at 21 42 08

```
slot.js:99 [
  {
    "code": "invalid_type",
    "expected": "boolean",
    "received": "undefined",
    "path": [
      "isReverse"
    ],
    "message": "Required"
  }
]
```

Since the remote data is unreliable, we should ignore the error when parsing failure.


Saul-Mirone (approved):
LGTM

doodlewind (approved):

Issue #5246, opened 2023-11-07
Link popup unexpectedly disappear when hovering from bottom


lawvs:
The `link-popup` will disappear suddenly sometimes if the link is hovered from the bottom.

https://github.com/toeverything/blocksuite/assets/18554747/4d0cf1ff-0d69-425e-998d-f0ab78a106c7

PR #5369, opened 2023-11-20
fix(page): link popup unexpectedly disappear when hovering from bottom

zqran:
close: #5246

https://github.com/toeverything/blocksuite/assets/15389209/12d4f693-ed8a-4f32-ab9e-236eca8d51ba


doodlewind (approved):
Good catch, thanks!

PR #5378, opened 2023-11-20
fix(edgeless): reset cursor when the selection changes

lawvs:
Fixing the cursor style will not restore when the shape is deleted.

https://github.com/toeverything/blocksuite/assets/18554747/fb64bb06-3fb8-4f6c-9fa8-b6188b57d5a0


doodlewind (approved):
Good catch!

PR #5396, opened 2023-11-21
refactor(store): rewrite page event to support partial update

Saul-Mirone:
Make it possible to:

```ts
page.updateBlock(model, () => {
  model.foo = 'foo';
  model.scope.bar = 'bar';
});
```

The whole event flow has been rewritten. Previously we use `observeDeep` to track prop changes in blocks. However, the problem of this flow is that when only part of a prop is changed, it's hard to track it in a top-down pattern.

Currently, we observe the changes in proxy to make it possible to notify changes when some part of the data has been updated.


doodlewind (commented):
Requires summary of the API surface update in PR desc 🤔

Flrande (approved):

Issue #5353, opened 2023-11-16
Click at doc blank area will auto scroll to page bottom


donteatfriedrice:
Click at doc blank area will focus on last text block and auto scroll to page bottom.  It is a little confused when page is long. 

https://github.com/toeverything/blocksuite/assets/99816898/c5d4c6b1-57ea-43ac-9869-0064f8cf9dfd

DeepakSR-12:
Hi @doodlewind  @donteatfriedrice . I'd like to work on this issue. Please update, what is the expected behaviour?

doodlewind:
@DeepakSR-12 clicking on the blank areas on the left and right sides should not result in scrolling, thanks!

PR #5405, opened 2023-11-22
fix(page): prevent scrolling to bottom of page when clicking on sides of document

zqran:
close: #5353


packages/blocks/src/page-block/doc/doc-page-block.ts
```diff
@@ -511,6 +511,25 @@ export class DocPageBlockComponent extends BlockElement<
       ) {
         return;
       }
+
+      // Click on blank area
```
doodlewind, line 515: I think you should extract the onclick handler into sub methods. Simply adding logic here makes it looks verbose.

doodlewind (approved):

PR #5409, opened 2023-11-23
fix(page): extra space between dot and ordinals

### LINKED ISSUE #5327: There are unwanted spaces between ordinals and dot (period) in numbered lists
Screenshot 2023-11-14 at 14 23 35

The above screenshot is taken from usage of the `<simple-affine-editor></simple-affine-editor>` element.

There should be no space, i.e. it should be:
```
1. This is the top level.
  a. Second level
    I. Third level
    II. Third level (2)
````

Notice how, in fact there is no space on `II. Third level (2)`, but all the other items include the unwanted space.

AliceLanniste:
close #5327


doodlewind (approved):
Thanks for the trying! Several test cases seems may be affected due to the change on spacing, would be great to land this once they are fixed.

PR #5416, opened 2023-11-23
fix(page): disable database action in readonly

lawvs:
https://github.com/toeverything/blocksuite/assets/18554747/0dad738e-b4b4-4063-949b-91cd3ad9bc15


doodlewind (approved):
Good catch!

Issue #4374, opened 2023-08-25
Refactor outdated rich text operation


Flrande:
related: [legacy.ts](https://github.com/toeverything/blocksuite/blob/33d15dc6e57152301426071cdc3887e871d825ea/packages/blocks/src/__internal__/rich-text/keymap/legacy.ts), [rich-text-operations.ts](https://github.com/toeverything/blocksuite/blob/33d15dc6e57152301426071cdc3887e871d825ea/packages/blocks/src/__internal__/rich-text/rich-text-operations.ts)

doodlewind:
Consider supported after #5878, the remaining public APIs for exposing new commands would require new issues to track.

PR #5417, opened 2023-11-23
fix(edgeless): limit getNextBlock/getPrevBlock to same note

### LINKED ISSUE #5096: Pressing delete will move note content from one note to another note.
https://github.com/toeverything/blocksuite/assets/59462000/095994e5-763e-474c-ae28-9cdbba29c751

AyushAgrawal-A2:
closes #5096 

Changes:
- In edgeless mode, notes can be scattered in random order and text operation should be contained within the same note. Added condition in getPrevBlock & getNextBlock to limit search to same note in edgeless mode.

Before:

https://github.com/toeverything/blocksuite/assets/54364088/02e7acf9-b45c-4ae7-9eab-01a2c3f391e4

After:

https://github.com/toeverything/blocksuite/assets/54364088/2b5940be-a838-4bc6-ac18-fcf9e4b04399


Flrande:
Most of the functions in `query.ts` and `rich-text-operations.ts` will be rewritten and deprecated. As public functions, they contain a large amount of logic for judging block flavour, which is not conducive to our future functional expansion and maintenance. In addition, the operational targets of these functions are block models, but under our new architecture, our operational objects should be block elements, because one block model may correspond to multiple block elements. I plan to implement a corresponding command for each block to extract these block flavour related logics from the public functions. The related issue is #4374.

Therefore, for this fix, a better approach would be to use the existing `get-next-block` command, which will require refactoring the related call chain.

Flrande:
You can check #5418 and know why we need to refactor these utils. :P

doodlewind (approved):
I think we can land this for now, leaving the updates of `query` in the refactoring afterwards.

PR #5420, opened 2023-11-24
feat(edgeless): add toggle switch component

donteatfriedrice:
Screenshot 2023-11-24 at 10 59 10

### DIFF
--- a/packages/blocks/src/_common/components/toggle-switch.ts
+++ b/packages/blocks/src/_common/components/toggle-switch.ts
@@ -59,12 +59,12 @@ export class ToggleSwitch extends LitElement {
   on = false;
 
   @property({ attribute: false })
-  onChange?: () => void;
+  onChange?: (on: boolean) => void;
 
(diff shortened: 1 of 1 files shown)


packages/blocks/src/_common/components/toggle-switch.ts
```diff
@@ -0,0 +1,82 @@
...
+    left: calc(100% - 1px);
+    transform: translateX(-100%);
+  }
+
+  label:active:after {
+    width: 24px;
+  }
+`;
+
+export class ToggleSwitch extends LitElement {
```
regischen, line 55: I think maybe the component is used by other components.
so when we use it, Is it supposed to be pass `state` and `onChange` to the switch component?
what do you think?

donteatfriedrice: That make sense. I think the 'on' state can be passed in, maybe I need to add onChange.

regischen (approved):

PR #5423, opened 2023-11-24
feat: add sessionStore singleton class

regischen:
Add singleton class `Storage` mainly for prefix and suffix encapsulation and provide a clean api.

### DIFF
--- a/packages/blocks/src/_common/utils/edgeless.ts
+++ b/packages/blocks/src/_common/utils/edgeless.ts
@@ -1,6 +1,4 @@
-function buildViewportKey(pageId: string) {
-  return 'blocksuite:' + pageId + ':edgelessViewport';
-}
+import { storage } from './storage.js';
 
 type ViewportData =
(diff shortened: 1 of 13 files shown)


doodlewind (changes requested):
Occupying the concept of "storage" here sounds overkill, I would suggest some naming change 🤔

Also this seems to break the Node compatibility test.

regischen:
> Occupying the concept of "storage" here sounds overkill, I would suggest some naming change 🤔
> 
> Also this seems to break the Node compatibility test.

Ok, after asking GPT, how about `SessionStore`.
Let me check the Node test problem.

doodlewind:
No, I don't think having another standalone store for session align with our plan. The plan is to unify multiple kinds of states with different durability. See https://chat.openai.com/share/6e5ee93b-1328-4d40-94ca-9db1623ef238 for current memo.

cc @Saul-Mirone

PR #5433, opened 2023-11-24
feat(store): add `key` field for props updated slot

Saul-Mirone:
## 1. `block.propsUpdated` will be triggered with the changed `key`.

```ts
model.slots.propsUpdated.on(({ key }) => {
  if (key === 'someProp') {
    doSomething();
  }
})
```

## 2. `page.blockUpdated` will be triggered with `type: updated` when it's props changed.


doodlewind (commented):
`name` -> `key` sounds better for the naming.

Saul-Mirone:
> `name` -> `key` sounds better for the naming.

Changed to `key`

PR #5438, opened 2023-11-24
fix(page): make sure closest note inside same editor page when there are multiple editors

### LINKED ISSUE #5418: `isPageMode` and some other utility functions does not count for multiple editor instances on DOM
As stated in the title, some utility functions like `isPageMode` [link](https://github.com/toeverything/blocksuite/blob/ce7ac88fc750fb465a6b4227b8f93eafe8e894fc/packages/blocks/src/_common/utils/query.ts#L205) does not take multiple editor instances into account. This may resulting in some weird behaviors if we render two editors side by side.

The above issue has caused problems in AFFiNE while working on the history modal feature. In our use case, we have a regular blocksuite editor and another editor that is located within a modal dialog.

If the background editor is in edgeless mode, and the modal is in page mode, one error will be thrown in the drag handle module in https://github.com/toeverything/blocksuite/blob/ce7ac88fc750fb465a6b4227b8f93eafe8e894fc/packages/blocks/src/_common/widgets/drag-handle/index.ts#L643

CleanShot 2023-11-24 at 01 23 26@2x

CleanShot 2023-11-24 at 01 25 32@2x

donteatfriedrice:
To close #5418 

https://github.com/toeverything/blocksuite/assets/99816898/57e352da-873c-49bf-bc1e-8a0ea02dc5a1

### DIFF
--- a/packages/blocks/src/_common/widgets/drag-handle/index.ts
+++ b/packages/blocks/src/_common/widgets/drag-handle/index.ts
@@ -641,6 +641,8 @@ export class AffineDragHandleWidget extends WidgetElement<
   }
 
   private _canEditing = (noteBlock: BlockElement) => {
+    if (noteBlock.page.id !== this.page.id) return false;
+
     if (isPageMode(this.page)) return true;
     const edgelessPage = this.pageBlockElement as EdgelessPageBlockComponent;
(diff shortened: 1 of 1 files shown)


doodlewind (approved):
Thanks for the swift fix!

Issue #5449, opened 2023-11-26
Right-Side Clipping Issue with 'More Toolbar'


alpakaka0o0:
**Description**
<img width="1512" alt="cut off screenshot" src="https://github.com/toeverything/blocksuite/assets/115137708/92dde8fe-564e-4d36-8fb8-604098a109d1">
If the window of 'more toolbar' is attached to the right side of the screen, it may be cut off.

**Use case**
_No response_

**Anything else?**
_No response_

**Are you willing to submit a PR?**
- [x] Yes I'd like to help by submitting a PR!

PR #5453, opened 2023-11-26
fix(edgeless): toolbar right-side clipping position

alpakaka0o0:
closes https://github.com/toeverything/blocksuite/issues/5449

Changes:

  The value of the _popperShow variable in the more-button is now passed to the component-toolbar. In the component-toolbar.ts, I modified the _computePosition() function to calculate the left based on the presence of the more-actions-container.
  
  
Before:
before

After:

https://github.com/toeverything/blocksuite/assets/115137708/a6979e11-3f95-4b82-adbb-a34d291c99df


packages/blocks/src/page-block/edgeless/components/component-toolbar/component-toolbar.ts
```diff
@@ -200,6 +203,11 @@ export class EdgelessComponentToolbar extends WithDisposable(LitElement) {
       : nothing;
   }
 
+  protected getPopperShow = (popperShow: boolean) => {
+    this._popperShow = popperShow;
```
doodlewind, line 207: You should not set a value inside a `getSomething` method, whose semantics should be read only.

doodlewind (commented):

packages/blocks/src/page-block/edgeless/components/component-toolbar/component-toolbar.ts
alpakaka0o0: Thank you for review my code. I will fix it quickly!

doodlewind (approved):
Thanks for the fix!

Issue #5451, opened 2023-11-26
Progress Filtering Issue


sseooh:
**Description**

https://github.com/toeverything/blocksuite/assets/105717406/961740ab-9595-444e-b715-571d025a5737

If you set the filter after setting the column type of the database table view as progress, it will not work properly.

**Use case**
_No response_

**Anything else?**
_No response_

**Are you willing to submit a PR?**
- [x] Yes I'd like to help by submitting a PR!

PR #5454, opened 2023-11-26
fix(database): table view progress filtering

sseooh:
The reason why filtering for the progress value did not work properly seems to be because the initialization did not proceed properly when a new column (type=progress) was added to the table view. Therefore, the code was modified to proceed with initialization in the function that sets the property type in the file 'database-block-datasource.ts'.

The video below shows filtering operating normally after modifying the code.

https://github.com/toeverything/blocksuite/assets/105717406/fd46b2a5-71b4-4436-a2ef-af209e94de17

Issue: #5451


doodlewind (commented):

zzj3720:
Considering the amount of data, doing this to add a column to a large table may require initializing a very large number of cells, and it's best to deal with this situation in the filter logic like this
```ts
  greatThan: {
    type: tFunction({
      args: [tNumber.create(), tNumber.create()],
      rt: tBoolean.create(),
    }),
    label: '>',
    impl: (value, target) => {
      value = value ?? 0;
      if (typeof value !== 'number' || typeof target !== 'number') {
        return true;
      }
      return value > target;
    },
  },
```

sseooh:
Thank you for reviewing my code. I will fix it quickly!

zzj3720 (approved):

doodlewind (approved):

PR #5455, opened 2023-11-26
fix(page): shift + arrow text selection on a page having multiple notes

AyushAgrawal-A2:
continuation of #5315


doodlewind:
I think this is a relatively rare edge case. We can just leave the PR here for anyone interested to continue, but our development resources need to be concentrated on tasks with higher priorities FTM :)

PR #5457, opened 2023-11-26
fix: incorrect page tag popup position on zooming

### LINKED ISSUE #5456: Incorrect page tag popup position on zooming
**Description**

https://github.com/toeverything/blocksuite/assets/89682253/d225b326-d7e2-4d0c-affc-7dfe3bdec82c

This problem occurs when you zoom in and out of the screen while selecting multiple tags.

**Use case**

**Anything else?**

**Are you willing to submit a PR?**

- [x] Yes I'd like to help by submitting a PR!

kimhyeonju:
Close #5456

This problem occurs when you zoom in and out of the screen while selecting multiple tags.

This problem occurs as the size of the selecting multiple tags window is automatically fixed when the screen is loaded. Therefore, we solved the above problem by adjusting the ratio to the percentage to fit the size of the screen.

I think the lower part of the "multi-tag-select.ts" file is a problem. I recommend that you delete or modify this part.
스크린샷 2023-11-26 오후 8 20 15


doodlewind (commented):

Flrande:
Close #5456

packages/blocks/src/_common/components/tags/styles.ts
```diff
@@ -3,13 +3,18 @@ import { css, unsafeCSS } from 'lit';
 
 export const styles = css`
   affine-multi-tag-select {
-    position: absolute;
-    width: 100%;
+    position: fixed;
```
doodlewind, line 6: I think we should not hard coded a `fixed` position, since the editor may be embedded in an container with dynamic offset (e.g., toolbars in AFFiNE). Also the `!important` style is something that should be avoided.

kimhyeonju: @doodlewind  I agree with your point of view. However, I've noticed an issue where the position of the modal becomes incorrect during dynamic use. How do you think this problem could be best resolved?

zzj3720:
This will break the multi-select columns of the database table.
This offset is also incorrect in the affine. which have top and sidebars.

zzj3720:
Suggest fixing this bug by adding autoUpdate here
https://github.com/toeverything/blocksuite/blob/178e79f0a88ab710633f0ee0b26f002a38b8530e/packages/blocks/src/_common/components/menu/menu.ts#L696-L732
It might look like this
```ts
  const unsub = autoUpdate(target, content, () => {
    computePosition(target, content, {
      middleware: options?.middleware ?? [shift({ crossAxis: true })],
    }).then(({ x, y }) => {
      Object.assign(content.style, {
        left: `${x}px`,
        top: `${y}px`,
      });
    });
  });
```

packages/blocks/src/_common/components/menu/menu.ts
```diff
@@ -693,6 +692,26 @@ export const positionToVRect = (x: number, y: number): VirtualElement => {
 export const eventToVRect = (e: MouseEvent): VirtualElement => {
   return positionToVRect(e.x, e.y);
 };
+
+const autoUpdate = (target: HTMLElement, updateFunction: () => void) => {
```
zzj3720, line 696: It is recommended to use autoUpdate in @floating-ui/dom

zzj3720 (changes requested):

PR #5460, opened 2023-11-26
fix(page): autoscroll during native selection

### LINKED ISSUE #4988: Content flickers during native selection across image
Image

AyushAgrawal-A2:
closes #4988 

Bug happens only when selection and page scroll happens together.
`rich-text` has auto scrolling which tries to keep the entire selection in view, hence preventing selected text to scroll out of view.

Changes:
- autoscroll only when vrange is updated in `sync` mode.
- added slot event on v-range-provider for setVRange to trigger `sync`.
- native selection events will update vrange in `async` mode i.e. `sync = false`

### DIFF
--- a/packages/virgo/src/utils/transform-input.ts
+++ b/packages/virgo/src/utils/transform-input.ts
@@ -10,35 +10,26 @@ function handleInsertText<TextAttributes extends BaseTextAttributes>(
 ) {
   if (!data) return;
   editor.insertText(vRange, data, attributes);
-  editor.setVRange(
-    {
-      index: vRange.index + data.length,
-      length: 0,
-    },
-    false
-  );
+  editor.setVRange({
+    index: vRange.index + data.length,
+    length: 0,
+  });
 }
 
 function handleInsertParagraph(vRange: VRange, editor: VEditor) {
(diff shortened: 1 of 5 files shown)


packages/virgo/src/utils/transform-input.ts
```diff
@@ -10,35 +10,26 @@ function handleInsertText<TextAttributes extends BaseTextAttributes>(
 ) {
   if (!data) return;
   editor.insertText(vRange, data, attributes);
-  editor.setVRange(
-    {
-      index: vRange.index + data.length,
-      length: 0,
-    },
-    false
```
Flrande, line 18: Why we use true here?

AyushAgrawal-A2: To trigger vertical / horizontal scroll on text input.
e.g. 
- on adding text / paragraphs,  page should scroll vertically.
- inside embedded code block to have horizontal scroll

earlier these scrolls was triggered indirectly by selection change event which is async.
text will be added to browser -> selection change event -> scroll

now scroll is being triggered on only on sync vrange changes.

Flrande (approved):

PR #5470, opened 2023-11-27
fix(store): block should add default value to props if not exist

Saul-Mirone:


doodlewind (approved):
We should try to revert #5437 after landing this @regischen

Flrande (approved):

regischen:
> We should try to revert #5437 after landing this @regischen

Alright, I will remove these.

PR #5478, opened 2023-11-27
fix(page): disable dnd when readonly

lawvs:
https://github.com/toeverything/blocksuite/assets/18554747/b4628a31-6b9f-45b5-96c9-bd89a03b7eea


doodlewind (commented):
The current error thrown looks as expected, since the store is frozen. And I don't think it's best practice to spread the concept of `readonly` in all widgets, otherwise there will be too much logic to adapt. Having better pluggable widget API (so that they can be unmounted during `readonly` state) should be the better way to go.

PR #5495, opened 2023-11-28
fix(edgeless): copy-as-png not work for nested frame

### LINKED ISSUE #5379: Unexpected behavior when framing other frames
### What happened?

After a third frame (frame 3) around two other frames (1 and 2), when you drag 3, the objects inside 1, and 2 move, but not the frame 

- I added 4 blocks
- Added a frame (1) around two and another frame (2) around two others 
- I added a frame (3) around the two frames 
- When i moved frame 3, the four blocks moved with it, but not frames 1 and 2
This is not what I would expect - Anything including frames inside a frame should be able to be dragged as one group. 

### Distribution version

Web (insider.affine.pro)

Chrome

### Relevant log output

_No response_

### Anything else?

_No response_

### LINKED ISSUE #5485: Copy as png broken on nested frame
I'm not sure if this is duplicate, if so please close

https://github.com/toeverything/blocksuite/assets/13579374/b9e84320-2322-4e2c-9817-ab3b98538f39

regischen:
close #5485 
close #5379

### DIFF
--- a/packages/blocks/src/page-block/edgeless/frame-manager.ts
+++ b/packages/blocks/src/page-block/edgeless/frame-manager.ts
@@ -24,14 +24,14 @@ const MIN_FRAME_WIDTH = 800;
 const MIN_FRAME_HEIGHT = 640;
 const FRAME_PADDING = 40;
 
-export const removeContainedFrames = (frames: FrameBlockModel[]) => {
+export function removeContainedFrames(frames: FrameBlockModel[]) {
   return frames.filter(frame => {
     const bound = Bound.deserialize(frame.xywh);
     return frames.some(
       f => f.id !== frame.id && Bound.deserialize(f.xywh).contains(bound)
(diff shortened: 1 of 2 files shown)


packages/blocks/src/page-block/edgeless/frame-manager.ts
```diff
@@ -23,6 +23,16 @@ import { isFrameBlock } from './utils/query.js';
 const MIN_FRAME_WIDTH = 800;
 const MIN_FRAME_HEIGHT = 640;
 const FRAME_PADDING = 40;
+
+export const removeContainedFrames = (frames: FrameBlockModel[]) => {
```
doodlewind, line 27: Please use `export function` by default.

regischen: Get it!

PR #5506, opened 2023-11-29
fix(page): format bar flickering

### LINKED ISSUE #5144: The position of the format bar sometimes flickers when formatting text
https://github.com/toeverything/blocksuite/assets/18554747/f014aef5-f0fa-4771-8f4b-7f08c3d47cb0

ddeonseo:
closes #5144 

This PR addresses and resolves the flickering issue observed in the positioning of the format bar. 
The problem was identified in the original code, where range.getClientRects() was being called on each invocation, leading to potential conflicts between the value at the time of the call and subsequent changes.

Changes:
Replaced dynamic range.getClientRects() calls with a constant clientRects to prevent conflicts.
Updated visualElement to use the precomputed clientRects value.
Verified and tested the solution, confirming the elimination of format bar flickering.

image

https://github.com/toeverything/blocksuite/assets/114056339/d5acfcc0-ebcd-4d41-8c7b-d7c2ab8dd35a


doodlewind (commented):
Thanks for your contribution! Could you sign our [CLA](https://github.com/toeverything/blocksuite/edit/master/.github/CLA.md) for the following review process?

ddeonseo:
Could someone please review the changes? Your feedback is highly appreciated.

doodlewind (commented):
Thanks for the fix! I think the only blocker here should be the remaining failed test case. Could you take a look at it and try fixing it locally? Thanks!

doodlewind:
@ddeonseo I think you can checkout the [CI log](https://github.com/toeverything/blocksuite/actions/runs/7027428452/job/19164300442?pr=5506) and workaround the test locally (we have the guide about how to run the tests [here](https://github.com/toeverything/blocksuite/blob/master/BUILDING.md#testing)). The only one failed case is `should format bar follow scroll`. cc @Flrande

ddeonseo:
Before I modified code, As i scrolled page, format bar's position is strange. Because I declared variable rangeRect. So rangeRect is not updated when we scroll. And I try it to add wheel event listener, to follow the scroll. But I failed the same test...
I'm facing difficulties understanding the reason behind the test failure, and I've reached a point where I'm unable to make further adjustments to the code. Could you please assist me in identifying what might be going wrong? Any insights or guidance would be greatly appreciated.

https://github.com/toeverything/blocksuite/assets/114056339/cfe09495-6881-4871-a6a2-35d73a8c06ce

image

PR #5531, opened 2023-11-30
refactor: remove legacy clipboard

Saul-Mirone:
\cc: @fourdim @zqran @doodlewind


packages/store/src/transformer/job.ts
```diff
@@ -136,11 +136,15 @@ export class Job {
...
       assets: this._assetsManager,
     });
-    const children = await Promise.all(
-      model.children.map(child => {
-        return this._blockToSnapshot(child);
-      })
-    );
+
+    const children =
+      model.role === 'root'
```
nawbc, line 141: What does this mean? refactoring?  It will make an empty export.

fourdim: Nice catch. It's a bug. And i is already fixed in thttps://github.com/toeverything/blocksuite/pull/5692

PR #5533, opened 2023-11-30
Feat presentation in readonly

regischen:


packages/blocks/src/page-block/edgeless/components/toolbar/edgeless-toolbar.ts
```diff
@@ -543,9 +547,9 @@ export class EdgelessToolbar extends WithDisposable(LitElement) {
   }
 
   override render() {
-    if (this.edgeless.page.readonly) return nothing;
-
     const { type } = this.edgelessTool;
+    if (this.edgeless.page.readonly && type !== 'frameNavigator')
```
doodlewind, line 551: Don't use this style, please stick to

`if (foo) return;`

or 

```
if (someLongCondition) {
  return;
}
```

packages/blocks/src/page-block/edgeless/services/tools-manager.ts
```diff
@@ -441,14 +441,16 @@ export class EdgelessToolsManager {
       editing: false,
     }
   ) => {
-    if (this.page.readonly && edgelessTool.type !== 'pan') return;
+    const { type } = edgelessTool;
+    if (this.page.readonly && type !== 'pan' && type !== 'frameNavigator')
```
doodlewind, line 445: The same `if` rule applies here

packages/blocks/src/page-block/edgeless/components/toolbar/edgeless-toolbar.ts
regischen: ok

doodlewind (approved):

PR #5546, opened 2023-12-01
Fix/edgeless copy frame and image

regischen:
This pr #5519 is implemented in `_legacy/edgeless-clipborad`. Now we're removing the _legacy code. So I moved the logic into the new clipboard. Meantime I completed the missing code for frame copy and add some tests.  \cc @donteatfriedrice

### DIFF
--- a/tests/edgeless/copy-as-png.spec.ts
+++ b/tests/edgeless/copy-as-png.spec.ts
@@ -1,0 +1,67 @@
+import {
+  createShapeElement,
+  edgelessCommonSetup,
+  Shape,
+  triggerComponentToolbarAction,
+} from 'utils/actions/edgeless.js';
+import {
+  pasteByKeyboard,
+  selectAllByKeyboard,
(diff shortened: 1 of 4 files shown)


tests/edgeless/copy-as-png.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+import {
+  pasteByKeyboard,
+  selectAllByKeyboard,
+} from 'utils/actions/keyboard.js';
+import { waitNextFrame } from 'utils/actions/misc.js';
+import { assertBlockCount } from 'utils/asserts.js';
+
+import { test } from '../utils/playwright.js';
+
+test.describe.only('copy as png', () => {
```
doodlewind, line 16: Remove `only`

doodlewind (commented):
I think these newly added tests are better to use the new test runner based on #5536 (require help from @doouding if necessary) in new PRs, which can be significantly faster.

PR #5565, opened 2023-12-04
feat(edgeless): connector elbowed indicator

lzlme:
Part of #4978 
redesigned connector for elbowed style


doodlewind:
Closing since this implementation doesn't reach smoke-test level usability for landing in nightly master. According to recent plan, we'll lower the priority of shipping edgeless features, focusing on bug fixes and refactors.

PR #5567, opened 2023-12-04
test: add test for layer manager

doouding:


packages/blocks/src/_common/utils/event.ts
```diff
@@ -126,3 +126,9 @@ export function delayCallback(callback: () => void, delay: number = 0) {
 
   return () => clearTimeout(timeoutId);
 }
+
+export function requestFrame(callback: () => void, element?: HTMLElement) {
+  return requestAnimationFrame(() => {
+    if (!element || element.isConnected) callback();
```
doodlewind, line 132: So if the element is not connected, the callback will never run?

doodlewind (approved):
Good examples to follow!

PR #5570, opened 2023-12-05
refactor: `@blocksuite/editor` -> `@blocksuite/presets`

doodlewind:
This is intended to expose more prebuilt *editors* and *auxiliary components* under the concept of *preset*. The auxiliary components will be defined as *fragments*.

* The `EditorPreset`s include `DocEditor`, `EdgelessEditor` and more.
* The `FragmentPreset`s include AI copilot panel and TOC panels, which shares the block tree under the same `page` instance, but may not share the same lifecycle of the editor component.

Migration steps:

1. Perform the package renaming with no API breaking changes in this PR.
2. Bump AFFiNE.
3. Expose the new entities and APIs in PR afterwards, with `EditorContainer` kept for backwards compatibility.


packages/blocks/src/_presets/index.ts
```diff
@@ -190,7 +190,7 @@ export const PagePreset: BlockSpec[] = [
       },
     },
   },
-  embedGithubBlockSpec,
+  EmbedGithubBlockSpec,
 ];
 
 export const EdgelessPreset: BlockSpec[] = [
```
Saul-Mirone, line 196: We may also want to rename `_presets` in blocks to `_blocks`

doodlewind: How about make it `_specs`?

PR #5572, opened 2023-12-05
feat(edgeless): expose `currentTool` in surface service

regischen:


doodlewind (changes requested):
I think your should expose `currentTool` instead.

doodlewind (approved):

Issue #5579, opened 2023-12-05
The selection can not go through the iframe


lawvs:
Related to https://github.com/toeverything/blocksuite/issues/4523 https://github.com/toeverything/blocksuite/pull/5243 https://github.com/toeverything/blocksuite/pull/5475

https://github.com/toeverything/blocksuite/assets/18554747/1e4ad20e-994f-421a-97e0-325078b61c62

PR #5580, opened 2023-12-05
fix(page): selection should be able to go through iframe

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/5579 by adding an overlay mask when the mouse is clicked down.

### DIFF
--- a/packages/blocks/src/attachment-block/attachment-block.ts
+++ b/packages/blocks/src/attachment-block/attachment-block.ts
@@ -52,6 +52,9 @@ export class AttachmentBlockComponent extends BlockElement<AttachmentBlockModel>
   @state()
   private _blobUrl?: string;
 
+  @state()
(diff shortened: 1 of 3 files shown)


packages/blocks/src/attachment-block/attachment-block.ts
```diff
@@ -101,6 +104,24 @@ export class AttachmentBlockComponent extends BlockElement<AttachmentBlockModel>
...
+      })
+    );
+    this.disposables.add(
+      this.std.event.add('pointerUp', e => {
+        const event = e.get('pointerState');
+        // 0: Main button pressed, usually the left button or the un-initialized state
+        if (event.button !== 0) return;
+        this._pointerPressed = false;
+      })
+    );
```
lawvs, line 124: I think this pattern is worth extracting to the `std` in the future.

PR #5589, opened 2023-12-06
fix(edgeless): connector in group copy bug

regischen:


doodlewind (approved):
I think the `EdgelessClipboard` should be migrated to new clipboard afterwards.

PR #5594, opened 2023-12-06
fix(edgeless): group element snapshot

regischen:


packages/blocks/src/surface-block/surface-transformer.ts
```diff
@@ -17,6 +18,11 @@ export class SurfaceBlockTransformer extends BaseBlockTransformer<SurfaceBlockPr
         [SURFACE_TEXT_UNIQ_IDENTIFIER]: true,
         delta: value.toDelta(),
       };
+    } else if (value instanceof Workspace.Y.Map) {
+      return {
+        [SURFACE_YMAP_UNIQ_IDENTIFIER]: true,
```
doodlewind, line 23: Which kind of data requires `toJSON` as ymap for now? I think there should be workaround avoid operating on `YMap`. cc @Saul-Mirone

Saul-Mirone: Seems the structure is `YMap<boolean>`, it's ok to use `toJSON`. I suggest to leave a comment here to explain the structure. Or use `YMap.entreis()` to iterate it.

doodlewind (approved):

PR #5599, opened 2023-12-06
fix: paste should replace selected text

### LINKED ISSUE #5592: Pasting new text doesn't replace originally selected text
Paste a new text will not replace the original selected text

https://github.com/toeverything/blocksuite/assets/110808979/10d46a5d-0772-4cc2-bee6-30718eb7ea3f

wfnuser:
It seems like this minor change can fix #5592 . I'm not so sure.

And I don't quite understand how paste is working. The last function I saw related to paste is:
```ts
        this._std.clipboard.paste(
          e,
          this._std.page,
          ctx.parentBlock.model.id,
          ctx.blockIndex ? ctx.blockIndex + 1 : undefined
        );
```

And the implementation looks like following:
```ts
      const json = this.readFromClipboard(data);
      const slice = await this._getSnapshotByPriority(
        type => json[type],
        page,
        parent,
        index
      );
      assertExists(slice);
      return slice;
```

How return slice makes the paste done?  I didn't get it. It would be great if someone help to explain.

### DIFF
--- a/packages/blocks/src/page-block/clipboard/index.ts
+++ b/packages/blocks/src/page-block/clipboard/index.ts
@@ -119,6 +119,8 @@ export class ClipboardController implements ReactiveController {
 
     this._std.command
       .pipe()
+      .withRoot()
+      .try(cmd => [cmd.getTextSelection().deleteText()])
       .try(cmd => [
         cmd.getTextSelection().inline<'currentSelectionPath'>((ctx, next) => {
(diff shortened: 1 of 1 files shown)


doodlewind (commented):
The issue should be related to certain cursor computation, so this doesn't look like the correct fix. See https://github.com/toeverything/blocksuite/issues/5592#issuecomment-1846491697

PR #5600, opened 2023-12-06
feat(edgeless): support adding edgeless templates

doouding:


doouding:
The PR is ready for review, but don't merge it yet.

packages/blocks/src/surface-block/service/template-middlewares.ts
```diff
@@ -0,0 +1,116 @@
+import { assertType } from '@blocksuite/global/utils';
+import { type SnapshotReturn } from '@blocksuite/store';
+
+import { generateElementId } from '../index.js';
+import type { SlotBlockPayload, TemplateJob } from './template.js';
+
+export const replaceIdMiddleware = (job: TemplateJob) => {
```
Saul-Mirone, line 7: naming

packages/store/src/transformer/job.ts
```diff
@@ -213,6 +213,25 @@ export class Job {
     return model;
   }
 
+  async snapshotToModelData(snapshot: BlockSnapshot) {
```
Saul-Mirone, line 216: Please keep using the arrow method like other methods.

doodlewind: Most places use default method by default. But aligning to the nearby implementation surely makes more sense.

packages/blocks/src/page-block/edgeless/components/toolbar/template/builtin-templates.ts
```diff
@@ -0,0 +1,61 @@
...
+  asserts?: Record<string, string>;
+  preview: string;
+};
+
+export type TemplateCategory = {
+  name: string;
+  templates: Template[];
+};
+
+interface ITemplateManager {
```
doodlewind, line 15: I think we are deprecating the `ISomething` naming convention.

doodlewind (approved):
Thanks for your hard work!

PR #5613, opened 2023-12-07
feat(edgeless): support consistent dragging behavior

### LINKED ISSUE #5561: Support consistent dragging behavior in edgeless mode
This includes following improvements to the drag handle and drag indicator:

- A. When dragging selected blocks inside the note to empty area of the edgeless canvas, dropping there would put these blocks into a new note.
- B. When dragging selected note A onto another note B, there should be the drag indicator, and dropping A should merge A into B.
- C. Dragging surface top level image over note should also show the indicator, dropping it should also merge the image into note. (This would also apply to top level bookmark afterwards.)

Case A:
image

Case B:
image

Case C: Same as B but apply to surface top level images.

We can ship this in multiple PRs.

AyushAgrawal-A2:
closes #5561 

Changes:
In edgeless mode,
- dragging edgeless-note onto another edgeless-note merge the notes.
- dragging multiple blocks from edgeless-note onto whiteboard move the blocks into a new edgeless-note.
- dragging single image/bookmark/embed-* block from edgeless-note onto whiteboard creates a top-level edgeless block.
- dragging top-level edgeless image/bookmark/embed-* block onto edgeless note move the block inside the edgeless note.


Saul-Mirone:
We may also need to support the `embed-*` blocks to drag between surface block and note block. You can view it on `/starter/?init=embed` to see an example.

doodlewind (commented):
Seems that we are unable to drag the image (inside the note) out via drag handle?

https://github.com/toeverything/blocksuite/assets/7312949/0ee35d3d-fffe-4c7b-b534-146d477b24c0

AyushAgrawal-A2:
> Seems that we are unable to drag the image (inside the note) out via drag handle?
> 
>  Screen.Recording.2023-12-11.at.11.30.58.mov

@doodlewind 
resolved image drop



> We may also need to support the `embed-*` blocks to drag between surface block and note block. You can view it on `/starter/?init=embed` to see an example.

should I complete `embed-*` blocks as well in this PR or should I wait for edgeless bookmark and do them together?

Saul-Mirone:
> > Seems that we are unable to drag the image (inside the note) out via drag handle?
> > Screen.Recording.2023-12-11.at.11.30.58.mov
> 
> @doodlewind resolved image drop
> 
> > We may also need to support the `embed-*` blocks to drag between surface block and note block. You can view it on `/starter/?init=embed` to see an example.
> 
> should I complete `embed-*` blocks as well in this PR or should I wait for edgeless bookmark and do them together?

Oh, I see. We may better to wait for edgeless bookmark.

Saul-Mirone:
We should add some integration tests for this feature to avoid future regressions.

hwangdev97 (approved):

doodlewind (commented):
May require @Flrande to help with bookmark-related merge conflicts?

doodlewind (approved):
Awesome work, we should definitely land this!

Two minor issues: 

1. After dragging out an image, you can undo this operation, but unable to redo it back:

https://github.com/toeverything/blocksuite/assets/7312949/114ecc9f-eb4e-40b1-a4d9-16b3c152dba3

This usually happens due to unexpected side effect (model mutation) that comes with the undo, which resets the history stack.

2. After dragging an image inside the note, the first undo/redo pair works fine. But from this moment, it requires 2 undo operations to make actual undo effect:

https://github.com/toeverything/blocksuite/assets/7312949/b401dae4-5461-43d5-8625-07807f5e3dd1

This usually happens when `page.captureSync()` is called during a series of model operations, which will add one undo step.

By default, the `UndoManger` merges all operations from the same user during about 500ms (even if these operations are using multiple `page.transact` calls) into one undo step. Using `page.captureSync()` is the way to add more undo steps manually. For certain cases (e.g., batch adding images), adding more undo steps allows user to undo each operation one after another.

PR #5616, opened 2023-12-08
fix: update reference id in group and connector in replaceIdMiddleware

regischen:


packages/store/src/transformer/job.ts
```diff
@@ -196,8 +196,18 @@ export class Job {
       index
     );
 
-    for (const [index, child] of children.entries()) {
-      await this._snapshotToBlock(child, page, id, index);
+    // surface-block needs to be initialized at last.
+    // because the reference id in group and connector need to be updated.
+    if (snapshot.flavour === 'affine:page') {
```
doodlewind, line 201: We should avoid bringing in special treatment about `affine:xxx` in store logic. Other workarounds? cc @Saul-Mirone

Saul-Mirone (approved):
LGTM

PR #5617, opened 2023-12-08
fix: database popup position fix while resizing window

gauravruhela07:


gauravruhela07:
Hi @doodlewind, as per the PR the pop-up opens in the center(40% from the top)  of the screen and it resizes well while resizing screen. 

However, the position of the pop-up need to be at the point where the user clicks. I have fixed this as well in my latest commit. Please review and let me know if you have any more suggestion.

packages/blocks/src/_common/components/menu/menu.ts
```diff
@@ -726,6 +726,23 @@ export const createPopup = (
       options?.onClose?.();
     }
   };
+
+  // logic to update pop-up position during resizing of screen
+  let currentScreenWidth = window.innerWidth;
+  window.addEventListener('resize', function () {
```
zzj3720, line 732: Unsubscribe logic needs to be handled

zzj3720: It is recommended to use autoUpdate in @floating-ui/dom

doodlewind:
Have to close this in favor of #5457, thanks for your contribution anyway!

PR #5641, opened 2023-12-11
refactor(edgeless): tear down edgeless element mixin

regischen:
Our current project involves rendering blocks onto a canvas, with an emphasis on scalability. These blocks must integrate specific methods and properties, especially required for seamless interaction with existing canvas elements. To address these requirements, we have formulated the following design strategy.

So if blocks need to be rendered on canvas, it must support three fields which are `xywh`, `index`, `rotate`.

The design ensures 
1. methods and properties such as `elementBound`, `hitTest` etc, have default implementations, yet allowing individual models the flexibility to easily override them as needed. 
2. acilitate easy expansion or addition of new models within the edgeless.

Usage:
```ts
const blockModelEdgeless = BlockEdgelessMixin(
  class extends BaseBlockModel<Props> {
     // model definitions here
  }
);
export blockModel extends blockModelEdgeless {
  // overrides here
}
```
cc\ @doodlewind @Saul-Mirone @doouding

### DIFF
--- a/packages/blocks/src/frame-block/frame-model.ts
+++ b/packages/blocks/src/frame-block/frame-model.ts
@@ -5,18 +5,8 @@ import { BaseBlockModel, defineBlockSchema } from '@blocksuite/store';
 import { getBlockElementByPath } from '../_common/utils/query.js';
 import { FRAME_BATCH } from '../surface-block/batch.js';
 import type { EdgelessBlockType } from '../surface-block/edgeless-types.js';
-import type {
-  HitTestOptions,
-  IEdgelessElement,
-} from '../surface-block/elements/edgeless-element.js';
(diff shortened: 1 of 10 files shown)


packages/blocks/src/frame-block/frame-model.ts
```diff
@@ -43,46 +33,20 @@ type Props = {
...
-  }
-
-  get batch() {
-    return FRAME_BATCH;
-  }
-
-  get rotate() {
-    return 0;
+const FrameBlockModelEdgeless = BlockEdgelessMixin(
+  class extends BaseBlockModel<Props> {
```
Saul-Mirone, line 40: This seems not right, I'll modify this.

Saul-Mirone (changes requested):

doodlewind:
Closing in favor of #5647

Issue #5638, opened 2023-12-11
Presentation mode disruption: sidebar color change


Yuhang1015:
version: 0.10.4 beta

Before the presentation, the mouse selection mode was "select". After the presentation, the sidebar changed from white to black. It took a few seconds to recover.

https://github.com/toeverything/blocksuite/assets/142964904/930a2c48-6c10-4519-aa91-03b63ce42e6f

regischen:
seems not reproduce in latest beta

PR #5643, opened 2023-12-11
fix(edgeless): default mode after exiting presentation

regischen:


doodlewind (commented):
Can this close #5638?

regischen:
> Can this close #5638?

Only resolve part of it. I feel the sider bar problem is related to AFFINE.

doodlewind (approved):

PR #5651, opened 2023-12-11
fix: update format bar styles

lawvs:
This pull request includes the following changes:

- Align the paragraph panel of the format bar.
- Add some offset to the hover panels on the format bar.
- Increase the hover delay for the hover panels.

https://github.com/toeverything/blocksuite/assets/18554747/8d8e1d4e-7f8a-4b0d-96fc-7c76c9a41954


doodlewind (approved):
Thanks for polishing details!

PR #5672, opened 2023-12-12
fix(presets): render frame preview incorrectly when switch between pages

donteatfriedrice:


doodlewind (approved):

packages/presets/src/fragments/frame-panel/card/frame-card.ts
```diff
@@ -211,7 +211,7 @@ export class FrameCard extends WithDisposable(LitElement) {
     });
   }
 
-  private _renderDraggingCardNumber() {
+  private DraggingCardNumber() {
```
doodlewind, line 214: Better use `_DraggingCardNumber` for private templates 🌝

donteatfriedrice: Get it.

PR #5677, opened 2023-12-13
refactor:  optimize the implementation of isNativeTextInVText

Yukiniro:
Current implementation will cause warning that is `The "??" operator here will always return the left operand`.


doodlewind (approved):

doodlewind:
LGTM, thanks for the improvement!

PR #5698, opened 2023-12-13
fix(edgeless): fix blur toolbar

Yukiniro:
`transform` an element with odd sizes may cause blur. So, I updated the odd width to be an even number.
After fix:

https://github.com/toeverything/blocksuite/assets/21242087/757a9ca5-f21d-419d-8bb5-c34bb9d42349


doodlewind (approved):
Thanks for polishing this!

PR #5707, opened 2023-12-14
refactor: remove redundant `blocks` util exports

doodlewind:
Exported method count: 114 -> 65.


doodlewind (commented):
Require @regischen to fix the regression migrating `saveViewportToSession` as command.

PR #5720, opened 2023-12-14
style: add some runtime related rules

AyushAgrawal-A2:
- @typescript-eslint/no-unused-vars
- @typescript-eslint/no-floating-promises
- @typescript-eslint/no-non-null-asserted-nullish-coalescing
- @typescript-eslint/no-misused-promises

### DIFF
--- a/packages/blocks/src/_common/adapters/image.ts
+++ b/packages/blocks/src/_common/adapters/image.ts
@@ -72,7 +72,7 @@ export class ImageAdapter extends BaseAdapter<Image> {
     for (const item of payload.file) {
       const blobId = await sha(await item.arrayBuffer());
       payload.assets?.getAssets().set(blobId, item);
-      payload.assets?.writeToBlob(blobId).catch(console.error);
+      await payload.assets?.writeToBlob(blobId);
       content.push({
         type: 'block',
         flavour: 'affine:image',
--- a/packages/blocks/src/_common/adapters/image.ts
+++ b/packages/blocks/src/_common/adapters/image.ts
@@ -72,7 +72,7 @@ export class ImageAdapter extends BaseAdapter<Image> {
     for (const item of payload.file) {
       const blobId = await sha(await item.arrayBuffer());
       payload.assets?.getAssets().set(blobId, item);
-      payload.assets?.writeToBlob(blobId);
+      payload.assets?.writeToBlob(blobId).catch(console.error);
       content.push({
         type: 'block',
         flavour: 'affine:image',
--- a/packages/blocks/src/_common/components/hover/when-hover.ts
+++ b/packages/blocks/src/_common/components/hover/when-hover.ts
@@ -102,19 +102,35 @@ export const whenHover = (
     const alreadyHover = element.matches(':hover');
     if (alreadyHover && !abortController.signal.aborted) {
       // When the element is already hovered, we need to trigger the callback manually
-      onHoverChange(new MouseEvent('mouseover'));
+      onHoverChange(new MouseEvent('mouseover')).catch(console.error);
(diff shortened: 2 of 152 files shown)


packages/blocks/src/_common/components/hover/when-hover.ts
```diff
@@ -102,19 +102,35 @@ export const whenHover = (
...
   const removeHoverListener = (element?: Element) => {
     if (!element) return;
-    element.removeEventListener('mouseover', onHoverChange);
-    element.removeEventListener('mouseleave', onHoverChange);
+    element.removeEventListener('mouseover', (e: Event) => {
+      onHoverChange(e).catch(console.error);
+    });
+    element.removeEventListener('mouseleave', (e: Event) => {
+      onHoverChange(e).catch(console.error);
+    });
```
fourdim, line 133: Incorrect usage.
https://github.com/toeverything/blocksuite/pull/5047#discussion_r1363871355
See https://github.com/toeverything/blocksuite/pull/5054 for more.

packages/store/src/providers/async-call-rpc.ts
```diff
@@ -148,13 +148,16 @@ export const createAsyncCallRPCProviderCreator = (
...
         event.added.forEach(doc => docMap.set(doc.guid, doc));
         event.added.forEach(doc => {
-          rpc.queryDocState(doc.guid).then(update => {
-            if (!update) {
-              return;
-            }
-            Y.applyUpdate(doc, update, channel);
-            doc.emit('load', []);
-          });
+          rpc
```
Saul-Mirone, line 151: @pengx17 Please help us to review the provider code

Saul-Mirone (approved):
The changes in `store` and `block-std` look good to me.

packages/blocks/src/_common/adapters/image.ts
```diff
@@ -72,7 +72,7 @@ export class ImageAdapter extends BaseAdapter<Image> {
     for (const item of payload.file) {
       const blobId = await sha(await item.arrayBuffer());
       payload.assets?.getAssets().set(blobId, item);
-      payload.assets?.writeToBlob(blobId);
+      payload.assets?.writeToBlob(blobId).catch(console.error);
```
pengx17, line 75: Do we need to await for this promise and make the error pops up?

doodlewind (approved):
Thanks for the feat!

tests/utils/actions/misc.ts
```diff
@@ -119,20 +119,18 @@ async function initEmptyEditor({
...
           if (!page) {
             throw new Error(`Failed to get page ${pageId}`);
           }
-          initPage(page);
+          await initPage(page);
         });
       } else {
         const page = workspace.createPage({ id: 'page:home' });
-        initPage(page);
+        await initPage(page);
```
pengx17, line 133: what is the rule of when to use async/await and attaching .catch(console.error)?

pengx17 (commented):
I am not sure but it seems to me that it is not good practice to use console.error to catch all of these floating promises.

tests/utils/actions/misc.ts
doodlewind: I have removed the console.error I found, it should not be the default behavior.

AyushAgrawal-A2: The eslint rule is that we need to avoid floating promises...
1. either catch them or use async/await or void the promise
2. if a callback function is returning a promise, the calling function should be expecting a promise returning function

In this pr, I have assumed that if an async operation is taking place without await, it was intended to be done asynchronously in background so I have just caught the error with `console.error`

PR #5724, opened 2023-12-14
feat(edgeless): add more templates

doouding:


doouding:
Adding the `assert { type: 'json' }` should be able to pass the node test but may require the downstream to add configuration to get the app successfully built. So I switch to `.ts` file

doodlewind (approved):

PR #5725, opened 2023-12-14
fix(edgeless): fix incorrect display of text-align and space interaction

### LINKED ISSUE #5723: Wrong display in canvas text editor
If there are spaces at the end of the text and it is right aligned, entering edit mode will display incorrectly.

https://github.com/toeverything/blocksuite/assets/21242087/42609e58-8283-4fa5-9166-cf2c072a9d5a

Yukiniro:
Closes #5723
Under the effect of text-align, ' ' and '\&nbsp;' have different display.
For example: https://codepen.io/Yukiniro/pen/bGzyYeb


packages/inline/src/components/v-text.ts
```diff
@@ -15,11 +15,18 @@ export class VText extends LitElement {
     'white-space': 'break-spaces',
   };
 
+  private parser = new DOMParser();
+
   override render() {
     // we need to avoid \n appearing before and after the span element, which will
     // cause the sync problem about the cursor position
+    const parsed = this.parser.parseFromString(
+      this.str.replace(/\s/g, '&nbsp;'),
```
doodlewind, line 24: Seems that this breaks the test cases that depends on the rendered result. To land this, corresponding test cases also needs be updated.

Yukiniro: Looks like yes, I will update this PR.

Flrande:
The reason isn't that; it's simply because text-align affects not only the bottom-most text but also some internal elements, resulting in the offset.

image

I've created another [PR](https://github.com/toeverything/blocksuite/pull/5752) to fix it, but still, thank you for your contribution :)

Yukiniro:
> The reason isn't that; it's simply because text-align affects not only the bottom-most text but also some internal elements, resulting in the offset.
> 
> image
> I've created another [PR](https://github.com/toeverything/blocksuite/pull/5752) to fix it, but still, thank you for your contribution :)

So that's it, I understand, thanks for your explanation. I will close this pr.

PR #5750, opened 2023-12-15
fix(edgeless): undo/redo on image dragging

### LINKED ISSUE #5722: Improve undo behavior for edgeless content dragging
See https://github.com/toeverything/blocksuite/pull/5613#pullrequestreview-1781889412

> Two minor issues: 
> 
> 1. After dragging out an image, you can undo this operation, but unable to redo it back:
> 
> https://github.com/toeverything/blocksuite/assets/7312949/114ecc9f-eb4e-40b1-a4d9-16b3c152dba3
> 
> This usually happens due to unexpected side effect (model mutation) that comes with the undo, which resets the history stack.
> 
> 2. After dragging an image inside the note, the first undo/redo pair works fine. But from this moment, it requires 2 undo operations to make actual undo effect:
> 
> https://github.com/toeverything/blocksuite/assets/7312949/b401dae4-5461-43d5-8625-07807f5e3dd1
> 
> This usually happens when `page.captureSync()` is called during a series of model operations, which will add one undo step.
> 
> By default, the `UndoManger` merges all operations from the same user during about 500ms (even if these operations are using multiple `page.transact` calls) into one undo step. Using `page.captureSync()` is the way to add more undo steps manually. For certain cases (e.g., batch adding images), adding more undo steps allows user to undo each operation one after another.

AyushAgrawal-A2:
closes #5722

### DIFF
--- a/packages/blocks/src/image-block/utils.ts
+++ b/packages/blocks/src/image-block/utils.ts
@@ -135,7 +135,7 @@ export async function uploadBlobForImage(
   setImageLoading(blockId, true);
   const storage = page.blob;
   let sourceId = '';
-  let imageBlock: BaseBlockModel | null;
+  let imageBlock: BaseBlockModel | null = null;
   try {
     imageBlock = page.getBlockById(blockId);
     if (!imageBlock) {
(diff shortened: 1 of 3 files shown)


doodlewind (approved):
Good catch!

Issue #5753, opened 2023-12-17
Pasting will raise error.


CharlieCho2017:
I use the quick start demo in my project. But when I paste some  text into the editor. It will raise error. The playground has the same error.

```
import '@blocksuite/presets/themes/affine.css';
import { createEmptyPage, DocEditor } from '@blocksuite/presets';
const page = createEmptyPage().init();
const editor = new DocEditor();
editor.page = page;
document.body.appendChild(editor);
```

# playground: https://try-blocksuite.vercel.app/starter/?init

# copied text:
Quick Start
The @blocksuite/presets package contains the prebuilt editors and opt-in additional UI components. Its nightly versions are released daily based on the master branch, which is also recommended for real world usage:

https://github.com/toeverything/blocksuite/assets/28870006/6d7bbc57-47b3-438c-853f-e4748932b5fc

Flrande:
Is your computer Windows? If so, it is likely because the clipboard did not filter out the Windows-specific '\r' character, which is a bug in the new clipboard codebase. We should add a test for this. cc @Saul-Mirone

CharlieCho2017:
> Is your computer Windows? If so, it is likely because the clipboard did not filter out the Windows-specific '\r' character, which is a bug in the new clipboard codebase. We should add a test for this. cc @Saul-Mirone 

Yes, my computer is WIndows. Thank you.

PR #5764, opened 2023-12-18
fix: replace `\r` for windows in clipboard

zqran:
close: #5753

https://github.com/toeverything/blocksuite/assets/15389209/a76a108a-7957-447c-9df2-e46befbef2d8

### DIFF
--- a/packages/store/src/reactive/text.ts
+++ b/packages/store/src/reactive/text.ts
@@ -20,6 +20,14 @@ export class Text {
     if (typeof input === 'string') {
       this._yText = new Y.Text(input);
     } else if (input instanceof Y.Text) {
+      const deltas = input.toDelta();
+      input.delete(0, input.length);
+      for (const delta of deltas) {
+        if (delta.insert) {
+          delta.insert = delta.insert.replaceAll('\r\n', '\n');
+        }
+      }
+      input.applyDelta(deltas);
       this._yText = input;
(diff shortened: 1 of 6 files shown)


packages/store/src/transformer/json.ts
```diff
@@ -41,7 +41,13 @@ export function fromJSON(value: unknown): unknown {
     }
     if (Reflect.has(value, TEXT_UNIQ_IDENTIFIER)) {
       const yText = new Y.Text();
-      yText.applyDelta(Reflect.get(value, 'delta'));
+      const deltas = Reflect.get(value, 'delta');
+      for (const delta of deltas) {
+        if (delta.insert) {
+          delta.insert = delta.insert.replaceAll('\r\n', '\n');
```
Saul-Mirone, line 47: I think we should implement this feature in text adapter(`store/reactive/text`) instead of in transformer.

packages/store/src/reactive/text.ts
```diff
@@ -20,6 +20,14 @@ export class Text {
     if (typeof input === 'string') {
       this._yText = new Y.Text(input);
```
Saul-Mirone, line 21: Should we also filter the normal string?

zqran: Done

fourdim (approved):

PR #5765, opened 2023-12-18
refactor: cleanup for multiple-editors

AyushAgrawal-A2:
Having global query functions, i.e. funtions having `document.querySelection`, prevents implementation of multiple-editors as the query result will always return element from first editor.

Changes:
- refactored utils and their usage, which directly / indirectly calls query on the document. `EditorHost` is used as the node for DOM queries instead.


doodlewind (approved):
Impressive work, thanks!

PR #5770, opened 2023-12-19
feat(edgeless): update presentation setting labels

donteatfriedrice:
The "Dark background" switch - when turned on, a black box is displayed - by default, it is not displayed. 
The "Hide toolbar" switch - when turned on, the toolbar is hidden - by default, it is displayed.


regischen (approved):

regischen:
It seems that "Dark background" is displayed by default.

doodlewind (approved):
LGTM

PR #5776, opened 2023-12-19
refactor: split fragments from doc-editor

### LINKED ISSUE #5145: affine-page-meta-data should be set as a configuration item
these should not be mandatory

https://github.com/toeverything/blocksuite/blob/f8b04fc94f24a44db581bec3883f2f785e36a704/packages/blocks/src/page-block/doc/doc-page-block.ts#L600C14-L600C14

### LINKED ISSUE #5757: Split fragments from `DocEditor`
For now, we have built the page title and meta tag panel inside the `DocEditor`. But with the new fragment mechanism landed, to ensure better reusability, we can further split `TitleFragment` and `PageMetaFragment` from the `DocEditor`: 

image

This could allowing us to quickly compose a `CommentEditor` using the same content component.

AyushAgrawal-A2:
closes #5757 
closes #5145


packages/blocks/src/page-block/widgets/drag-handle/drag-handle.ts
```diff
@@ -358,16 +353,16 @@ export class AffineDragHandleWidget extends WidgetElement<
     }
     dragPreview.classList.add('grabbing');
     dragPreview.style.display = 'block';
-    this.pageBlockElement.appendChild(dragPreview);
+    document.body.appendChild(dragPreview);
```
doodlewind, line 356: Is this something that needs to be refactored for supporting multiple editors?

packages/blocks/src/_common/embed-block-helper/embed-block-element.ts
```diff
@@ -81,7 +81,7 @@ export class EmbedBlockElement<
       const blockComponent = anchorComponent as this;
       const isInSurface = blockComponent.isInSurface;
       if (!isInSurface) {
-        this.host.selection.setGroup('block', [
+        this.host.selection.set([
```
Flrande, line 84: Why we don't use `setGroup` here?

packages/presets/src/editors/editor-container.ts
```diff
@@ -154,10 +163,10 @@ export class AffineEditorContainer
     return html`${keyed(
       this.model.id,
       this.mode === 'page'
-        ? html`<doc-editor
+        ? html`<affine-doc-editor
```
Flrande, line 166: I think `doc-editor` and `affine-doc-editor` need to be distinguished in a better way. What do you think? @doodlewind

Flrande: Maybe we can just load fragments in `editor-container` and no need to create `affine-doc-editor`?

packages/blocks/src/page-block/widgets/drag-handle/drag-handle.ts
AyushAgrawal-A2: not an issue for multiple editors as there can be only one dragPreview globally...
but I agree that adding elements on document.body should be avoided.... I left this by mistake while debugging incorrect dragPreview position due to container offset

packages/blocks/src/_common/embed-block-helper/embed-block-element.ts
AyushAgrawal-A2: we need to clear other selections to prevent side effects... while dragging only the dragging elements should be selected.
this was changed to solve a dragging bug... where existing text-selection was causing drag/drop to fail.

packages/presets/src/editors/editor-container.ts
AyushAgrawal-A2: I agree... "Hardest thing in programming is naming variables" :P

I just used the naming in the github issue:-
https://github.com/toeverything/blocksuite/issues/5757#issue-2045838265
image

packages/blocks/src/_common/embed-block-helper/embed-block-element.ts
Flrande: Oh, I suddenly realized that none of the selections belong to the 'block' group, only 'note' and 'edgeless'. Why was it written this way before?

image

AyushAgrawal-A2: oops... that was the reason for the bug then... corrected to `setGroup('note'`

doodlewind (approved):
Thanks for the refactoring!

PR #5779, opened 2023-12-20
chore: exclude from coverage

doouding:


doouding:
The `exclude` option does not work. Let's set it aside for now and consider moving the templates to the AFFiNE side in the next iteration so that it can be used offline. @doodlewind

doodlewind:
LGTM

PR #5791, opened 2023-12-20
perf(edgeless): optimize zoom in/out performance

doouding:
### Change
- delete unnecessary hit tests when the mouse moves around
    - remove the `edgeless-hover-rect` component
    - optimize note hover status detection
- apply transform on the entire layer
- apply block position through `left` and `top` position rather than `transform` property
- optimize `frame` block hitTest
    - add `GridManager` for the frame block
    - frame title hitTest

### Result 

Tests on Apple M2 Pro, 32GB, with CPU 4x slowdown, four templates inserted

**Before**

https://github.com/toeverything/blocksuite/assets/9301743/572e6c70-2e8b-4a6d-80e4-962f7b93b8fd

**After**

https://github.com/toeverything/blocksuite/assets/9301743/209ba8a7-4a49-49a4-bb17-6905b36373d8


doodlewind (approved):

doodlewind:
Better attach your test page snapshot zip for reproduction.

PR #5804, opened 2023-12-21
feat(edgeless): persist last editing state during session

### LINKED ISSUE #5211: Add session memory for attribute settings of whiteboard objects
**Goal**：
Minimize the wastage of users' energy on repetitive settings when creating new objects on the whiteboard.

**Current**：
- Not all whiteboard objects have remembered attribute settings, such as Connector, Note, and Frame, which require modification every time they are used.
- The duration for recording attribute settings is too long. Objects that support attribute setting memory will still retain their settings when the page is closed and reopened.

**Expected effect**:
- Set opening and closing a page as one session.
- Within a session, the attribute settings of the previous object will be recorded. When creating new objects of the same type, the previously set attribute settings will be applied. After closing the page, the attribute settings will be restored to their default values.

In the future, we will consider adding preferential attribute settings for various objects in the Settings.

regischen:
Close #5211 

https://github.com/toeverything/blocksuite/assets/58546692/a09e3936-c9eb-4416-8c02-e5351cfdd6f7


doodlewind (commented):
Could you elaborate what's the meaning of "prop inheritance"?

doodlewind (changes requested):
This looks overly designed. Having static fields about last editing states in surface service should be more appropriate.

doodlewind (changes requested):
1. Move the static definition and methods from `SurfaceService` to `EditSessionStorage` at `surface-block/managers/edit-session-storage.ts`, use it as `service.editSession.record()` / `service.editSession.apply()`. This should also involve removing `saveViewportToSessionCommand` / `getViewportFromSessionCommand` and other session-related code.
2. Remove the batch of hard-coded `undefined` with zod ([hint](https://chat.openai.com/share/d30f576f-c79d-41d5-959a-0b56ba2588c1)).

packages/blocks/src/surface-block/managers/edit-session-manager.ts
```diff
@@ -0,0 +1,317 @@
...
+  text: z.object({
+    color: z.string(),
+    fontFamily: CanvasTextFontFamilySchema,
+    textAlign: TextAlignSchema,
+    fontWeight: CanvasTextFontWeightSchema,
+    fontStyle: CanvasTextFontStyleSchema,
+    fontSize: z.number(),
+    hasMaxWidth: z.boolean(),
+  }),
+  ['affine:note']: z.object({
```
doodlewind, line 93: Prefer using `'affine:note'` instead of `['affine:note']` 🤔

packages/blocks/src/surface-block/managers/edit-session.ts
```diff
@@ -0,0 +1,318 @@
...
+    this._tryLoadLastProps();
+  }
+
+  private _tryLoadLastProps() {
+    const props = sessionStorage.getItem(SESSION_PROP_KEY);
+    if (!props) return;
+
+    try {
+      const _props = LastPropsSchema.parse(JSON.parse(props));
+      recursive(this._lastProps, _props);
```
doodlewind, line 199: It's a good step switching to zod. But with this typed schema written, I'm not sure if the `merge` library is appropriate to be used together with it. See [snippet](https://chat.openai.com/share/06ae83df-3b4c-4d28-b498-0299498027fe).

packages/blocks/src/surface-block/managers/edit-session-manager.ts
regischen: Okay

packages/blocks/src/surface-block/managers/edit-session.ts
```diff
@@ -0,0 +1,312 @@
...
+        borderRadius: z.number(),
+        borderSize: z.number(),
+        borderStyle: StorkeStyleSchema,
+        shadowType: NoteShadowsSchema,
+      }),
+    }),
+  }),
+});
+
+const defaultLastProps = {
```
regischen, line 107: we need to get the static keys of schema. So here we don't use .default().
.default() will add a wrapper causing not fetch the keys.

packages/blocks/src/surface-block/managers/edit-session.ts
```diff
@@ -280,9 +259,28 @@ export class EditSessionStorage {
   }
 }
 
-export function isLastPropType(
-  type: EdgelessElementType
-): type is keyof LastProps {
+function extractProps(
```
doodlewind, line 262: Please double check unnecessary exports. And utils not using `this` should better be defined as plain function rather than class methods.

doodlewind (approved):

packages/blocks/src/surface-block/managers/edit-session.ts
regischen: OK, Thanks for mentioning!

regischen:
> I appreciate your efforts on the recent projects, but I must express my serious concerns regarding the ongoing issues we've encountered in the review process. I've noticed a consistent pattern where our coding standards and practices are not being adequately met. This has led to increased review cycles and challenges in our workflow. It is critical for the team's success that these standards are consistently adhered to. I suggest we refocus on concrete bug fixes and product features with a clear scope, and I expect to see significant improvements in these areas. Please understand that continued challenges in meeting these expectations could impact your future with the team. I am open to any questions or discussions on how we can address these issues together.

Thank you for your feedback. I acknowledge the issues in the review process and understand the importance of adhering to our coding standards and practices. I am committed to making the necessary improvements and focusing on concrete bug fixes and product features.

PR #5807, opened 2023-12-21
fix(edgeless): top level block drag handle position

donteatfriedrice:
https://github.com/toeverything/blocksuite/assets/99816898/deccb39e-0ede-4655-be1e-4556b8598afe


doodlewind (approved):
Thanks!

PR #5815, opened 2023-12-21
fix: drag handle ui bugs

AyushAgrawal-A2:
Bugs fixed:
- drag handle showing after block is deleted
- drag handle showing over pop-up menu
- drag handle disappears on failed drop. need to deselect and select again to show drag handle.

added throttle to limit too many update

### DIFF
--- a/packages/blocks/src/page-block/edgeless/components/component-toolbar/component-toolbar.ts
+++ b/packages/blocks/src/page-block/edgeless/components/component-toolbar/component-toolbar.ts
@@ -383,7 +383,7 @@ export class EdgelessComponentToolbar extends WithDisposable(LitElement) {
     return html` <style>
         :host {
           position: absolute;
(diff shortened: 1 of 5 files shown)


doodlewind (approved):

packages/blocks/src/page-block/edgeless/components/component-toolbar/component-toolbar.ts
```diff
@@ -383,7 +383,7 @@ export class EdgelessComponentToolbar extends WithDisposable(LitElement) {
     return html` <style>
         :host {
           position: absolute;
-          z-index: 1;
+          z-index: 3;
```
doodlewind, line 386: There was a plan to make the z-index of widgets derived from CSS variables, we can do this in later refactoring.

doodlewind (approved):

PR #5820, opened 2023-12-22
perf(edgeless): update selected rect state per frame

regischen:


packages/blocks/src/page-block/edgeless/components/rects/edgeless-selected-rect.ts
```diff
@@ -621,7 +624,7 @@ export class EdgelessSelectedRect extends WithDisposable(LitElement) {
     this.slots.cursorUpdated.emit(cursor);
   };
 
-  private _updateSelectedRect() {
+  private _updateSelectedRect = withRequestConnectedFrame(() => {
```
doodlewind, line 627: Is this equivalent of plain `requestAnimationFrame`? If so, don't encapsulation opaque utils since I haven't seem the chance reusing it elsewhere.

doodlewind (changes requested):
Good to see the optimization, but please avoid high-level abstraction that we don't have agreement about its usability yet.

packages/blocks/src/page-block/edgeless/components/rects/edgeless-selected-rect.ts
regischen: It's not equivalent of `requestAnimationFrame`.
It ensures that the function will be executed at most once per frame, no matter how many times it is called.

image
image

I see some patterns here. I can replace them.

doodlewind (approved):

PR #5834, opened 2023-12-23
fix(edgeless): correct overlay position after #5791

AyushAgrawal-A2:
Due to the change in rendering style of edgeless block portal items in #5791, overlay position calculation needs to be updated.
Affected items:
- drag preview of top-level blocks in edgeless
- note slicer (on zoom in/out)
- more menu (on edgeless surface pan)

https://github.com/toeverything/blocksuite/assets/54364088/e1b75899-7e4a-401b-8963-c7ed95f25e3d


doodlewind (approved):
LGTM

PR #5839, opened 2023-12-24
chore: remove the legacy import page snapshot

Saul-Mirone:
The legacy snapshot has been deprecated for a lone time. We'll remove the API since we have transformers now and the transformer API is stable enough.

\cc: @pengx17 @joooye34 @doodlewind


doodlewind (approved):
LGTM, merry Christmas!

PR #5849, opened 2023-12-25
feat(presets): support display heading only in TOC

donteatfriedrice:
Add a note sortable option for TOC.
Default:
1. Hide preview block type icon. 
2. Only display headings.

Screenshot 2023-12-25 at 18 30 31

User can:
1. Set preview icons to visible by preview setting menu.
2. Active note sortable option. When note sortable option active, TOC panel will display all the blocks and support to be drag to reorder.

https://github.com/toeverything/blocksuite/assets/99816898/75016395-1f5f-4e64-b620-32177a875e9f


doodlewind (approved):
Great work!

PR #5850, opened 2023-12-25
refactor: move to single contenteditable

### LINKED ISSUE #1218: Unable to select text across blocks in Safari
https://user-images.githubusercontent.com/27926/219260686-ffc552ac-4b2a-43ed-8116-fdcfea90e885.mov

Flrande:
Close #1218


packages/lit/src/utils/range-binding.ts
```diff
@@ -0,0 +1,258 @@
...
+    if (!to || PathFinder.equals(from.path, to.path)) return;
+
+    const range = this.rangeManager.value;
+    if (!range) return;
+
+    const blocks = this.rangeManager.getSelectedBlockElementsByRange(range, {
+      mode: 'flat',
+    });
+
+    const start = blocks.at(0);
```
Saul-Mirone, line 130: Seems repeat with line 196

packages/lit/src/utils/range-manager.ts
```diff
@@ -177,185 +131,126 @@ export class RangeManager {
...
+      },
+      to:
+        startBlock === endBlock
+          ? null
+          : {
+              path: endBlock.path,
+              index: endInlineRange.index,
+              length: endInlineRange.length,
+            },
+      isReverse: reverse,
```
Saul-Mirone, line 215: Do we still need this property? I remember that this property was used to navigate with keyboard. Maybe we can try to remove it.

packages/lit/src/utils/range-binding.ts
Flrande: Do you mean them?

```ts
const start = blocks.at(0);
const end = blocks.at(-1);
if (!start || !end) return;

const startText = start.model.text;
const endText = end.model.text;
if (!startText || !endText) return;
```

packages/lit/src/utils/range-manager.ts
Flrande: Maybe we can use it to improve UX in remote-selection. 🤔

Saul-Mirone: You mean the position of the name flag?

Flrande: Yes

Flrande: And I want to rename `isReverse` to `reverse` 🤪

packages/lit/src/utils/range-binding.ts
Saul-Mirone: yep

Saul-Mirone (approved):

nawbc:
img_v3_027b_d05a63df-7bb1-415d-8f7e-1ea25aacaa9g

PR #5851, opened 2023-12-25
fix(edgeless): canvas text inconsistency between view and editing state

AyushAgrawal-A2:
[TOV-198](https://linear.app/affine-design/issue/TOV-198/fix-the-canvas-text-inconsistency-between-view-and-editing-state)

### DIFF
--- a/packages/blocks/src/page-block/edgeless/components/text/edgeless-text-editor.ts
+++ b/packages/blocks/src/page-block/edgeless/components/text/edgeless-text-editor.ts
@@ -2,7 +2,7 @@ import '../../../../_common/components/rich-text/rich-text.js';
 
 import { assertExists } from '@blocksuite/global/utils';
 import { ShadowlessElement, WithDisposable } from '@blocksuite/lit';
-import { css, html } from 'lit';
+import { css, html, nothing } from 'lit';
 import { customElement, property, query } from 'lit/decorators.js';
 import { styleMap } from 'lit/directives/style-map.js';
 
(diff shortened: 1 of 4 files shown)


packages/blocks/src/page-block/edgeless/components/text/edgeless-text-editor.ts
```diff
@@ -63,10 +56,10 @@ export class EdgelessTextEditor extends WithDisposable(ShadowlessElement) {
       overflow-wrap: normal !important;
     }
 
-    /* We cannot add styles directly from the top, as this would cause a shift in the inline elements inside. */
-    /* https://github.com/toeverything/blocksuite/issues/5723 */
```
Flrande, line 67: This will cause a bug regression.

packages/blocks/src/page-block/edgeless/components/text/edgeless-text-editor.ts
```diff
@@ -395,66 +341,49 @@ export class EdgelessTextEditor extends WithDisposable(ShadowlessElement) {
...
-    const paddingOffset = this.getPaddingOffset(textAlign);
+    const { hasMaxWidth, w } = this.element;
     const rect = getSelectedRect([this.element]);
-    const hasMaxWidth = this.element.hasMaxWidth;
-    const w = this.element.w;
-    const [x, y] = this.getVisualPosition(this.element);
-    const placeholder = this._renderPlaceholder();
-    const hasPlaceholder = placeholder !== nothing;
-
-    const transformOperation = [
```
Flrande, line 411: The consistency between edit and non-edit states during rotation now seems to rely on these transformation operations.

https://github.com/toeverything/blocksuite/assets/50035259/cfeec723-42a8-4456-b3f2-226eec0aa98d

Flrande (changes requested):

packages/blocks/src/page-block/edgeless/components/text/edgeless-text-editor.ts
AyushAgrawal-A2: Well, in an expanded text box... the editor text is always left aligned... 
to get proper alignment of canvas-text with editor and even the placeholder having top level text align works very well. This is what I intended to solve with this pr. 
Is there some specific bug with this implementation?

https://github.com/toeverything/blocksuite/assets/54364088/9352c360-32ab-4a90-b318-e02e39eac465

packages/blocks/src/page-block/edgeless/components/text/edgeless-text-editor.ts
AyushAgrawal-A2: will fix this

packages/blocks/src/page-block/edgeless/components/text/edgeless-text-editor.ts
AyushAgrawal-A2: finally understood the bug and problem
found a way to use top level textAlign an prevent excess shift in inline elements
textAlign right in edit mode:-
image

packages/blocks/src/page-block/edgeless/components/text/edgeless-text-editor.ts
AyushAgrawal-A2: I was trying to simplify the code... but these transforms that too in this specific order are necessary, reverted this part of change

doodlewind (approved):
Generally LGTM, thanks!

Flrande (approved):

doouding (approved):
Good to see the code has been simplified 😃

PR #5861, opened 2023-12-26
feat(presets): support double click to disable note sorting and fix typo

donteatfriedrice:
1. Support double click at blank area to disable note sorting option.
2. Fix highlight mask disappeared after mode switched.
3. Fix typo.


packages/playground/apps/starter/components/debug-menu.ts
```diff
@@ -204,7 +204,7 @@ export class DebugMenu extends ShadowlessElement {
   contentParser!: ContentParser;
 
   @property({ attribute: false })
-  navigationPanel!: CustomTOCOutlinePanel;
+  tocOutlinePanel!: CustomTOCOutlinePanel;
```
doodlewind, line 207: How about renaming `TOCPanel` -> `OutlinePanel` 😈

PR #5874, opened 2023-12-27
refactor(edgeless): surface elements api

doouding:
Move surface elements related API onto the surface model. Introduce a new pure data element model to replace the existing one to make it independent from the renderer. Additionally, this new element model will not depend on the DOM API.


doodlewind (commented):
Is it possible to remove legacy code defining fields on previous elements? I would supposed to see more deletions :)

doouding:
> Is it possible to remove legacy code defining fields on previous elements? I would supposed to see more deletions :)

Yes, but I'm not planning to do this in this PR. I'll start a new PR on cleanup and replacement or something once this gets merged.

doodlewind (approved):
Look forward to the housekeeping :)

PR #5878, opened 2023-12-27
refactor: support multiple-editors

### LINKED ISSUE #5153: playground example "multiple-editor" not working
# Description

open

https://blocksuite-toeverything.vercel.app/starter/?init=multiple-editor

or

http://localhost:5173/starter/?init=multiple-editor

The editor on the left doesn't work, the editor on the right works fine.

https://github.com/toeverything/blocksuite/assets/13579374/347f6735-701f-4d76-83f9-5969cc5fb140

# Possible Cause of the Issue

On this page, the two `EditorContainer`s utilize the same `Workspace` and reference the `LocalSelection` state within a shared `AwarenessStore`. Consequently, the selection state in the left editor consistently gets overridden by the right editor, causing this issue.

### LINKED ISSUE #5625: multiple-editor Uncaught Error: val does not exist
When I create multiple editors
  editor's editor.mode="page"
When multiple editors are displayed on the page at the same time, one editor in the editor will report an error: Uncaught Error: val does not exist
How should I solve this problem? I have a scenario where I use multiple editors at the same time.

image

AyushAgrawal-A2:
continuation of #5765 

closes #5625 
closes #5153

Changes:
1. Completed pending cleanup global querySelectors and events
2. UIEventDispatcher: 
   I tried moving events from document to editorHost, but there were few events which will originate from the document only. In multiple editor scenario each editor needs to be aware whether it is active or not, and respond to events only is it active.

   I have added a static field on UIEventDispatcher, which stores which dispatcher and thus which editor is active. It will run event handlers only if it is active.

   - The editor gets activated when the selection changes to an element inside the editor or the user clicks inside the editor.
   - The editor gets deactivated If there is a selection change to an element outside the editor.
   - There are a few overlays appended to document.body. I have added class "blocksuite-overlay" to these elements. If the user interacts with any such element the state will not change, i.e. the previous active editor which created the overlay will remain active.
4. For ease of testing of all three scenarios, added examples of independent multiple editors - doc-doc, doc-edgeless & edgeless-edgeless at
   - /examples/multiple-editors/doc-doc/
   - /examples/multiple-editors/doc-edgeless/
   - /examples/multiple-editors/edgeless-edgeless/

### DIFF
--- a/packages/block-std/src/event/control/clipboard.ts
+++ b/packages/block-std/src/event/control/clipboard.ts
@@ -16,12 +16,7 @@ export class ClipboardControl {
   }
 
   private _cut = (event: ClipboardEvent) => {
-    // if (
-    //   !event.target ||
-    //   !this._dispatcher.host.contains(event.target as Node)
-    // ) {
-    //   return;
-    // }
+    if (!this._dispatcher.focus) return;
 
     const clipboardEventState = new ClipboardEventState({
       event,
@@ -33,12 +28,7 @@ export class ClipboardControl {
   };
 
   private _copy = (event: ClipboardEvent) => {
-    // if (
-    //   !event.target ||
-    //   !this._dispatcher.host.contains(event.target as Node)
-    // ) {
-    //   return;
-    // }
+    if (!this._dispatcher.focus) return;
 
     const clipboardEventState = new ClipboardEventState({
       event,
@@ -50,12 +40,7 @@ export class ClipboardControl {
   };
 
   private _paste = (event: ClipboardEvent) => {
-    // if (
-    //   !event.target ||
-    //   !this._dispatcher.host.contains(event.target as Node)
-    // ) {
-    //   return;
-    // }
+    if (!this._dispatcher.focus) return;
 
     const clipboardEventState = new ClipboardEventState({
       event,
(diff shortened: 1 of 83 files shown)


doodlewind (commented):
Seems that I'm still unable to type paragraph content in the `examples/multiple-editors/doc-doc/` entry, is this expected?

AyushAgrawal-A2:
> * /examples/multiple-editors/doc-doc/

works for me 

https://github.com/toeverything/blocksuite/assets/54364088/098f3c4f-7060-4cc2-9998-030c1b9fdb1d

there may be some design / behaviour changes required for some widgets/overlays. except that everything should work.

doodlewind:
The `starter/init=multiple-editor` works great! but here's my recording running the `doc-doc` one:

https://github.com/toeverything/blocksuite/assets/7312949/84aa90b5-8ec5-46eb-ba61-b2258d0694d0

AyushAgrawal-A2:
> The `starter/init=multiple-editor` works great! but here's my recording running the `doc-doc` one:
> 
>  Screen.Recording.2023-12-30.at.02.13.12.mov

is the same issue occurring for doc-edgeless , edgeless-edgeless as well ?

doodlewind (commented):
@AyushAgrawal-A2 I double checked and all examples work now, that's awesome!

Generally LGTM but for some updates in `block-std`, I'd require @Saul-Mirone for better review.

AyushAgrawal-A2:
@doodlewind 
I noticed that the other editor's selection is not getting cleared. I will push another commit to fix that.
I should have checked again after cleanup... :{

Flrande (approved):

PR #5888, opened 2023-12-29
feat(edgeless): add more templates

doouding:


doodlewind (changes requested):
I think we should consider make the loading of these templates async.

doodlewind (approved):
LGTM

Issue #5880, opened 2023-12-27
Change code block to typescript cause text to wrap


EYHN:
https://github.com/toeverything/blocksuite/assets/13579374/bbb918f5-4fe4-4ad3-86c5-f7b42f5eb18f

PR #5891, opened 2023-12-29
fix(page): unexpected line wrap in code block

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/5880


doodlewind (approved):
Thanks for the fix!

PR #5895, opened 2023-12-30
fix: complete draghandle fixme

AyushAgrawal-A2:
completing draghandle fixme for surface selection path after #5852


doodlewind (approved):
LGTM, happy new year!

PR #5898, opened 2023-12-31
fix(edgeless): console error on block delete

AyushAgrawal-A2:
to fix console error from:
1. note-slicer on deleting top level note block
2. GridManager on deleting image/bookmark block inside top level note

https://github.com/toeverything/blocksuite/assets/54364088/a9c8aaa2-a2c6-460e-a8fb-688902e9b851


doodlewind (approved):
LGTM

PR #5900, opened 2024-01-01
feat: add print formatter for the block model

doouding:
The block model is wrapped in a proxy which makes it difficult to print useful information in the console. The [Custom Formatters](https://firefox-source-docs.mozilla.org/devtools-user/custom_formatters/index.html) allow us to config how a variable can be printed in the console which can help us solve this problem. 

I wrote a small print formatter that can print all the props of the given block model. It should be sufficient in most cases.

Before:

Screenshot 2024-01-01 at 10 31 25 PM

After:
Screenshot 2024-01-01 at 10 32 33 PM

**Note**: You need to check the `Enable custom formatters` in the Devtools preference to make it work.

Screenshot 2024-01-01 at 10 34 26 PM

### DIFF
--- a/packages/playground/apps/starter/utils.ts
+++ b/packages/playground/apps/starter/utils.ts
@@ -142,13 +142,13 @@ export function createEditor(page: Page, element: HTMLElement) {
   return editor;
 }
 
-const toStyledEntry = (key: string, value: string) => {
+function toStyledEntry(key: string, value: unknown) {
   return [
     ['span', { style: 'color: #c0c0c0' }, ` ${key}`],
     ['span', { style: 'color: #fff' }, `: `],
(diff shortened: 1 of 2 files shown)


packages/playground/apps/starter/utils.ts
```diff
@@ -141,3 +141,62 @@ export function createEditor(page: Page, element: HTMLElement) {
...
+        return ['div', { style: 'padding-left: 1em' }, ...propsArr];
+      }
+
+      return null;
+    },
+  },
+];
+
+// @ts-ignore
+window.devtoolsFormatters = devtoolsFormatter;
```
fourdim, line 202: Modify it in https://github.com/toeverything/blocksuite/blob/master/packages/playground/apps/starter/env.d.ts

fourdim (approved):

doodlewind (approved):

PR #5909, opened 2024-01-02
feat: support add custom inline nodes

Flrande:
After this PR is merged, externals can implement customization for inline elements by overriding the service for specific blocks.

### InlineSpec

`InlineSpec` is a format for defining inline elements. Externals can describe the rendering method and corresponding data structure for a particular inline element using this format.

```ts
type InlineSpecs<
  TextAttributes extends BaseTextAttributes = BaseTextAttributes,
> = {
  // Field name, corresponding to the key in delta attributes
  name: string;
  // Schema for this field
  schema: ZodTypeAny;
  // Matching function, if the function returns true, an inline element will be rendered internally using the renderer defined by this InlineSpec
  match: (delta: DeltaInsert<TextAttributes>) => boolean;
  // Rendering function, includes two parameters: delta and selected, where the latter indicates whether the element is selected
  renderer: AttributeRenderer<TextAttributes>;
  // An optional field, indicating whether the element is of embed type
  embed?: boolean;
};
```

### Renderer

The renderer in `InlineSpec` needs to return a `TemplateResult` to implement the rendering of inline elements. For its implementation, you can refer to the following link:

`affine-text` is the most common non-embed type of inline node in blocksuite, responsible for bolding, underlining, and other basic styles.
https://github.com/toeverything/blocksuite/blob/8fd0f98bf086a5416167971ee33c491bdac131ec/packages/blocks/src/_common/inline/presets/nodes/affine-text.ts#L45-L78

`latex-node` is an embed-type node used as an example in this PR, capable of rendering mathematical formulas.
https://github.com/toeverything/blocksuite/blob/8fd0f98bf086a5416167971ee33c491bdac131ec/packages/playground/apps/default/specs-examples/latex/latex-node.ts#L17-L65

### InlineMarkdownMatch

Currently, we have preset a mechanism to trigger inline formatting changes using markdown syntax. For example, in "abc", if you want to change "b" to bold, you need to enter 'a' + '\*\*b\*\*' + space + 'c'. `InlineMarkdownMatch` is a format that describes such shortcuts. Apart from realizing inline formatting transformations through these preset shortcuts, it is more common to achieve this through extending the `format-bar`. However, this goes beyond the scope of this PR, so no further description will be provided.

```ts
type InlineMarkdownMatch<
  TextAttributes extends BaseTextAttributes = BaseTextAttributes,
> = {
  name: string;
  pattern: RegExp;
  action: (props: {
    inlineEditor: InlineEditor<TextAttributes>;
    prefixText: string;
    inlineRange: InlineRange;
    pattern: RegExp;
    undoManager: Y.UndoManager;
  }) => ReturnType<KeyboardBindingHandler>;
};
```

### Latex Example

This PR uses new API to write a simplified example, which can be experienced at [branch playground](https://blocksuite-git-flrande-inline-custom-0102-toeverything.vercel.app/?exampleSpec=latex), and the main core code is as follows:
https://github.com/toeverything/blocksuite/blob/8fd0f98bf086a5416167971ee33c491bdac131ec/packages/playground/apps/default/specs-examples/latex/latex.ts#L20-L121

https://github.com/toeverything/blocksuite/assets/50035259/542da25f-c3da-49a7-bf96-3e312f58e591


doodlewind (commented):
Briefing the API surface in PR description will be very helpful :)

Saul-Mirone:
How to view the demo in playground?

Flrande:
done @doodlewind @Saul-Mirone

Saul-Mirone (approved):

pengx17:
There are more features in AFFiNE requires us to extend the default blocksutie doc editor. For example, we want to let the inline page reference to have different ui/ux in some cases - like localization for journal pages.
If we try to extend "LinkedPage" reference, this is what we can do right now:
1. define a custom`affine-reference`for AFFiNE
   1. use Lit or plain HTML/web component as the renderer expects lit template return values
   1) use use Lit if we want to extend the existing reference component
1. override `.mounted`in`ParagraphService`,`ListService`,`DatabaseService`to use the customized`AffineInlineSpec`
1. find and replace the services in the default `DocSpecs` preset 
1. assign the custom specs to the Editor component
It looks to me that this approach is still **not mature enough** for use in affine yet.
1. Lit issues:

   1. If we want to use React, then how to pass individual React component/element to Lit?

   1) If not, we may need to setup a Lit toolset in Affine (including`experimentalDecorators`?)
1. Behaviour extension:

   1. There is actually no way to extend the editor, but to replace the internal implementation. Not easy for behavior extension - the bs is a *white box*

   1) On the other hand, if blocksuite allows the AFFiNE side to define custom block renderers, the affine side may easily write bad data which breaks the data integrity given the blocksuite's schema. This makes it hard for the blocksuite side to revolve onwards.

Flrande:
> There are more features in AFFiNE requires us to extend the default blocksutie doc editor. For example, we want to let the inline page reference to have different ui/ux in some cases - like localization for journal pages.

If it's just about changing the UI, maybe more reasonable to provide some kind of interface directly in `affine-reference`? (just like #5954)

Issue #5922, opened 2024-01-04
Editor host getter return undefined after editor created. 


donteatfriedrice:
In starter and default playground example, the editor host getter return undefined after editor created. Cause contentParser not to work properly。

<img width="417" alt="Screenshot 2024-01-04 at 17 23 06" src="https://github.com/toeverything/blocksuite/assets/99816898/ace8817e-0c2e-47f1-9d2f-e2f5151112e7">

PR #5926, opened 2024-01-04
fix(playground): editor host undefined

donteatfriedrice:
Before converting contentParser to page service, make sure that the playground can work normally for the time being.

Note: ContentParser and its related usage will be removed later.


doodlewind (approved):
Should add `FIXME` comment

donteatfriedrice:
close #5922

Issue #5931, opened 2024-01-06
Indent is lost when wrap code


lawvs:
Related to https://github.com/toeverything/blocksuite/pull/5694

https://github.com/toeverything/blocksuite/assets/18554747/299e495d-5802-4b41-9569-339f219b5487

PR #5932, opened 2024-01-06
fix(page): show whitespace when wrap code

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/5931


doodlewind (approved):
Thanks for the fix!

PR #5934, opened 2024-01-07
refactor: remove legacy contentParser and make it as exportManager of page service

donteatfriedrice:


packages/playground/apps/default/components/quick-edgeless-menu.ts
```diff
@@ -164,6 +161,10 @@ export class QuickEdgelessMenu extends ShadowlessElement {
     return this.editor.page;
   }
 
+  get pageService() {
+    return this.editor.host.spec.getService('affine:page') as PageService;
```
doodlewind, line 165: @Saul-Mirone how about adding a shortcut getter at editor?

doodlewind (approved):

doodlewind:
Congratulations that we finally fully removed the `ContentParser`!

PR #5942, opened 2024-01-08
chore: reduce delay of user input behavior in test

lawvs:
It is a reliability test for the `InlineEditor`

This pull request reduces the typing delay in the test from `50` milliseconds to `20` milliseconds.

If set to `10ms`, it may cause random test failures. Most failures occur due to insufficient delay after `pressEnter`.

### DIFF
--- a/tests/image.spec.ts
+++ b/tests/image.spec.ts
@@ -105,8 +105,10 @@ test('can click and copy image', async ({ page }) => {
   await activeEmbed(page);
   await copyByKeyboard(page);
   await pressEnter(page);
+  await waitNextFrame(page);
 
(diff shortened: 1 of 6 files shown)


doodlewind:
@Flrande seems that the missing character problem is mitigated recently?

Flrande:
> @Flrande seems that the missing character problem is mitigated recently?

Yes, it may be related to me adjusting the rendering timing. If the test passes stably, we can lower the delay.

doodlewind:
Seems that there are 2 flaky test cases. Would you try modifying them to get this landed @lawvs?

Flrande (approved):

PR #5947, opened 2024-01-09
feat(playground): add pages panel

zzj3720:
image


doodlewind (approved):
Great to see this finally happen!

PR #5959, opened 2024-01-11
fix(edgeless): use default value when editsession parse error

regischen:


packages/blocks/src/surface-block/managers/edit-session.ts
```diff
@@ -190,9 +190,12 @@ export class EditSessionStorage {
...
   constructor(private _service: SurfaceService) {
-    const props = sessionStorage.getItem(SESSION_PROP_KEY);
-    if (props) {
-      this._lastProps = LastPropsSchema.parse(JSON.parse(props));
+    try {
+      this._lastProps = LastPropsSchema.parse(
+        JSON.parse(sessionStorage.getItem(SESSION_PROP_KEY) ?? '')
+      );
+    } catch {
+      /* empty */
```
doodlewind, line 198: You should NOT swallow error in this way. Log error message or throw it explicitly.

doodlewind (changes requested):
Should take care about error handling.

doodlewind (approved):

PR #5968, opened 2024-01-12
fix(page): change text attribute on converting inline-link to card

AyushAgrawal-A2:
[TOV-406](https://linear.app/affine-design/issue/TOV-406/optimize-the-implementation-effect-of-turning-inline-view-into-card)


houjoe0829 (approved):
Meets the expected requirements, great.

doodlewind (approved):

doodlewind (approved):

PR #5969, opened 2024-01-12
feat: export set link preview / image proxy api & update docs

darkskygit:


packages/blocks/src/_common/embed-block-helper/helper.ts
```diff
@@ -48,3 +54,78 @@ export function createEmbedBlock<
     view,
   };
 }
+
+// ========== Link Preview ==========
+
+const linkPreviewEndpoint = (() => {
```
Flrande, line 60: The worker endpoint here is a global singleton, which does not support multiple editors well. For instance, if two editors use different workers, I suggest managing the instances within the service.

darkskygit: but still need to share instances in a single editor instance, for example, link preview is used in bookmark/youtube/github block. does blocksuite currently provide an editor-level config sharing mechanism?

Flrande: For me, it's acceptable that each block service maintains an instance. Currently, our inline style configurations also follow one per block.

image

Flrande: Blocksuite is usually configured for block-level, focusing on blocks rather than the editor.

doodlewind (approved):

packages/blocks/src/_common/embed-block-helper/helper.ts
darkskygit: fixed

packages/blocks/src/_common/embed-block-helper/helper.ts
```diff
@@ -48,3 +54,73 @@ export function createEmbedBlock<
     view,
   };
 }
+
+// ========== Link Preview ==========
+
+export class LinkPreviewer {
```
doodlewind, line 60: I'm not sure if it's appropriate to add the preview config with the common embed block helper @Saul-Mirone.

Also, would be good to clarify the updated API in PR description.

darkskygit: pr's description has been updated

PR #5970, opened 2024-01-12
feat(presets): support `Ask AI` in format-bar

zzj3720:
- [x] creating mind-map by AI

https://github.com/toeverything/blocksuite/assets/17165520/ad8a254e-c9c3-4d24-85a0-552a69f5be4d

- [x] creating presentation by AI

https://github.com/toeverything/blocksuite/assets/17165520/75aabbdf-c27a-4abc-9441-f38da1f60453


packages/presets/src/fragments/copilot-panel/edgeless/logic.ts
```diff
@@ -39,36 +41,37 @@ export class AIEdgelessLogic {
...
+
   private unsub?: () => void;
   public toggleAutoGen = () => {
     if (this.unsub) {
       this.unsub();
       this.unsub = undefined;
       return;
     }
-    const edgeless = getEdgelessPageBlockFromEditor(this.editor.host);
+    const edgeless = getEdgelessPageBlockFromEditor(this.host);
```
doodlewind, line 52: I think this kind of util should be avoided, or you can just expose a getter on `edgeless-page-service` - and you can access the service here.

What are the APIs that are being used here? I think they should also be defined as standard methods on edgeless page service, cc @doouding

zzj3720: How do I get `edgeless-page-service`?

doodlewind (commented):
Seems that "insert below" doesn't work with text selection:

https://github.com/toeverything/blocksuite/assets/7312949/53702db2-56da-49b0-98f9-c847f8ef0a99

I think there are `selectedBlocks` and `selectedModels` getters on page service that could help with this.

PR #5972, opened 2024-01-12
refactor(edgeless): edgeless related api

doouding:
## Surface model & element model

- New `observe` decorator to allow observe the yjs type property. E.g., the `children` property of group element.
- New `convert` decorator to allow convert the new value before assigning to or updating the model property. It will gradually replace the existing function `propsToY`.
- Introduce middleware that adheres to the lifecycle of the surface model. This middleware can be utilized to implement features that depend on the creation, reading, updating, and deletion (CRUD) of elements and are also closely tied to the model. For example, the path of the `connector` is dynamically generated based on its `source` and `target` properties in real-time, and it is not stored persistently in Yjs. To ensure the surface model is plug-and-play, the logic for generating the connector path is integrated through middleware.
- New local property `externalXYWH` for element model, to store the rect dimensions of content associated with the element, such as the bounding rect of a group's title. This attribute is updated by the renderer during the rendering process. Once this data is stored, the relevant content becomes interactive and is also decoupled from the renderer.

## Edgeless API & canvas renderer
- Move most edgeless API from surface block to edgeless service.
- Replace the old canvas element model with the new one. And the renderer is now just a stateless renderer, all models' rendering implementation is just a pure function.
- Simplify the edgeless type to just two: `EdgelessBlockModel` and `ElementModel`. Both of them can be used with the `instanceof` operator.
- Remove redundancy slot including previous `elementUpdated`, `elementAdded`, and `elementRemoved`. Use the slot of the model instead.


doodlewind (commented):
TODO: remove redundancy in `onYBlockUpdated`

Saul-Mirone (approved):

PR #5980, opened 2024-01-13
fix: dragPreview / dropIndicator / dragEndPos viewport offsets

AyushAgrawal-A2:
before:

https://github.com/toeverything/blocksuite/assets/54364088/12633a02-cbf2-410b-935e-a94a60548944

after:

https://github.com/toeverything/blocksuite/assets/54364088/22db2e16-a9db-43c4-9dc5-4934594783ec

### DIFF
--- a/packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
+++ b/packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
@@ -24,6 +24,7 @@ import {
   type EdgelessTool,
   Point,
   type Selectable,
+  type Viewport,
 } from '../../_common/utils/index.js';
 import {
   asyncFocusRichText,
@@ -240,20 +241,7 @@ export class EdgelessPageBlockComponent extends BlockElement<
     return this._viewportElement;
   }
 
-  get viewport(): EdgelessViewport {
-    if (!this.viewportElement) {
-      return {
-        left: 0,
-        top: 0,
-        scrollLeft: 0,
-        scrollTop: 0,
-        scrollWidth: 0,
-        scrollHeight: 0,
-        clientWidth: 0,
-        clientHeight: 0,
-      };
-    }
-
+  get viewport(): Viewport {
     const {
       scrollLeft,
       scrollTop,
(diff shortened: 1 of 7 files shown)


packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
```diff
@@ -90,6 +90,17 @@ import {
 import { xywhArrayToObject } from './utils/convert.js';
 import { getCursorMode, isCanvasElement, isFrameBlock } from './utils/query.js';
 
+export interface EdgelessViewport {
```
Flrande, line 93: We already have a `PageViewport` interface, and I think `EdgelessPageBlockComponent` and `DocPageBlockComponent` can share the same interface.

https://github.com/toeverything/blocksuite/blob/ee3a4ff482d2459a58cdfa2bb752d055e3774f05/packages/blocks/src/page-block/doc/doc-page-block.ts#L24-L33

packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
```diff
@@ -229,6 +240,41 @@ export class EdgelessPageBlockComponent extends BlockElement<
     return this._viewportElement;
   }
 
+  get viewport(): EdgelessViewport {
+    if (!this.viewportElement) {
```
Flrande, line 244: If the `viewportElement` does not exist, I think it should throw an error directly, so as to maintain consistency with the implementation of `get viewportElement()`.

packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
AyushAgrawal-A2: @Flrande Done

packages/blocks/src/page-block/edgeless/edgeless-page-block.ts
AyushAgrawal-A2: @Flrande Done
thanks for the quick review

doodlewind (approved):
Awesome fix!

PR #5986, opened 2024-01-14
fix(page): show backlink for `affine:embed-linked-doc`

AyushAgrawal-A2:
adding backlink support for `affine:embed-linked-doc`

### DIFF
--- a/packages/blocks/src/page-block/edgeless/controllers/clipboard.ts
+++ b/packages/blocks/src/page-block/edgeless/controllers/clipboard.ts
@@ -523,7 +523,7 @@ export class EdgelessClipboardController extends PageClipboard {
   }
 
   private _createLinkedDocEmbedBlocks(linkedDocEmbeds: BlockSnapshot[]) {
-    const embedlinkedDocIds = linkedDocEmbeds.map(({ props }) => {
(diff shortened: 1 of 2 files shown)


packages/blocks/src/embed-linked-doc-block/embed-linked-doc-model.ts
```diff
@@ -12,6 +12,7 @@ export const EmbedLinkedDocStyles: EmbedCardStyle[] = [
 
 export type EmbedLinkedDocBlockProps = {
   pageId: string;
+  type: 'Subpage' | 'LinkedPage';
```
doodlewind, line 15: We should try to remove Subpage in newly created data and code. Is it possible to use only LinkedPage here?

AyushAgrawal-A2: I have "Subpage" is for forward compatibility. default prop is "LinkedPage" only
image

AyushAgrawal-A2: @doodlewind
done

doodlewind (approved):

PR #5993, opened 2024-01-15
feat: support config `reference-node` externally

Flrande:
In some situations, we need to pass certain configurations from the outside to the `reference-node`. This PR enables us to convey these configurations through the service.

```ts
// support affine:paragraph, affine:list and affine:database now
const service = block.service;
const referenceNodeConfig = service.referenceNodeConfig;
// setup `Page`
this.referenceNodeConfig.setPage(page);
// config custom icon and title
this.referenceNodeConfig.setCustomIcon(customIcon);
this.referenceNodeConfig.setCustomTitle(customTitle);
// fully customized content (interactions of click and hover will still be retained)
this.referenceNodeConfig.setCustomContent(customContent);
```

### DIFF
--- a/packages/blocks/src/_common/inline/presets/nodes/reference-node/reference-config.ts
+++ b/packages/blocks/src/_common/inline/presets/nodes/reference-node/reference-config.ts
@@ -1,10 +1,15 @@
 import type { Page } from '@blocksuite/store';
 import type { TemplateResult } from 'lit';
 
+import type { AffineReference } from './reference-node.js';
+
 export class ReferenceNodeConfig {
-  private _customIcon: TemplateResult | null = null;
-  private _customTitle: string | null = null;
-  private _customContent: TemplateResult | null = null;
+  private _customIcon: ((reference: AffineReference) => TemplateResult) | null =
+    null;
+  private _customTitle: ((reference: AffineReference) => string) | null = null;
+  private _customContent:
+    | ((reference: AffineReference) => TemplateResult)
+    | null = null;
   private _page: Page | null = null;
 
   get customIcon() {
@@ -23,15 +28,15 @@ export class ReferenceNodeConfig {
     return this._customContent;
   }
 
-  setCustomContent(content: TemplateResult | null) {
+  setCustomContent(content: ReferenceNodeConfig['_customContent']) {
     this._customContent = content;
   }
 
-  setCustomIcon(icon: TemplateResult | null) {
+  setCustomIcon(icon: ReferenceNodeConfig['_customIcon']) {
     this._customIcon = icon;
   }
 
-  setCustomTitle(title: string | null) {
+  setCustomTitle(title: ReferenceNodeConfig['_customTitle']) {
     this._customTitle = title;
   }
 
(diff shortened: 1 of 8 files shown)


packages/blocks/src/_common/inline/presets/nodes/reference-node/reference-config.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  get customTitle() {
+    return this._customTitle;
+  }
+
+  get page() {
+    return this._page;
+  }
+
+  setCustomIcon(icon: TemplateResult | null) {
```
pengx17, line 21: The interface here only allows a single value for all reference-node rendered by a paragraph service, right?
From affine side what we may want here is to pass in a render function (possibly with delta + block context), instead of a single value.

Flrande: Do you need to retain the basic interactive logic, such as hover and click?

pengx17: Yes please.

Flrande: done

pengx17: Can you please declare the icon/title as a renderer function like

```ts
  setCustomContent(content: TemplateResult | ((context: { delta: Delta, block: Block } ) => TemplateResult)  | null) {
    this._customContent = content;
  }
```

It is not so useful if you don't give the context for downstream to know what to render

Flrande: done

packages/blocks/src/list-block/list-service.ts
```diff
@@ -2,24 +2,28 @@ import { BlockService } from '@blocksuite/block-std';
...
   override mounted(): void {
     super.mounted();
-    (
-      this.inlineManager as unknown as InlineManager<AffineTextAttributes>
-    ).registerSpecs(affineInlineSpecs);
-    (
-      this.inlineManager as unknown as InlineManager<AffineTextAttributes>
-    ).registerMarkdownMatches(affineInlineMarkdownMatches);
+
+    this.referenceNodeConfig.setPage(this.page);
```
pengx17, line 21: The page reference may change.

Flrande: What do you mean? When a new page is loaded, the mount method will be re-executed, and the config will be updated.

pengx17: Good to know!

doodlewind (approved):

xiaofeiniuuu:
How to update the config value of
const config = provider.getOptional(ReferenceNodeConfigIdentifier) ​​?? {}; after upgrading
Do I need to use ReferenceNodeConfigExtension? How to use it?
I see that affine-paragraph no longer has referenceNodeConfig

PR #5994, opened 2024-01-15
feat: change note visibility

donteatfriedrice:
Support to change note visibility:
- [x] `note:props:hidden` -> `note:props:displayMode` . Note now has three different display mode:
  - [x] `DocOnly`: only display the note (note content) in doc mode and can not be selected in edgeless mode.
  - [x] `EdgelessOnly`: only display the note (note content) in edgeless mode.
  - [x] `DocAndEdgeless`: display on both doc and edgeless modes.
- [x] User can change note display mode with note toolbar or outline panel. 
- [x] Update note auto complete buttons, only keep the buttons on the left and right sides，and the left button would be hidden when zoom greater than 150%.
- [x] Remove auto connect line.
- [x] Update index label. Page visible and edgeless only note will have different index label and tooltip.
- [x] Update outline panel empty placeholder. 
- [x] Note added on doc mode will display on both doc and edgeless by default. While added on edgeless mode will display on edgeless mode by default. ( notes in templates will display on edgeless only by default ). 
- [ ] Add a notice board in outline panel to notice user some contents hidden on edgeless page if there are doc only notes.


doodlewind (approved):
LGTM, thanks for the feat!

PR #5995, opened 2024-01-15
feat(edgeless): support drag to add note

donteatfriedrice:
Support drag to add note with:
- [x] Customized width and height.
- [x] Background last edited.
- [x] The default restriction is that the added note needs to be greater than the minimum width and minimum height. A toast component needs to be added to prompt the user.

https://github.com/toeverything/blocksuite/assets/99816898/3d383200-8196-41a8-97de-8e15a68330f5


doodlewind (approved):
Awesome work!

Issue #6001, opened 2024-01-16
@blocksuite/preset@0.11.0 package style issues


nawbc:
@blocksuite/preset package styles are inconsistent with Playground,

WebKitGTK 2.38.0 , Chrome120
![image](https://github.com/toeverything/blocksuite/assets/45007226/308de144-34fc-4373-bfcd-20cd4ae6dfd5)

WebKitGTK 2.38.0
![image](https://github.com/toeverything/blocksuite/assets/45007226/29848aca-e39e-4ded-9cd1-62df7d43af27)

Source code
https://github.com/toeverything/blocksuite/blob/7fdc76819a3dfdb677c2c4f23c94e2d50a567d3e/packages/playground/examples/basic/doc/main.ts

<details>
  <summary><mark><font color=darkred>Sys info</font></mark></summary>
<code>
 System:
    OS: Linux 6.6 EndeavourOS
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
    Memory: 3.84 GB / 15.40 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 20.4.0 - ~/.nvs/default/bin/node
    Yarn: 3.6.3 - ~/.nvs/default/bin/yarn
    npm: 9.7.2 - ~/.nvs/default/bin/npm
    pnpm: 8.12.0 - ~/.nvs/default/bin/pnpm
    bun: 1.0.13 - ~/.nvs/default/bin/bun
    Watchman: 20231109.143709.0 - /usr/bin/watchman
  Managers:
    Cargo: 1.73.0 - ~/.cargo/bin/cargo
    RubyGems: 3.3.25 - /usr/bin/gem
  Utilities:
    CMake: 3.27.7 - /usr/bin/cmake
    Make: 4.4.1 - /usr/bin/make
    GCC: 13.2.1 - /usr/bin/gcc
    Git: 2.43.0 - /usr/bin/git
    Clang: 16.0.6 - /usr/bin/clang
    Ninja: 1.11.1 - /usr/bin/ninja
    FFmpeg: 6.0 - /usr/bin/ffmpeg
    Curl: 8.4.0 - /usr/bin/curl
  Virtualization:
    Docker: 24.0.6 - /usr/bin/docker
  SDKs:
    Android SDK:
      API Levels: 31, 33, 34
      Build Tools: 30.0.3, 33.0.1, 34.0.0
[...]

PR #6005, opened 2024-01-17
fix: input overflow menu (#6001)

nawbc:
#6001 Drag handle wrong position due to webkit2gtk bug.


doodlewind (approved):
LGTM, thanks for the fix!

PR #6007, opened 2024-01-17
refactor: remove global query for affine-editor-container

AyushAgrawal-A2:


pengx17 (approved):

doodlewind (approved):
Thanks for the swift fix!

PR #6009, opened 2024-01-17
fix(edgeless): enable double click on cards in surface

AyushAgrawal-A2:
[AFF-540](https://linear.app/affine-design/issue/AFF-540/double-clicking-on-the-linked-doc-does-not-open-the-corresponding-doc)


houjoe0829 (approved):
it meets expectations.

PR #6020, opened 2024-01-18
docs: move docs package to docs/main

doodlewind:


fourdim:
Hey, I don't think we need to rename this for adding a blog. What about wait and see?

doodlewind:
There will be a new presets.blocksuite.io site, so having multiple doc packages could make sense. Also there will be packages/framework dir for grouping the store, block-std, lit and inline packages, so the structure is supposed to be reused more widely.

fourdim:
> There will be a new presets.blocksuite.io site, so having multiple doc packages could make sense. Also there will be packages/framework dir for grouping the store, block-std, lit and inline packages, so the structure is supposed to be reused more widely.

Sure. Then no problem on my end.

PR #6031, opened 2024-01-19
fix: remove link panel from affine-doc-editor

regischen:
\cc @pengx17

### DIFF
--- a/packages/presets/src/fragments/bi-directional-link/bi-directional-link-panel.ts
+++ b/packages/presets/src/fragments/bi-directional-link/bi-directional-link-panel.ts
@@ -201,16 +201,13 @@ export class BiDirectionalLinkPanel extends WithDisposable(LitElement) {
   }
 
   private get _docPageElement() {
-    const affineDocEditor = this.closest('affine-doc-editor');
-    assertExists(affineDocEditor);
-    const docPageElement = affineDocEditor.querySelector('affine-doc-page');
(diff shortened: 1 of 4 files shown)


packages/presets/src/fragments/bi-directional-link/bi-directional-link-panel.ts
```diff
@@ -201,16 +201,13 @@ export class BiDirectionalLinkPanel extends WithDisposable(LitElement) {
   }
 
   private get _docPageElement() {
-    const affineDocEditor = this.closest('affine-doc-editor');
-    assertExists(affineDocEditor);
-    const docPageElement = affineDocEditor.querySelector('affine-doc-page');
+    const docPageElement = this.parentElement?.querySelector('affine-doc-page');
```
pengx17, line 204: The parent element is not necessarily the one that wraps both `affine-doc-page` & `bi-directional-link`

pengx17: I think you can add a new property to let the upstream to pass in an doc-page element instead

regischen: Alright, that sounds reasonable.

regischen: done

pengx17 (approved):

doodlewind (approved):

PR #6055, opened 2024-01-21
fix: code hotkey should not effect in global

### LINKED ISSUE #6052: page doc - if a codeblock exists, we can't delete any block anymore
https://github.com/toeverything/blocksuite/assets/10269911/51d4dbc6-aa3f-4d1b-909c-781508d99280

I can't delete my text block anymore after a codeblock is created. 

It auto-selects to the codeblock instead

Flrande:
Close #6052


Saul-Mirone (approved):

packages/block-std/src/event/dispatcher.ts
```diff
@@ -165,16 +165,12 @@ export class UIEventDispatcher {
     const handlers = this._handlersMap[name];
     if (!handlers) return;
 
-    let output: EventScope | undefined;
+    let output = this._buildEventScopeBySelection(name);
```
Saul-Mirone, line 168: After discussion, we should control the event scope by event type.

Saul-Mirone (changes requested):

package.json
```diff
@@ -74,7 +74,7 @@
     "eslint": "^8.56.0",
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-lit": "^1.11.0",
-    "eslint-plugin-prettier": "^5.1.3",
```
Saul-Mirone, line 77: Why downgrade?

packages/block-std/src/event/control/keyboard.ts
```diff
@@ -9,6 +9,17 @@ export class KeyboardControl {
   constructor(private _dispatcher: UIEventDispatcher) {}
 
   private _createContext(event: Event, keyboardState: KeyboardEventState) {
+    const selection = window.getSelection();
+    if (selection && selection.rangeCount > 0) {
+      const range = selection.getRangeAt(0);
```
Saul-Mirone, line 14: I mean selection model, not native selection.

PR #6060, opened 2024-01-22
fix(edgeless): should update index label when note order changed

donteatfriedrice:
https://github.com/toeverything/blocksuite/assets/99816898/8f3d6676-d784-4957-b7c8-f080ef0fe144


doodlewind (approved):
Thanks for the fix!

PR #6064, opened 2024-01-22
fix: attachment block should not contenteditable

Flrande:


lawvs (approved):
I think declaring the property directly is also a good pattern.

```ts
  override readonly contentEditable = 'false';
```

PR #6066, opened 2024-01-22
chore: remove `blockVversions` in snapshot

Saul-Mirone:
In favor of https://github.com/toeverything/blocksuite/pull/6065

### DIFF
--- a/packages/store/src/transformer/base.ts
+++ b/packages/store/src/transformer/base.ts
@@ -4,7 +4,10 @@ import type { AssetsManager } from './assets.js';
 import { fromJSON, toJSON } from './json.js';
 import type { BlockSnapshot } from './type.js';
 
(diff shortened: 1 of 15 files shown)


packages/store/src/transformer/base.ts
```diff
@@ -46,27 +50,29 @@ export class BaseBlockTransformer<Props extends object = object> {
...
   }: FromSnapshotPayload): Promise<SnapshotReturn<Props>> {
-    const { flavour, id, props: _props } = json;
+    const { flavour, id, version, props: _props } = json;
 
     const props = this._propsFromSnapshot(_props);
 
     return {
       id,
       flavour,
+      version: version ?? -1,
```
doodlewind, line 60: Will this be breaking change when opening existing snapshots?

doodlewind (approved):

PR #6073, opened 2024-01-23
fix(page):  shortcuts work in the captions of embed-card and attachment

### LINKED ISSUE #5912: Shortcuts do not work in the captions of custom blocks
`Ctrl+A/C/X/V` shortcuts do not work in the captions of custom blocks

https://github.com/toeverything/blocksuite/assets/18554747/e7b1dc02-b56b-4bd2-aff5-ee008cc3210b

AliceLanniste:
close #5912

Here is a  similar bug in embed-card,attachment and image block. I think common template and logic can be  encapsulated.


doodlewind:
The issue is fixed by #6119 so I'd close this one in favor of it. Still thanks for your effort!

PR #6089, opened 2024-01-23
fix: bookmark and image block should not contenteditable

Flrande:


doodlewind (approved):

doodlewind (commented):
Maybe this can be defined in the `createEmbedBlock` helper @Saul-Mirone?

PR #6099, opened 2024-01-24
fix(edgeless): prevent file drop indicator on block elements in surface

AyushAgrawal-A2:
[TOV-426](https://linear.app/affine-design/issue/TOV-426/图片拖入后，卡住-insert-的-ui，并且添加图片不生效。)

### DIFF
--- a/packages/blocks/src/_common/components/file-drop-manager.ts
+++ b/packages/blocks/src/_common/components/file-drop-manager.ts
@@ -103,7 +103,10 @@ export class FileDropManager {
     let result: DropResult | null = null;
     if (element) {
       const model = getModelByBlockComponent(element);
-      result = calcDropTarget(point, model, element);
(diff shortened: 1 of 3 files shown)


doodlewind (commented):
I think TOV-426 can still be reproduced:

https://github.com/toeverything/blocksuite/assets/7312949/621ea09c-73fd-4e19-92a5-63402fad3bb5

Dropping in page mode (e.g. the page info fragment) still has the issue.

AyushAgrawal-A2:
> I think TOV-426 can still be reproduced:
> 
>  Screen.Recording.2024-01-25.at.10.51.56.mov 
> Dropping in page mode (e.g. the page info fragment) still has the issue.

@doodlewind 
Done

doodlewind (approved):
Thanks for the fix!

Issue #6092, opened 2024-01-24
Unexpected behavior when cancel import image


zkwolf:
![GIF](https://github.com/toeverything/blocksuite/assets/8555127/3fa4ff94-d9a9-49a5-833f-dd7c1e08ab61)

The original input command block has disappeared.
And you can also see, when I input a and backspace, the editor delete two  characters.

PR #6104, opened 2024-01-24
fix(page): should not remove empty lines when canceling an action

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/6092

In the previous logic, empty lines were removed by some slash actions regardless of the outcome. Now, only successful actions will remove the empty lines.


doodlewind (approved):
Thanks for the fix!

PR #6105, opened 2024-01-24
feat(presets): add cover page for AI slides

zzj3720:


packages/presets/src/fragments/copilot-panel/chat/templates/basicCover.ts
```diff
@@ -0,0 +1,95 @@
+export const basicCover = () => ({
```
doodlewind, line 1: Please stick to the `basic-cover.ts` naming convention.

PR #6112, opened 2024-01-25
chore(edgeless): surface service cleanup

doouding:


doodlewind (approved):
Requires bumping `setNavigatorMode` API usage in AFFiNE

PR #6117, opened 2024-01-25
fix(edgeless): prevent native selection on drag select in edgeless

AyushAgrawal-A2:
Bug:

https://github.com/toeverything/blocksuite/assets/54364088/d2d3a321-e657-450a-acc9-a66dcb7db3dc


doodlewind (approved):
Good catch!

PR #6122, opened 2024-01-25
feat: add setup function for spec

Saul-Mirone:
```ts
const DocPageSpec: BlockSpec<DocPageBlockWidgetName, DocPageService> = {
  schema: PageBlockSchema,
  service: DocPageService,
  view: {
    /* ... */
  },
  setup: (slots, disposables) => {
    disposables.add(slots.mounted.on(onMounted));
    disposables.add(slots.viewConnected.on(onViewConnected));
    disposables.add(slots.widgetConnected.on(onWidgetConnected));
  }
}

```


doodlewind (commented):
Guess the `docs/guide/block-spec` docs could also be updated.

And wondering for block-specific setup logic, what' the best practice to distinguish between using the `service.mount` and `spec.setup`?

Saul-Mirone:
> Guess the `docs/guide/block-spec` docs could also be updated.
> 
> And wondering for block-specific setup logic, what' the best practice to distinguish between using the `service.mount` and `spec.setup`?

1. I'll add the doc in next pr.
2. The `setup` can be configured when you try **use** a spec, the `service.mount` method is used to define a spec.

doodlewind (approved):
🚀🚀🚀

PR #6124, opened 2024-01-25
fix(presets): no frame preview in sidebar

donteatfriedrice:
To fix:
[TOV-487](https://linear.app/affine-design/issue/TOV-487/右侧边栏preview-settings被遮挡)
[TOV-489](https://linear.app/affine-design/issue/TOV-489/edgeless-toc无法预览)

https://github.com/toeverything/blocksuite/assets/99816898/ab22c381-e2dd-4e7c-a040-06ec800b054c

Todo:
- [ ] Add some tests.


packages/presets/src/fragments/frame-panel/header/frame-panel-header.ts
```diff
@@ -182,7 +182,7 @@ export class FramePanelHeader extends WithDisposable(LitElement) {
         this._settingPopperShow = display === 'show';
       },
       14,
-      -120
```
CatsJuice, line 185: Is this modification made to solve the [TOV-487](https://linear.app/affine-design/issue/TOV-487/右侧边栏preview-settings被遮挡) issue? It seems to have shifted only a few pixels. In AFFiNE, this panel is set to `overflow:hidden`, can Blocksuite mount the `Popup` to the body instead?

donteatfriedrice: The frame panel setting popup seems not to be hidden. This line is just used to adjust the position.

doodlewind (approved):

packages/presets/src/fragments/frame-panel/header/frame-panel-header.ts
CatsJuice: It may be clipped in AFFiNE, I think it's better to mount `Popup` under `body` if it's supported in blocksuite.

doouding (approved):
Adding some basic tests is expected :)

donteatfriedrice:
> Adding some basic tests is expected :)

I'll add some later.

packages/presets/src/fragments/frame-panel/header/frame-panel-header.ts
donteatfriedrice: OK, I'll check whether can mount the panel Popup under body.

PR #6138, opened 2024-01-26
test: add basic frame panel tests

donteatfriedrice:


doodlewind (commented):
We already have tests in the presets package as integration test, so making this `tests/fragments` sounds better.

donteatfriedrice:
> We already have tests in the presets package as integration test, so making this `tests/fragments` sounds better.

That make sense. I have updated it.

doodlewind (approved):

PR #6146, opened 2024-01-29
fix: export png incorrectly when svg without width and height

### LINKED ISSUE #5552: SVG can't be displayed after exporting to PNG
https://github.com/toeverything/AFFiNE/assets/142964904/10cac547-9fb7-4729-9c8a-a5d6de6e6915

donteatfriedrice:
To fix #5552 

This issue is due to the fact that when the svg does not set the width and height, but only sets the viewBox, it will adaptively change the size, causing only a small part of the rendering result to be displayed. If the width and height of an svg are set correctly, the exported svg will be complete. 

In order to solve this problem, special processing needs to be done when the img src is svg, and the width and height of svg need to be set.

Since there is no way to directly determine whether the img is of svg type from the url suffix of the img src, we need to fetch the blob first and then determine the blob type, which may cause the export time to become longer.

https://github.com/toeverything/blocksuite/assets/99816898/8f6137fa-8079-4966-b008-46cf45f6dc87

### DIFF
--- a/packages/blocks/src/_common/export-manager/export-manager.ts
+++ b/packages/blocks/src/_common/export-manager/export-manager.ts
@@ -158,6 +158,7 @@ export class ExportManager {
         });
 
         await this._replaceRichTextWithSvgElement(element);
+        await this.replaceImgSrcWithSvg(element);
       },
       useCORS: this._exportOptions.imageProxyEndpoint ? false : true,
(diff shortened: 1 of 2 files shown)


doodlewind (approved):

doodlewind (approved):
Thanks for the fix!

PR #6152, opened 2024-01-30
fix: remove duplicated title when exporting markdown

lawvs:
This pull request removes the duplicated title when exporting markdown.


doodlewind (approved):

doodlewind:
Also encountered this bug, thanks for the fix!

PR #6153, opened 2024-01-30
feat(page): keep same doc scroll position when mode change

### LINKED ISSUE #3172: Hope that AFFiNE can remember the previously opened position of the page when switching
- This corresponds to the https://github.com/toeverything/blocksuite/issues/2479
- The progress is only preserved during switching between Page and Edgeless modes within the same page. If switching between different pages, the progress will still be reset

regischen:
close #3172

### DIFF
--- a/packages/blocks/src/page-block/doc/doc-page-block.ts
+++ b/packages/blocks/src/page-block/doc/doc-page-block.ts
@@ -1,5 +1,5 @@
 import { PathFinder, type PointerEventState } from '@blocksuite/block-std';
-import { assertExists, Slot } from '@blocksuite/global/utils';
+import { assertExists, debounce, Slot } from '@blocksuite/global/utils';
 import { BlockElement } from '@blocksuite/lit';
 import type { Text } from '@blocksuite/store';
 import { css, html } from 'lit';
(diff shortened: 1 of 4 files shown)


doodlewind (commented):
We tend to keep the editor component a plain wrapper without feature implemented here. Any chance supporting this at the doc page block component level?

regischen:
> We tend to keep the editor component a plain wrapper without feature implemented here. Any chance supporting this at the doc page block component level?

sure

packages/blocks/src/page-block/doc/doc-page-block.ts
```diff
@@ -224,6 +227,26 @@ export class DocPageBlockComponent extends BlockElement<
         })
       );
     });
+
+    const scrollElement = this.closest(
+      'div.affine-doc-viewport'
```
Flrande, line 232: Can we use the nearest parent scroll container here to avoid relying on external implementations?

regischen: em, after thinking, I feel this feature better be implemented in affine side.  Because now blocksutie provides components, then composed in affine.

PR #6169, opened 2024-01-31
perf: optimize frame preview performance

donteatfriedrice:
### Change:
- Frame preview need to update viewport when updated surface element overlap with the frame, so we should listen the elementUpdated slot. To optimize the performance, I try to debounce the handleElementUpdated function call.
- Frame preview also need to be updated when top level block overlap with the frame, we can move the slot listener to surface-ref portal to reduce the update rate (only update when that portal model inside that frame).

before:
Screenshot 2024-02-01 at 12 04 08

after:
Screenshot 2024-02-01 at 11 59 55

But I think this can only reduce part of the CPU usage. In fact, the performance bottleneck lies in the update of the whiteboard. When a frame containing multiple notes, images, brushs, shapes and other elements is moved in the whiteboard, even if the frame panel is not opened at this time CPU usage will also skyrocket. **We need to continue to optimize the performance of the whiteboard for this situation**.

https://github.com/toeverything/blocksuite/assets/99816898/78df43ef-d0f1-413f-b420-e7fa6b9b5a6d

Screenshot 2024-02-01 at 12 08 58


doodlewind (approved):
Could you elaborate in PR description about where the perf bottleneck is, and what's the best practice to avoid it? Thanks!

PR #6171, opened 2024-02-01
chore: replace ambiguous "page" with "doc"

JimmFly:
In order to distinguish between `doc` and `page mode` [TOV-532](https://linear.app/affine-design/issue/TOV-532/affine-侧-page-名称更改为-doc)


doodlewind (approved):
Following this PR, I also plan to rename all the `page` objects to `doc` in the future.

PR #6180, opened 2024-02-02
refactor: move edgeless page slots to edgeless page service

donteatfriedrice:
- Move edgeless page slots to edgeless page service and remove some slots.
- Move zoom related actions to service.
- Uniformly use toolManager.setEdgelessTool when changing tool types.


doodlewind (approved):
LGTM, thanks for the feat!

PR #6182, opened 2024-02-02
refactor: simplify the usage of `affine-block-selection`

AyushAgrawal-A2:
Changes:
- simplify the usage of `affine-block-selection`

i.e. replace
```ts 
${this.selected?.is('block')
    ? html`<affine-block-selection></affine-block-selection>`
    : nothing}
```
with 
```ts 
<affine-block-selection .block=${this}></affine-block-selection>
```


doodlewind (approved):
Neat work, thanks!

PR #6185, opened 2024-02-02
fix: hide format-bar / component-toolbar while dragging

### LINKED ISSUE #2963: Optimization for dragging and dropping to create copies
related with https://github.com/toeverything/blocksuite/issues/1811

Adding some optimization options
- To display a double-arrow mouse cursor when holding down the Option key, similar to how Figma handles it, to indicate that a copy is about to be created.
- Disabling the display of component menu while dragging.

AyushAgrawal-A2:
closes #2963


doodlewind (approved):
LGTM, thanks for the fix!

PR #6196, opened 2024-02-05
refactor: move doc page block slots to service

donteatfriedrice:
- [x] Move doc page block slots to doc page service.
- [x] Remove slots that are no longer used


fourdim (approved):

doodlewind (approved):
Nice work! I think the legacy `forwardSlot` https://github.com/toeverything/blocksuite/blob/master/packages/presets/src/editors/editor-container.ts and the `AbstractEditor` can also be removed and migrated here. cc @Flrande

Issue #3803, opened 2023-08-02
Tooltips when export PNG failed


Brooooooklyn:
If network is down, export PNG just silence failed without any tips

PR #6205, opened 2024-02-07
fix(playground): tip user when exporting failed (#3803)

water-in-stone:


water-in-stone:
@Brooooooklyn @Flrande PTAL

doodlewind (changes requested):
Thanks for the enhancement! But I don't think the toast until should be exported for public reuse. Since the playground extensively uses shoelace for auxiliary UI, could you use the alert component in shoelace instead? https://shoelace.style/components/alert

PR #6212, opened 2024-02-13
fix(inline): line break issue

golok727:
Subject: Resolve Word Break Issue in Table Layout

### Description:
This pull request addresses and resolves the word break issue identified in [@AFFiNE#5676](https://github.com/toeverything/AFFiNE/issues/5676). Previously, the table layout experienced undesired breaks between words. With this update, the issue has been rectified, and the table now breaks words appropriately.

Link to Issue:
[@AFFiNE#5676](https://github.com/toeverything/AFFiNE/issues/5676)

Before
Before
After
After

### DIFF
--- a/packages/docs/index.md
+++ b/packages/docs/index.md
@@ -19,7 +19,7 @@ hero:
 
 features:
   - title: 🧩 Headless Editor Framework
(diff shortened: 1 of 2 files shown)


fourdim (approved):

fourdim:
Thank you for your contribution!

PR #6215, opened 2024-02-14
feat: use spacebar to switch to pan in brush and eraser mode

golok727:
Subject: Feature Enhancement: Spacebar Toggle for Pan Mode in EdgelessMode

### Reason
I've observed that using the middle mouse button is quite handy for switching to hand mode during drawing in EdgelessMode, as a dedicated Photoshop user, I believe it is more ergonomic to use the spacebar with the LMB to switch to the pan mode while drawing. To enhance the user experience, I've refactored to enable the use of the spacebar to toggle between pan mode and brush/eraser mode seamlessly.

Pull Request Changes:
- Use the spacebar to switch to pan mode while in brush | eraser mode


doodlewind (approved):
LGTM, thanks for the enhancement!

PR #6223, opened 2024-02-19
fix(page): synced block placeholder should not appear after adding content

AyushAgrawal-A2:
bug: 

https://github.com/toeverything/blocksuite/assets/54364088/a2b2091b-4126-4f0d-9dcf-0ee5138812cf


doodlewind (approved):
LGTM

PR #6233, opened 2024-02-20
fix(edgeless): change connector toolbar tip

regischen:
close [TOV-413](https://linear.app/affine-design/issue/TOV-413/add-shortcut-keys-to-the-connectors-hovertips)


doodlewind (approved):
LGTM

PR #6235, opened 2024-02-20
fix(edgeless): more button copy not work

regischen:
close [TOV-575](https://linear.app/affine-design/issue/TOV-575/group-frame-copy之后粘贴无效)


doodlewind (approved):
LGTM

doodlewind (approved):

PR #6236, opened 2024-02-20
fix: improve code block performance

Flrande:
Test 600+ lines code:

before:
image

after:
image

For further optimization, we may need to refactor the rendering mechanism of the inline editor.


packages/framework/inline/src/services/range.ts
```diff
@@ -72,7 +72,10 @@ export class RangeService<TextAttributes extends BaseTextAttributes> {
     if (this.editor.mounted) {
       // range change may happen before the editor is prepared
       await this.editor.waitForUpdate();
-      this.editor.requestUpdate(false);
+      // improve performance
+      setTimeout(() => {
```
doodlewind, line 76: Can we use `requestIdleCallback` here instead?

doodlewind (approved):

PR #6239, opened 2024-02-20
feat: limit entry to synced doc block using awareness store flag

AyushAgrawal-A2:
Changes:
- Added `enable_synced_doc_block` flag in awareness store flags. This is enabled for playground environment.

To-Do
- Add this flag in AFFiNE and enable for canary.


doodlewind (approved):
LGTM

PR #6240, opened 2024-02-20
fix: tooltip flickers when the mouse is moved slightly

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/6241


doodlewind (approved):
LGTM, thanks for the fix

PR #6249, opened 2024-02-20
fix: cursor cannot be moved as expected in firefox

### LINKED ISSUE #6217: Text breaks out of layout
First off: Thank you for this amazing colab editor - please keep up the good work!

Reproduction:
In the demo on your page, place the cursor before the first character of the first line, then hit arrow up and write characters.

https://github.com/toeverything/blocksuite/assets/2837147/d8038476-c7ed-465d-b28b-b3fe96c95236

### LINKED ISSUE #6218: Arrow key down loops back to first line
As a user, I would not expect that when pressing `arrow key down` on the last sentence, that the cursor would loop back to the first line.

https://github.com/toeverything/blocksuite/assets/2837147/6bfc96db-ec5a-4734-9208-0d2aaa0d8bff

(text breaking of layout in previous bug report)

Flrande:
Close #6217
Close #6218

### DIFF
--- a/packages/blocks/src/page-block/doc/doc-page-block.ts
+++ b/packages/blocks/src/page-block/doc/doc-page-block.ts
@@ -49,8 +49,14 @@ export class DocPageBlockComponent extends BlockElement<
   DocPageBlockWidgetName
 > {
   static override styles = css`
+    editor-host:has(> affine-doc-page) {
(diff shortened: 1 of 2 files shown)


doodlewind (approved):
LGTM, thanks for the fix!

PR #6250, opened 2024-02-20
fix(edgeless): cannot click more button in AFFiNE

donteatfriedrice:
To fix:
 [TOV-453](https://linear.app/affine-design/issue/TOV-453/新-note-通过选区创建的-note，点击后-more-后，直接关闭了-toolbar)
[TOV-471](https://linear.app/affine-design/issue/TOV-471/group-toolbar-点more无法激活)

Should not select edgeless component toolbar text and should not trigger selectionchange event when click toolbar button.

https://github.com/toeverything/blocksuite/assets/99816898/1e4701d6-e71b-491f-ade8-7a1bc6b112ce


doodlewind (approved):
LGTM

PR #6251, opened 2024-02-20
fix(page): correct image fetch via proxy with response.ok check

AyushAgrawal-A2:
Changes:
- Added response.ok check for image fetch via proxy, as fetch does not throw error for non-200 statuses.

Linked with https://github.com/toeverything/AFFiNE/pull/5849


doodlewind (approved):
Thanks for the fix!

PR #6255, opened 2024-02-21
fix(edgeless): inner frame hittest

regischen:
close [TOV-431](https://linear.app/affine-design/issue/TOV-431/frame-内的-frame-无法通过标题点击移动)


doodlewind (approved):
LGTM

PR #6256, opened 2024-02-21
fix(playground): missing root model after refresh page

zkwolf:
Step to reproduce bug:

1. new a link page
2. refresh page
3. click page

What is actually happening:

page have not redirect but console log root model miss.


doodlewind (approved):
Thanks for the fix!

PR #6265, opened 2024-02-22
feat: add context in command result

Saul-Mirone:


doodlewind (approved):
So I think this can be used to avoid previous pattern to pass context content via side effect assign in closure? Some code pattern can be migrated in this case.

Saul-Mirone:
> .

Yeah, we can get it from result now.

PR #6268, opened 2024-02-22
feat(page): add `appendParagraph` api in pageService

regischen:


packages/blocks/src/page-block/doc/doc-page-block.ts
```diff
@@ -187,6 +207,13 @@ export class DocPageBlockComponent extends BlockElement<
     );
   };
 
+  appendParagraph() {
```
AyushAgrawal-A2, line 210: pls use arrow function

AyushAgrawal-A2 (changes requested):

packages/blocks/src/page-block/doc/doc-page-block.ts
regischen: done

packages/blocks/src/page-block/doc/doc-page-block.ts
```diff
@@ -187,6 +207,13 @@ export class DocPageBlockComponent extends BlockElement<
     );
   };
 
+  appendParagraph = () => {
```
doodlewind, line 210: Note sure where is the use case of this API?

regischen: This is intened to used for Affine side.
To complete clicking end area of editor to append paragraph.
I found that because editor is composed in affine, it is convenient that the feature is implementd in affine.

doodlewind: In this case, think the API set should be exposed on page service.

AyushAgrawal-A2: @regischen 
"To complete clicking end area of editor to append paragraph." - can this be done in blocksuite itself ?

regischen: Done

packages/blocks/src/page-block/page-service.ts
```diff
@@ -233,4 +236,35 @@ export class PageService extends BlockService<PageBlockModel> {
...
+      note = page.getBlockById(noteId) as NoteBlockModel;
+    }
+    return note;
+  }
+
+  appendParagraph = () => {
+    const note = this._getLastNoteBlock();
+    if (!note) return;
+    const id = this.page.addBlock('affine:paragraph', {}, note);
+    const host = document.querySelector('editor-host');
```
AyushAgrawal-A2, line 266: pls don't use global queries... this will fail in case of multiple editors

packages/blocks/src/page-block/page-service.ts
```diff
@@ -233,4 +236,35 @@ export class PageService extends BlockService<PageBlockModel> {
...
+    if (!note) {
+      const noteId = page.addBlock('affine:note', {}, page.root?.id);
+      note = page.getBlockById(noteId) as NoteBlockModel;
+    }
+    return note;
+  }
+
+  appendParagraph = () => {
+    const note = this._getLastNoteBlock();
+    if (!note) return;
```
AyushAgrawal-A2, line 264: if there is no note, should'nt we add a new note, and append paragraph in that note ?

AyushAgrawal-A2 (changes requested):

packages/blocks/src/page-block/page-service.ts
regischen: so how we get editor host in service?

AyushAgrawal-A2: @regischen `this.std.host`

packages/blocks/src/page-block/page-service.ts
regischen: here is a type problem. but I can move logic from `_getLastNoteBlock` to here.

packages/blocks/src/page-block/page-service.ts
regischen: okay, ty!

doodlewind (approved):

PR #6271, opened 2024-02-22
fix(page): change toolbar position for all block cards to top-left

AyushAgrawal-A2:
Changes:
- position of all card toolbar to top-left
- fixed synced doc toolbar bug


Svaney-ssman (approved):
Thank you

doodlewind (approved):

PR #6273, opened 2024-02-22
fix(page): allow block cards to expand horizontally on note resize

AyushAgrawal-A2:
[AFF-615](https://linear.app/affine-design/issue/AFF-615/once-the-card-exceeds-752px-it-will-not-adapt-to-the-width)
Changes:
- allow all block card to expand horizontally on note resize, affected blocks :
image


Svaney-ssman (approved):
Goood

doodlewind (approved):

PR #6281, opened 2024-02-22
fix(page): popup menu should not close, when on top of hoverable block

AyushAgrawal-A2:
Bug: due to 4px offset b/w toolbar and popup menu, if the gap is over another block which shows toolbar on hover, then current toolbar will close on moving mouse from toolbar through the gap to the popup.

https://github.com/toeverything/blocksuite/assets/54364088/ecbba892-902b-446c-abb9-edf1abe4cb0f


doodlewind (approved):
Thanks for the fix!

PR #6288, opened 2024-02-23
fix(edgeless): shift panel-wrapper to keep it in view

fundon:
### Before

Screenshot 2024-02-23 at 15 57 33

### After

Screenshot 2024-02-23 at 15 56 36


doodlewind (approved):
Thanks for the fix!

PR #6289, opened 2024-02-23
feat: add pdf annotation support

doouding:
There are two types of annotation, one is TextAnnotation which allows you to select text in PDF and add comments on it; the other is ClipAnnotation which is basically an image.

```typescript
export type ClipAnnotation = {
  type: AnnotationType.Clip;
  comment: Y.Text;
  highlightRects: Record<number, [number, number, number, number][]>;
};

export type TextAnnotation = {
  type: AnnotationType.Text;
  comment: Y.Text;
  /**
   * hightlighted rects in each page
   *
   * Rect<page, [x, y, w, h]>
   */
  highlightRects: Record<number, [number, number, number, number][]>;

  /**
   * The highlight text
   */
  text: string;
};

type ClipAnnotation = {
  type: AnnotationType.Clip;
  highlightRect: [number, number, number, number];
};
```


ChildWangWorld:
很期待的功能，请问为啥这个PR关闭了？ 😩

PR #6292, opened 2024-02-24
fix(edgeless): no need to adjust position of component-toolbar when more-actions-container is opened

fundon:
Related to #6288 

### Before
Screenshot 2024-02-24 at 16 18 34

### After
Screenshot 2024-02-24 at 16 18 04


doodlewind (approved):
LGTM

PR #6293, opened 2024-02-24
feat:  add columns-block

ununian:
features:
* Overall, it's similar to Notion's columns and Feishu's split columns.
* It can be created through a Slash menu.
* The columns should be able to nest editors, just the doc type is fine, no need for edgeless.
* It should be possible to drag blocks from outside of the columns-block into the columns-block and vice versa.

Feishu's split columns:

image


ununian:
https://github.com/toeverything/blocksuite/assets/4506072/e2b481d7-2d3e-4a69-b54d-859ee55e8b43

I finished a preliminary version.
Please see if it fulfills the requirements

doodlewind (commented):
Thanks for the feat! It's great capabilities, and it's also in our long-term roadmap to support column block. But since the editor composed by these blocks is used in AFFiNE, so if this is going to land in AFFiNE as a product feature, it would require more effort in UI design and UX details, but we may not have resource in current iteration plan.

So I'd propose checking in your implementation but not bundle it in `AffineSchemas` in `CommonFirstPartyBlockSpecs` by default. In playground examples (like `specs-examples-latex`), we can still import the column block spec and compose it into a new editor entry, just like how it shows in the https://blocksuite.io/guide/working-with-block-tree.html#defining-new-blocks docs.

packages/blocks/src/columns-block/utils.ts
```diff
@@ -0,0 +1,55 @@
+export function normalSizes(
+  sizes: number[],
+  targetSum: number,
+  min: number,
+  max: number
+) {
+  let sum = sizes.reduce((acc, val) => acc + val, 0);
+  const diff = targetSum - sum;
+  let adjustments = sizes.map(() => diff / sizes.length); // 初始均匀分配调整值
```
doodlewind, line 9: Please use English for comment :)

packages/blocks/src/columns-block/index.ts
```diff
@@ -0,0 +1,12 @@
+import type { ColumnsBlockModel } from './columns-model.js';
+
+export * from './columns-block.js';
+export * from './columns-model.js';
+
+declare global {
+  namespace BlockSuite {
+    interface BlockModels {
+      'affine:columns': ColumnsBlockModel;
```
doodlewind, line 9: The flavour is supposed to be `affine:column` and make it `blocks/src/column`, instead of using the plural term of "columns"  🤔

In slash menu, keeping the "Columns" entry text for user looks fine.

ununian:
@doodlewind  I'll create a new pr to provide columns for playground, and another to work around the bugs encountered in writing the columns block.

ununian:
I'll create a new pr to provide columns block for playground.

PR #6300, opened 2024-02-26
fix(edgeless): add default text color of shape-element

fundon:
### Before
Screenshot 2024-02-26 at 16 29 24

### After
Screenshot 2024-02-26 at 16 28 36


doouding (commented):
I'm not sure if it's a good idea to set a CSS variable as a fallback value 'cause that means in order to render an element with a fallback property you need to set a proper CSS variable in advance.

Or assign the CSS variable as the field initialization.

doouding (approved):

PR #6302, opened 2024-02-26
feat: comment poc

Flrande:
https://github.com/toeverything/blocksuite/assets/50035259/af9cfca9-8272-4182-89f4-da4ef6a1a298


doodlewind (approved):
Thanks for the feat! We should land this prototype as an internal feature FTM.

PR #6305, opened 2024-02-27
fix(page): the style of editor-host does not take effect

wumo1016:
When affine-page-root is used as a direct child element of editor-host, the style of editor-host does not take effect.

https://github.com/toeverything/blocksuite/assets/47590453/e71a18dd-300e-4fdd-a19c-c8853cf014e2


wumo1016:
> Thanks for the fix, could you sign the CLA.md as new first-contributor? Thanks!

It's already finished！

doodlewind (approved):
Thanks for your fix!

Issue #6063, opened 2024-01-22
Support importing external image URL in markdown


zkwolf:
`toPageSnaptshot` get assets to write blob，but only pass file to this func, is there should pass `job.assetsManager` to assets?

https://github.com/toeverything/blocksuite/blob/f2054e2d231d7160af85683b8b81ca49365e66d4/packages/blocks/src/page-block/widgets/linked-doc/import-doc/import-doc.ts#L38

fourdim:
Is this a question or bug report?

zkwolf:
I think this is a bug.

fourdim:
What's your markdown file looks like?
Does it contain any image link to the local file?
If so, it's not a bug. Import it using zip option.

doodlewind:
Tried importing the markdown, seems that the actual issue is that we are not able to import following image in markdown:

```
![](https://artem.krylysov.com/images/2023-rocksdb/rocksdb-lookup.png)
```

fourdim:
Thank you for your contribution!

PR #6321, opened 2024-02-28
feat(page): support import external images

zkwolf:
To support #6063 , add `assets` param to `toDocSnapshot` func.

Btw when the time of import is too long , it should add some operation like loading animate to make ensure the page link are inserted.


fourdim (approved):
I'm ok with this change, but will this actually solve this issue?

zkwolf:
Yes, I've tested importing the markdown file, and the images do get imported.

PR #6326, opened 2024-02-28
feat: disable embed view button when linked doc inside another embed synced doc

donteatfriedrice:
To close [AFF-626](https://linear.app/affine-design/issue/AFF-626/should-hide-the-embed-view-toolbar-entry-in-nested-synced-doc)

https://github.com/toeverything/blocksuite/assets/99816898/fb469f03-6a4e-4e33-8b2d-4b010e685a13


doodlewind (approved):
LGTM

PR #6329, opened 2024-02-29
feat: support fuzzy matching in `matchFlavours`

wumo1016:
Optimize the matching mode of the matchFlavours method to support subsequent matching of a certain type of custom component. For example：
`matchFlavours('affine:note-block-container', ['affine:note-block-*']) => true`
`matchFlavours('affine:note-block-sheet', ['affine:note-block-*']) => true`


doodlewind (approved):
LGTM since there might be usage for `affine:embed:*`, thanks for the feat.

Issue #6322, opened 2024-02-28
AutoCompleteOverlay displays the wrong shape when the shape is an ellipse


fundon:
<img width="459" alt="Screenshot 2024-02-28 at 19 17 17" src="https://github.com/toeverything/blocksuite/assets/27926/c58fb9a1-c4aa-4aba-9b5c-ae265d8b80bc">

PR #6334, opened 2024-02-29
fix(edgeless): auto-complete displays incorrect shape

lawvs:
Fix https://github.com/toeverything/blocksuite/issues/6322

This pull request adds a draw function to the shape utils.

This allows the `AutoCompleteOverlay` to render different shapes.

Screenshot 2024-02-29 at 17 32 00


donteatfriedrice (approved):
Great job! Thanks for the fix!

PR #6338, opened 2024-02-29
fix(edgeless): width or height should not be negative

fundon:
Screenshot 2024-02-29 at 20 27 45

https://github.com/toeverything/blocksuite/assets/27926/f76dc01e-1692-4053-aef7-1fa5a7c9a57f


donteatfriedrice (approved):
Thanks for the fix.

PR #6341, opened 2024-02-29
fix(edgeless): fix laggy selection & legacy template rendering

doouding:
### Change
- Add new method `walk` into transformer `Job` to traverse a template (use to modify the snapshot content before importing start)
- Add property fallback to fix legacy template rendering
- Add `watch` decorator to watch element property change
- Add new local property `commands` to brush to avoid double computation
- Fix laggy multi-selection

### DIFF
--- a/packages/blocks/src/_common/transformers/zip.ts
+++ b/packages/blocks/src/_common/transformers/zip.ts
@@ -6,7 +6,7 @@ import type {
   Workspace,
   WorkspaceInfoSnapshot,
 } from '@blocksuite/store';
-import { getAssetName, Job } from '@blocksuite/store';
+import { getAssetName, Job, sha } from '@blocksuite/store';
 import JSZip from 'jszip';
 
 import { replaceIdMiddleware } from './middlewares.js';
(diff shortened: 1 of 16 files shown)


doodlewind (commented):
Could you elaborate about how the multi-selection is optimized and how is the measured result? Thanks.

It's weird that I haven't observed the laggy case in the windows laptop in current master, so the perf update could really be subjective for the moment.

doouding:
> Could you elaborate about how the multi-selection is optimized and how is the measured result? Thanks.
> 
> It's weird that I haven't observed the laggy case in the windows laptop in current master, so the perf update could really be subjective for the moment.

Well, I recorded the performance of the multi-selection action and eliminated the unnecessary render and computation during the action to reduce the frame times. It may still take some time with CPU 6x slow down, but it's better than before optimization.

Before
Screenshot 2024-03-03 at 8 42 03 PM

After
Screenshot 2024-03-03 at 8 42 25 PM

But like you said, the next step that needs to be done more than anything else is a stable performance metrics testing.

doodlewind (approved):

PR #6344, opened 2024-03-01
refactor: encode url key in playground

regischen:


packages/presets/src/fragments/copilot-panel/copilot-service/copilot-config.ts
```diff
@@ -24,6 +24,43 @@ type VendorPack<T extends AllServiceKind> = {
   impl: ServiceImpl<GetMethod<T>, unknown>;
 };
 
+class SimplePRNG {
+  constructor(private seed: number) {}
+  // 一个简单的PRNG，使用线性同余生成器
```
doodlewind, line 29: Please mind not using Chinese comment, and link to some reference material of this kind of algorithm would be better

packages/presets/src/fragments/copilot-panel/copilot-service/copilot-config.ts
```diff
@@ -24,6 +26,52 @@ type VendorPack<T extends AllServiceKind> = {
...
+function shuffleArray(array: number[], seed: number) {
+  const prng = new SimplePRNG(seed);
+  for (let i = array.length - 1; i > 0; i--) {
+    const j = Math.floor(prng.next() * (i + 1));
+    [array[i], array[j]] = [array[j], array[i]];
+  }
+}
+
+/**
+ * It's used for encoding the OpenAI key in the URL.
```
doodlewind, line 54: Don't put the sensitive encoding logic (and also don't mention OAI 😅) in this public repo. Instead, you can [use a secret gist](https://chat.openai.com/share/a855a26b-fcf8-41ea-8cd6-c0d452e96321) and share the encode script internally.

doodlewind (approved):

PR #6347, opened 2024-03-01
fix(inline): failed to delete embed node using backspace

### LINKED ISSUE #6275: Page link can't use backspace to delete after adding a new line
Step to reproduce:

1. use @ to add a new page
2. add new line 
3. refresh page
4. cursor to the back of page link
5. press backspace but can't delete the page link

Flrande:
Close #6275


doodlewind (approved):

doodlewind (commented):
Should be better to add a test here to prevent regressions?

PR #6348, opened 2024-03-01
feat(edgeless): add pressure support for brush mode

L-Sun:
https://github.com/toeverything/blocksuite/assets/20479050/090727e4-7cba-4f47-8d8f-3feed5e30caf

### DIFF
--- a/packages/blocks/src/surface-block/element-model/brush.ts
+++ b/packages/blocks/src/surface-block/element-model/brush.ts
@@ -24,6 +24,8 @@ export type BrushProps = BaseProps & {
    * [[x0,y0],[x1,y1]...]
    */
   points: number[][];
+  // empty if the hardware dose not suppot pressure such as mouse.
+  pressures: number[];
   color: string;
   lineWidth: number;
 };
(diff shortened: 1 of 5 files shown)


packages/blocks/src/surface-block/element-model/brush.ts
```diff
@@ -56,6 +58,9 @@ export class BrushElementModel extends ElementModel<BrushProps> {
   @yfield()
   points: number[][] = [];
 
+  @yfield()
+  pressures: number[] = [];
```
doodlewind, line 62: For existing data without this field, we need to make sure it doesn't break - you can edit in try-blocksuite.vercel.app and export the snapshot to test it.

Just to double check with @doouding that I think this yfield is supposed to store using plain array in yjs, not YArray?

doouding: > For existing data without this field, we need to make sure it doesn't break - you can edit in try-blocksuite.vercel.app and export the snapshot to test it.
> 
> Just to double check with @doouding that I think this yfield is supposed to store using plain array in yjs, not YArray?

Yes， this is going to be plain array. 

I wonder if the pressure could be store in the points array, make it a [number, number, number] array.

L-Sun: It is okay to store it in a points array, but I don't know whether this is a breaking change. I will test this later.

L-Sun: Currently, pressure and coordinates are stored in the `points` array, and it is compatible with previous snapshots.

doodlewind (approved):
Thanks for the feat!

PR #6351, opened 2024-03-02
fix(edgeless): menu-container scrolling in safari

fundon:
https://github.com/toeverything/blocksuite/blob/0beb354f4bead57a2cc4fe7b42db71db6641c600/packages/blocks/src/root-block/edgeless/components/toolbar/common/slide-menu.ts#L132

here should use `event.deltaX`


doodlewind (approved):
Thanks for the fix!

PR #6359, opened 2024-03-04
fix: remove static slot and controling state

doouding:
Use non-static slot and state to avoid conflict in multiple editor instances scenario.


doodlewind (approved):
LGTM

PR #6361, opened 2024-03-04
chore: update import path of `AFFINE_BLOCK_HUB_WIDGET`

wumo1016:


doodlewind (approved):
LGTM but note that we are going to remove the block hub since it's deprecated in the product. So we're also open to PR removing all of its related code :)

PR #6364, opened 2024-03-04
fix(page): handle delete or backspace align with embed block

### LINKED ISSUE #6327: An error occurs when deleting or backspacing adjacent to an embed block.
https://github.com/toeverything/blocksuite/assets/99816898/f3c34553-3994-40a7-9738-176c60c42911

Screenshot 2024-02-28 at 22 27 59

donteatfriedrice:
To close #6327 

Forward delete before embed block will be ignored. 

Pressing backspace when adjacent to an embed block will select the embed block.

### DIFF
--- a/packages/blocks/src/_common/components/rich-text/rich-text-operations.ts
+++ b/packages/blocks/src/_common/components/rich-text/rich-text-operations.ts
@@ -23,7 +23,8 @@ import {
 } from '../../../_common/utils/selection.js';
 import type { ListBlockModel } from '../../../list-block/index.js';
 import type { RootBlockModel } from '../../../root-block/index.js';
-import type { ExtendedModel } from '../../types.js';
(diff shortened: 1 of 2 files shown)


doodlewind (changes requested):
I think this should a use case for #6329

donteatfriedrice:
> I think this should a use case for #6329

It seems that we can not directly pass 'affine:embed-*' to matchFlavours since the expected parameter Key should be a type extending (keyof BlockSuite.BlockModels)[]. But pattern like 'affine:embed-*' is not a key of BlockSuite.BlockModels. Should we relax the type restrictions of functions, such as changing key to string type?

doodlewind:
Is it possible to use template literal type here? But based on current type check state, this PR should also be fine to land.

doodlewind (approved):

PR #6365, opened 2024-03-04
fix(page): update code theme by page theme

### LINKED ISSUE #6355: Code block theme should update when toggling themes
https://github.com/toeverything/blocksuite/assets/18554747/52abb37f-9fc0-4f51-ba9f-3c6de251e777

JunIce:
Close #6355


doodlewind (approved):

PR #6367, opened 2024-03-04
fix(edgeless): position should be calculated after the update is complete

fundon:
### Before

https://github.com/toeverything/blocksuite/assets/27926/3d8c4d7c-26d3-45ac-bd69-fc9fab58a9e7

### After

https://github.com/toeverything/blocksuite/assets/27926/b7805c12-bc39-4ed9-b841-7fbad365d5f3


doodlewind (approved):
Thanks for the fix!

Issue #6366, opened 2024-03-04
Block id mutates when pasting text


lawvs:
Every time text is pasted, a new block instance is created. This issue can result in a poor collaboration experience.

https://github.com/toeverything/blocksuite/assets/18554747/75973f42-f7d5-4e4a-8a32-1e5bad92140a

PR #6373, opened 2024-03-04
feat(page): add language detection

lawvs:
This pull request introduces a new feature that automatically detects the language of code blocks.

The feature utilizes machine learning (AI!) to predict the language. (Power by [guesslang](https://github.com/yoeo/guesslang))

The language detection will be triggered in the following scenarios:

- Nothing will happen if
  - The detect language module has not been loaded. (This module will only load when a code block is connected to the page.)
  - The detection confidence is lower than 20%. (You can inspect the confidence in the [guesslang playground](https://guesslang.waterwater.moe/))
- Code block will set the language automatically when
  - Text pasted into an empty plaintext code block.
  - Text is pasted into a code block and the user selects **all** text within the plaintext code block.
- When a user attempts to select a language for a code block. It will offer a language recommendation.

https://github.com/toeverything/blocksuite/assets/18554747/b1107382-32e1-4593-9ce4-dd148958809d

~~Automatic language selection when pasting code from other editors is helpful for users, but it is blocked by https://github.com/toeverything/blocksuite/issues/6366.~~

## Trade-offs

- Despite making this module lazy-load, it will still add an extra 800kb of dependencies to the codebase.
- The model (700kb) is loaded from the CDN, resulting in slow loading times initially.

Screenshot 2024-03-05 at 05 22 58

### Bundles

| Entry          | Size                 | Gzip                  | Brotli                 |
| -------------- | -------------------- | --------------------- | ---------------------- |
| examples/basic | 14 MB (**+46.4 kB**) | 2.9 MB (**+10.7 kB**) | 1.79 MB (**+5.63 kB**) |

### Packages

| Name   | Size                 | Gzip                  | Brotli                |
| ------ | -------------------- | --------------------- | --------------------- |
| blocks | 2.2 MB (**+5.8 kB**) | 513 kB (**+2.01 kB**) | 374 kB (**+1.14 kB**) |
| editor | 84 B                 | 89 B                  | 63 B                  |
| store  | 83 B                 | 88 B                  | 63 B                  |
| inline | 84 B                 | 88 B                  | 63 B                  |

### DIFF
--- a/packages/blocks/package.json
+++ b/packages/blocks/package.json
@@ -26,6 +26,7 @@
     "@types/mdast": "^4.0.3",
     "@types/sortablejs": "^1.15.8",
     "@types/webfontloader": "^1.6.38",
+    "@vscode/vscode-languagedetection": "^1.0.22",
     "buffer": "^6.0.3",
     "date-fns": "^3.3.1",
     "file-type": "^16.5.4",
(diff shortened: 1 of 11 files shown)


Saul-Mirone:
This is really cool.

PR #6378, opened 2024-03-05
chore(examples): add start script

Flrande:


doodlewind (changes requested):
The `pnpm dev react-basic` is documented in the README

PR #6391, opened 2024-03-05
test: add test for legacy template

doouding:
I uploaded two onboarding template snapshots from AFFiNE and used them as a legacy template rendering test. The test checks the properties of every surface element to ensure they are not nullish values. Additionally, the xywh property should match the correct pattern, and the counts of surface elements should also match the snapshot JSON. And, of course, there should be no error during the render process.


doodlewind (approved):
LGTM, we may need to further increase the usage of CDN template data in testing.

PR #6393, opened 2024-03-05
feat(docs): link examples

Flrande:


doodlewind (changes requested):
I don't think it's expected to expose these "app-level" examples in the doc site recently - their functionalities are basically the same. Seeing the demo is less important than grabbing the code. And the links to current examples are already added in https://github.com/toeverything/blocksuite/pull/6375/commits/ff0a69139f5ec1acc85ab3e9b6aca8d5f059b559

Also, for the incoming expected "embed-like" (instead of "app-like") examples in the doc site, it'll be far too heavy to add one example in this manner (with per-example bundler configs).

PR #6395, opened 2024-03-05
fix(edgeless): bebas-neue light url

fundon:
Screenshot 2024-03-07 at 13 36 36


doodlewind (approved):
Thanks for the fix!

fundon:
@doodlewind If we want to pass the tests, we need to add `BebasNeue-Light.woff` font to the server.

Just only copy `BebasNeue%20Light.woff` to `BebasNeue-Light.woff`.

Screenshot 2024-03-07 at 14 07 26
Screenshot 2024-03-07 at 14 07 45

PR #6406, opened 2024-03-06
feat: add text style commands

Saul-Mirone:


doodlewind (commented):
How to expose these commands in the doc site? Try running the docs package locally and you can see the typing of commands seems to be missing in typedoc.

Saul-Mirone:
> How to expose these commands in the doc site? Try running the docs package locally and you can see the typing of commands seems to be missing in typedoc.

Added

doodlewind (approved):

PR #6409, opened 2024-03-06
feat: initial @blocksuite/indexer

EYHN:
# goal

* Ability to search doc content and locate blocks.
* Ability to search for backlinks
* Able to search blob usage and assist blob cleaning logic
* Can replace the underlying search engine backend, compatible with sqlite & elasticsearch & memory (for debug) backend.

# implement

The implementation will be divided into 2 parts,

### indexer

Inverted index of entities and their properties, and be able to search for entities through properties, for example, search for doc through doc content segments.

need support:

* Basic matching (find document by a (string|int|date) value)
* boolean combination query
* fulltext searching
* highlight matched segments
* range searching (search by date range)
* terms aggregation (Count how many pages each tag has)
* ranking based on match score
* observable search result

### crawler (WIP)

Automatic process to synchronize external data sources and indexer.

Possible external data sources:

* workspace ydoc
* graphql (for affine)

Need to synchronize `create`, `modify` and `delete` operations

# concept

### Schema

Defines the types of all fields and tells the backend how to initialize the data table.
All subsequently inserted must comply with this schema.

### Indexer

Aggregation of a single index data table. An indexer only indexes one type of entity, so we can have multiple indexers.

### IndexWriter

A transaction written to Indexer, with `commit` and `rollback` methods, using `indexer.write()` to create an IndexWriter. Writing should be atomic.

### Query

Query conditions, there are many different queries, which can also be combined, currently there are `boolean` and `match`

### Collector

Collector is a set of parameters that tells the search engine what search results the caller wants.

There are currently 4 types.

* **TopDocs**: Output the item IDs with the highest matching score.
* **Count**: Output the total number of matched items.
* **TermsAggregation**: Counts the number of occurrences of each value in the specified field.
* **Highlight**: Return highlighted snippets so you can show users where the query matches are. (only for `FullText` Field)

# example

basic using: https://github.com/toeverything/blocksuite/blob/eyhn/feat/indexer/packages/framework/indexer/src/__tests__/memory.unit.spec.ts


doodlewind (commented):
Is this going to replace the flexsearch dependent in store package? If so we should mark the deprecated APIs.

EYHN:
> Is this going to replace the flexsearch dependent in store package? If so we should mark the deprecated APIs.

It will eventually replace flexsearch and the current backlink indexer.

PR #6414, opened 2024-03-06
feat(edgeless): note sliced support to keep edgeless style

donteatfriedrice:
To close [AFF-668](https://linear.app/affine-design/issue/AFF-668/剪刀剪开以后，样式没有继承)


doodlewind (approved):
I think this should be closing a linear issue?

donteatfriedrice:
> I think this should be closing a linear issue?

Yes, it is. I have update the linear status.

PR #6415, opened 2024-03-06
fix(edgeless): paste group with same relative index

### LINKED ISSUE #6242: ​Copy paste a group in edgeless mode doesn't not create the result expected
### What happened?

I copied and pasted a group composed of a shape with text + an image (from the monthly calendar template).
The result of the paste was not a group, it was a shape and an image separated, and their display order was incorrect ( the image was behind the shape whereas in the original group the image was in front). This make the usage of this kind of group almost unusable.

### Distribution version

macOS ARM 64 (Apple Silicon)

_No response_

### Relevant log output

_No response_

### Anything else?

I got the same issue on PC

regischen:
Close #6242


doouding:
Remapping element index is a common use case (also used in edgeless template inserting). Extending the original `compare` function to accept an index map rather than reimplementing it would be great. But it is ok to do this later.

doouding (approved):

PR #6416, opened 2024-03-06
feat: add commands for text styles

Saul-Mirone:


packages/blocks/src/_common/configs/text-format/config.ts
```diff
@@ -30,87 +25,95 @@ export const textFormatConfigs: TextFormatConfig[] = [
...
+    activeWhen: host => {
+      const [result] = host.std.command
+        .pipe()
+        .isTextStyleActive({ key: 'bold' })
+        .run();
+      return result;
+    },
     action: host => {
-      handleCommonStyle(host, 'bold');
+      host.std.command.pipe().toggleBold().run();
```
doodlewind, line 36: Could we rename the `pipe` naming to `chain`? Or else which naming looks more intuitive for you?

Saul-Mirone: > Could we rename the `pipe` naming to `chain`? Or else which naming looks more intuitive for you?

I think we should rename the infra apis in other pr. And both `chain` and `pipe` look good to me.

PR #6428, opened 2024-03-07
refactor: remove withHost

Saul-Mirone:
The `withHost` command is no longer needed since we can get `std` from the command `ctx` now.


doodlewind (commented):
Well you should also update the corresponding markdown codes.

PR #6430, opened 2024-03-07
feat: extract config for format bar widget

Saul-Mirone:


doodlewind (commented):
Any example using the config API in PR description?Q

Saul-Mirone:
> Any example using the config API in PR description?Q

The API will be introduced in next pr. This pr is only for config renderer.

PR #6437, opened 2024-03-09
refactor: use native matrix transform

fundon:
Maybe we can use these native apis to refactor `math-utils`.

* [DOMPoint](https://caniuse.com/?search=DOMPoint)
* [DOMRect](https://caniuse.com/?search=DOMRect)
* [DOMQuad](https://caniuse.com/?search=DOMQuad)
* [DOMMatrix](https://caniuse.com/?search=DOMMatrix)


doodlewind:
Not sure if we can get perf optimization (and if so, how much) from the refactoring?

fundon:
Bad news, the native is slow. You can check out this repo https://github.com/fundon/dom-matrix-benchmark.
Screenshot 2024-03-10 at 03 52 55

So I think we can be clear about the DOMMatrix-related code.

PR #6451, opened 2024-03-11
perf(edgeless): filter icons in line-width-panel

fundon:
1. `flex-direction` does not need to be `row-reverse`
2. filter icons in a single loop


doodlewind (approved):
LGTM

PR #6454, opened 2024-03-11
fix(store): use lib0 sha256 when api not available

EYHN:
fix https://github.com/toeverything/AFFiNE/issues/6049

### DIFF
--- a/packages/framework/store/src/persistence/blob/utils.ts
+++ b/packages/framework/store/src/persistence/blob/utils.ts
@@ -1,7 +1,11 @@
 import { toBase64 } from 'lib0/buffer.js';
+import { digest } from 'lib0/hash/sha256';
 
 export async function sha(input: ArrayBuffer): Promise<string> {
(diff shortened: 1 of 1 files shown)


darkskygit:
i don't think it's a good idea to maintain a hash implementation ourself
i think this library is more suitable: https://github.com/Daninet/hash-wasm , what do you think?

fourdim:
What about lib0? We have already introduced it as a dependency. 
https://github.com/dmonad/lib0/blob/main/hash/sha256.js

doodlewind (commented):
I'd say the idea using lib0 looks pretty nice!

Saul-Mirone (approved):

Saul-Mirone:
The changes look good to me but I prefer the idea using lib0.

Brooooooklyn (changes requested):
Let's use the impl in `lib0`

darkskygit (approved):

doodlewind (approved):

PR #6457, opened 2024-03-12
fix(edgeless): add paragraph block when note is empty

### LINKED ISSUE #6455: Can add block when last block is removed from note
https://github.com/toeverything/blocksuite/assets/13579374/da337c4e-5cce-4a20-9cbb-f917ece8ea7d

donteatfriedrice:
To close #6455 

https://github.com/toeverything/blocksuite/assets/99816898/6349a733-9f2b-4056-aa2f-790ef1a915d9


doodlewind (commented):
Looks good but this seems to be a good edge case to cover test!

PR #6461, opened 2024-03-12
fix(edgeless): drag note to merge after a list item

doouding:
Closes: [BS-56](https://linear.app/affine-design/issue/BS-56/白板上，note-合并的功能坏了)


doouding:
> Can we add a test for this case?

Got it.

donteatfriedrice (approved):

Saul-Mirone (changes requested):
Wait for test case

doodlewind (approved):

Saul-Mirone (approved):

PR #6462, opened 2024-03-12
fix: remote cursor shouldn't display on other docs

Saul-Mirone:
Closes: [BS-165](https://linear.app/affine-design/issue/BS-165/whiteboard-出现其他用户光标)

### DIFF
--- a/packages/framework/store/src/yjs/awareness.ts
+++ b/packages/framework/store/src/yjs/awareness.ts
@@ -18,7 +18,8 @@ export type RawAwarenessState<
   user?: UserInfo;
   color?: string;
   flags: Flags;
-  selection: Record<string, UserSelection>;
+  // use v2 to avoid crush on old clients
+  selectionV2: Record<string, UserSelection>;
 };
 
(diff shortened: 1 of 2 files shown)


packages/framework/store/src/yjs/awareness.ts
```diff
@@ -9,14 +9,16 @@ export interface UserInfo {
...
+
 // Raw JSON state in awareness CRDT
 export type RawAwarenessState<
   Flags extends Record<string, unknown> = BlockSuiteFlags,
 > = {
   user?: UserInfo;
   color?: string;
   flags: Flags;
-  selection: Array<Record<string, unknown>>;
+  selection: Record<string, UserSelection>;
```
EYHN, line 21: Could this change cause the old version to crash? It is better to use `v2` field?

Saul-Mirone: Done

EYHN (approved):

doodlewind (approved):

PR #6463, opened 2024-03-12
fix(edgeless): dropout embed synced doc block should fit height

donteatfriedrice:
To fix [AFF-613](https://linear.app/affine-design/issue/AFF-613/dropped-embedded-sync-doc-in-edgeless-doesnt-fit-in-height)

https://github.com/toeverything/blocksuite/assets/99816898/ec81c5b1-d9e6-4d7f-8e64-be52ec3e38e9

Add some tests, part of #6357


Saul-Mirone:
Great! Can we add some tests for this case?

donteatfriedrice:
> Great! Can we add some tests for this case?

Sure, I will add some tests for embed synced doc.

Saul-Mirone (changes requested):
Wait for test case

Saul-Mirone (approved):

PR #6493, opened 2024-03-17
feat(edgeless): support pie menu widget

golok727:
A pie menu inspired by Blender's interface. 

For preview I have added a tools menu for edgeless the trigger is 'q'.

> tests are wip...
# Usage

- Trigger Key: Each menu is associated with a trigger key to open it. To activate a menu, press and hold the trigger key.

- Navigation: While holding the trigger key, move the cursor towards the desired action within the pie menu. Hovering over a wedge of the pie menu highlights that action. Release the trigger key to select the highlighted action.

- Submenu: Submenus can be accessed by hovering over the submenu node (with a blue dot) within the pie menu. To close a submenu just hover over the desired center node

- Keeping Menu Open: If the trigger key is released before the "SELECT_ON_RELEASE_TIMEOUT" (150ms default) threshold, the pie menu remains open and you can select with your mouse.

- Shortcut Selection: The numeric keys (0-9) correspond to actions within the pie menu. Pressing a numeric key selects the pie node associated with that number directly.

I have also provided a pie builder api to build pie menus layouts easier. Looking forward for help to make the api better. 

# Preview
https://github.com/toeverything/blocksuite/assets/123532141/5fd90b6b-ea1b-466f-84ae-2c32afa09bdd

Pie-preview3
Pie-preview2
Pie-preview1

#### Color Node
image
image

https://github.com/toeverything/blocksuite/assets/123532141/4d7c0847-fbe1-4d8a-9ef0-97f4b71d75ba


golok727:
>Hey there! While working on it I used the hover color (--affine-hover-color) but it had some transparency, which didn't look right when hovering over it in the current setup. so for now I just used a hardcoded value. I will fix it 👍🏻

@hwangdev97 take a look at this hover color in the light theme this will look nice and the dark theme is fine with the current color. could you help me locate the file where you set the CSS variables? 


image



#### Using `--affine-hover-color`
image
image

golok727:
Hey, @doodlewind I need some help!

1. What's the best z-index variable to ensure the menu is always on top of pop-ups? Here's a snapshot for reference: image
image

2. Do we need a flag to enable or disable this menu for users who don't want it? 

3. Edgeless needs this menu the most, but I have added support in doc mode too (eg: switch between page and edgeless, block commands, etc..). I don't know how relevant it is but if you say we can remove it.

doodlewind (commented):
@golok727 thanks again for the great work! I've seen previous comment regarding feature flags and some best practice, and here's the advice:

* It's really a smooth and polished feature to try, so landing this without feature flag should be fine - we may fallback if needed but no worries.
* Limiting the usage of this widget in edgeless mode should be expected.

I think @Saul-Mirone would land this after some style checks.

Thanks again for your feat. If you are interested in BlockSuite, AFFiNE and related contributions, feel free to DM me: doodlewind at toeverything dot info

packages/blocks/src/_specs/_specs.ts
```diff
@@ -61,6 +62,7 @@ const DocPageSpec: BlockSpec<PageRootBlockWidgetName> = {
       //   AFFINE_BLOCK_HUB_WIDGET
       // )}`,
       [AFFINE_MODAL_WIDGET]: literal`${unsafeStatic(AFFINE_MODAL_WIDGET)}`,
+      [AFFINE_PIE_MENU_WIDGET]: literal`${unsafeStatic(AFFINE_PIE_MENU_WIDGET)}`,
```
Saul-Mirone, line 65: This spec is used by affine and we need the designers and pms review this feature. So, maybe we can't add them here for now. Play we can enable them in our playground :)

Saul-Mirone (changes requested):
Thanks for your contribution!

packages/blocks/src/_specs/_specs.ts
golok727: Hey, @doodlewind mentioned it. 
> Limiting the usage of this widget in edgeless mode should be expected.
I will fix it. Thanks!

golok727:
@Saul-Mirone, just to clarify, you mentioned removing it from both the DocSpec and EdgelessSpec, correct? I've commented it out as per your instructions. Additionally, as @doodlewind pointed out, since this is an edgeless-only feature, I've removed its support in Page. PTA👀

doodlewind (commented):
@golok727 you can search for `enable_synced_doc_block` to see how feature flags is being used in the codebase. Adding a `enable_pie_menu` flag and test this flag on the key handler for `q` should be adequate. Also in this way you can enable the flag by default in BlockSuite playground, since there are `defaultFlags` that can be used to override.

golok727:
> @golok727 you can search for `enable_synced_doc_block` to see how feature flags is being used in the codebase. Adding a `enable_pie_menu` flag and test this flag on the key handler for `q` should be adequate. Also in this way you can enable the flag by default in BlockSuite playground, since there are `defaultFlags` that can be used to override.

Thanks! I added the flag...

Saul-Mirone (approved):

PR #6508, opened 2024-03-19
feat(edgeless): add mindmap support

doouding:
### Notable change
- Add new `mindmap` surface element
- Add new `GroupLikeModel` abstract class, `group` and `mindmap` element are derived from this
- Add new `StatelessModel` class, `StatelessModel` is used to render elements that do not exist in the yjs data, and they are not interactive, only visible in canvas.
- Add `local` property in surface slot payload


packages/blocks/src/root-block/edgeless/components/auto-complete/edgeless-auto-complete.ts
```diff
@@ -306,6 +308,28 @@ export class EdgelessAutoComplete extends WithDisposable(LitElement) {
     this.removeOverlay();
   }
 
+  private _addMindmapNode(direction: LayoutType.LEFT | LayoutType.RIGHT) {
+    const mindmap = this.current.group;
+
+    if (!(mindmap instanceof MindmapElementModel)) return;
+
+    const newNode = mindmap.addNode(
```
Saul-Mirone, line 316: We should use object to replace this long arguments list.

Saul-Mirone (approved):

PR #6510, opened 2024-03-19
fix: create linked doc from block selection loss data

### LINKED ISSUE #6487: 'Create Linked Doc' looses data​
### What happened?

On a page: after creating some content, I wanted to extract to a page via the context menu (see first screenshot)..
Screenshot from 2024-03-15 00-58-42

... I realized, that only the first level bullet points arrived. The other data was lost (see second screenshot). In another example, a quote was also lost.
Screenshot from 2024-03-15 00-59-05

### Distribution version

Linux

_No response_

### Are you self-hosting?

- [ ] Yes

### Relevant log output

```shell
Affine Version: 0.12.2
Editor Version: 0.12.0-canary-202402220729-0868ac6
```

### Anything else?

_No response_

donteatfriedrice:
To fix #6487

### DIFF
--- a/packages/blocks/src/root-block/commands/model-crud/get-selected-models.ts
+++ b/packages/blocks/src/root-block/commands/model-crud/get-selected-models.ts
@@ -12,6 +12,7 @@ export const getSelectedModelsCommand: Command<
 > = (ctx, next) => {
   const types = ctx.types ?? ['block', 'text', 'image'];
   const mode = ctx.mode ?? 'flat';
+  const selectedModels: BlockModel[] = [];
   ctx.std.command
     .chain()
     .tryAll(chain => [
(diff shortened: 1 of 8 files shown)


packages/blocks/src/root-block/commands/model-crud/get-selected-models.ts
```diff
@@ -7,9 +7,11 @@ export const getSelectedModelsCommand: Command<
   'selectedModels',
   {
     types?: Extract<BlockSuite.SelectionType, 'block' | 'text' | 'image'>[];
+    mode?: 'all' | 'flat' | 'highest';
```
doodlewind, line 10: Since this should be part of our public commands API, could you clarify the mode definition in the `/** ... */` comment?

doodlewind (approved):
Should land this and consider the mode clarification afterwards.

PR #6534, opened 2024-03-22
refactor: rewrite view store

Saul-Mirone:


doodlewind (approved):
Good to see the refactoring! Any description about the refactoring or its corresponding API updates?

PR #6537, opened 2024-03-22
fix(edgeless): copyAsPng on shape with rotation has cutoff edges

golok727:
# Fix
- Bounds of shapes copied should have rotation applied
- Background of copied image  should be transparent

# Before
image

# After
image


doodlewind (approved):
Thanks for the fix, tested and I think it's good to make the background transparent for space area in edgeles by default.

PR #6538, opened 2024-03-22
fix(edgeless): copyAsPng failed to copy on image blocks

golok727:
# Issue 
Converting ImgSrc with SVG was causing issue
- Fixes failedToCopyToClipboard error


donteatfriedrice:
Remove the code will cause copy as png  incorrectly when a svg inside a note.

original:
Screenshot 2024-03-22 at 14 30 15

result:
Screenshot 2024-03-22 at 14 30 34

doodlewind (changes requested):
I'm afraid this will lead to regression of #6146

golok727:
> I'm afraid this will lead to regression of #6146

Or filter out the top level images and render it directly on the canvas?

donteatfriedrice:
> > I'm afraid this will lead to regression of #6146
> 
> Or filter out the top level images and render it directly on the canvas?

I think we should figure out why the fetchImage function failed. When the top level image is a svg, the copy as png function still not works correctly.

Screenshot 2024-03-22 at 14 57 42

donteatfriedrice:
> > I'm afraid this will lead to regression of #6146
> 
> Or filter out the top level images and render it directly on the canvas?

I think changing the replaceImgSrcWithSvg in exportManager function to an arrow function can solve this problem. Would you be willing to help make this change?

golok727:
> > > I'm afraid this will lead to regression of #6146
> > 
> > 
> > Or filter out the top level images and render it directly on the canvas?
> 
> I think changing the replaceImgSrcWithSvg in exportManager function to an arrow function can solve this problem. Would you be willing to help make this change?

sure!

donteatfriedrice:
> > > > I'm afraid this will lead to regression of #6146
> > > 
> > > 
> > > Or filter out the top level images and render it directly on the canvas?
> > 
> > 
> > I think changing the replaceImgSrcWithSvg in exportManager function to an arrow function can solve this problem. Would you be willing to help make this change?
> 
> sure!

Thanks!

golok727:
@donteatfriedrice Changing to arrow fn resolves the `copyAsPng` function.
@doodlewind @donteatfriedrice. Take a look at the copied image below. There is a lot of compression which is needed when multiple objects are selected. What about the case where a user wants to copy a single image wouldn't it be nice to copy the same image data to the clipboard with the same quality and also, remove the padding in this case 


image
image

golok727:
@doddlewind Image quality is fixed.  No need for `ctx.scale(dpr, dpr) ` which was blurring the image. Scaling the width and height by dpr is enough. Also, I changed the 100px padding I don't think it is relevant. figma and other tools do not have that padding. If you think we need it we can add it back

donteatfriedrice:
> Image quality is fixed. No need for `ctx.scale(dpr, dpr) ` which was blurring the image. Scaling the width and height by dpr is enough. Also, I changed the 100px padding I don't think it is relevant. figma and other tools do not have that padding. If you think we need it we can add it back

It seems that the note or other elements copy as png result will be not that good if remove `ctx.scale(dpr, dpr) ` 

Screenshot 2024-03-22 at 17 06 21

golok727:
> > Image quality is fixed. No need for `ctx.scale(dpr, dpr) ` which was blurring the image. Scaling the width and height by dpr is enough. Also, I changed the 100px padding I don't think it is relevant. figma and other tools do not have that padding. If you think we need it we can add it back
> 
> It seems that the note or other elements copy as png result will be not that good if remove `ctx.scale(dpr, dpr) `
> 
Sorry, my bad. When I was writing a renderer for an engine I did the same thing `ctx.scale(dpr, dpr) ` which made the images blurry so I thought that would be the issue. But it is not, when a Top-level image is scaled and then copied it is getting blurred. with the initial scale, it does not have that issue.

donteatfriedrice (approved):
LGTM. Thanks for the fix.

golok727:
@doodlewind PTAL

doodlewind (approved):
Thanks for the fix!

PR #6540, opened 2024-03-22
feat(edgeless): paste SVG in text/plain mime

golok727:
# Changes
- This PR allows to paste SVG in text format from the clipboard

# Usage
- Copy as SVG from tools like Figma and directly paste it into edgeless


golok727:
@doodlewind  do we need sanitization? as scripts in SVG usually do not get executed.

Saul-Mirone (dismissed):
Please add integration test for this case.

fourdim:
> @doodlewind do we need sanitization? as scripts in SVG usually do not get executed. Also we are loading it as image src

Yeah, we do need to sanitize it.

golok727:
> > @doodlewind do we need sanitization? as scripts in SVG usually do not get executed. Also we are loading it as image src
> 
> Yeah, we do need to sanitize it

Is a simple sanitization like below enough or does blocksuite uses any specific tool? haven't found anything yet.

```ts
const attributes = Array.from(svgEl.attributes).map(({ name }) => name)
 const hasScriptAttr = !!attributes.find((attr) => attr.startsWith('on'))
 const disallowedSvgElements = svgEl.querySelectorAll(svgDisallowed.join(','))
  return disallowedSvgElements.length === 0 && !hasScriptAttr ? svg : null
```

fourdim:
> > > @doodlewind do we need sanitization? as scripts in SVG usually do not get executed. Also we are loading it as image src
> > 
> > 
> > Yeah, we do need to sanitize it
> 
> Is a simple sanitization like below enough or does blocksuite uses any specific tool? haven't found anything yet.
> 
> ```ts
> const attributes = Array.from(svgEl.attributes).map(({ name }) => name)
>  const hasScriptAttr = !!attributes.find((attr) => attr.startsWith('on'))
>  const disallowedSvgElements = svgEl.querySelectorAll(svgDisallowed.join(','))
>   return disallowedSvgElements.length === 0 && !hasScriptAttr ? svg : null
> ```

You can use https://www.npmjs.com/package/dompurify.

golok727:
@fourdim PTAL

fourdim (approved):

Issue #6541, opened 2024-03-23
Support `Shift` + `Arrow keys` to move elements on Edgeless


zanwei:
Please refer the video down below

Would love use `Shift` + `Arrow keys` to move element on the Edgeless mode 10 pixels each time

https://github.com/toeverything/blocksuite/assets/40333372/9aee8d01-6a43-49c4-8702-2687b103d93a

PR #6544, opened 2024-03-23
feat(edgeless): shift + arrow keys to move elements with 10px inc

golok727:
Closes https://github.com/toeverything/blocksuite/issues/6541


packages/blocks/src/root-block/edgeless/edgeless-keyboard.ts
```diff
@@ -247,6 +236,33 @@ export class EdgelessPageKeyboardManager extends PageKeyboardManager {
       { global: true }
     );
   }
+  private _bindArrowKeys() {
```
Saul-Mirone, line 239: Please stick to `bindHotkey` instead of use `handleEvent`.

Saul-Mirone (changes requested):

Saul-Mirone (approved):

PR #6553, opened 2024-03-24
refactor(edgeless): abstract pie menu rendering logic

golok727:
# Changes
- Abstracts all the rendering logic in the `PieNode` class
- Remove Unnecessary comments
- Viewport should not pan when the pie menu is open


packages/blocks/src/root-block/edgeless/edgeless-root-block.ts
```diff
@@ -765,10 +766,9 @@ export class EdgelessRootBlockComponent extends BlockElement<
     );
   }
   private _isPieMenuOpen() {
-    return false; // TODO add it back later
-    // return (
-    //   this.widgetElements['affine-pie-menu-widget'] as AffinePieMenuWidget
-    // ).isOpen;
+    return (
+      this.widgetElements['affine-pie-menu-widget'] as AffinePieMenuWidget
+    ).isOpen;
```
doouding, line 771: I don't think this is necessary. Looks like you're using this method to determine whether to execute the wheel event handler. Am I right?
Since the widget is rendered inside the `edgeless-root-block`, you could just call `e.stopPropagation()` in the pie-menu widget wheel handler, preventing the event from propagating to the upper element.

packages/blocks/src/root-block/widgets/pie-menu/base.ts
```diff
@@ -61,21 +61,13 @@ export interface IPieCommandNode extends IPieBaseNode {
...
 
-export type IPieNodeWithAction = IPieCommandNode; // | IPieToggleNode;
-
-export type IPieNonRootNode = IPieCommandNode | IPieColorNode | IPieSubmenuNode;
-// | IPieToggleNode;
-
-export type IPieNode = IPieRootNode | IPieNonRootNode;
-
-// For a color picker sub menu
 export interface IPieColorNode extends IPieBaseNode {
```
doouding, line 71: Although this is an interface, we usually don't prefix with the letter 'I'. `PieColorNode` is fine.

doouding (changes requested):

doouding (approved):
Thanks for the work :)

PR #6560, opened 2024-03-25
feat(database): column stats

golok727:
image
image
image
image
image


golok727:
Hey @zzj3720, mind lending a hand?
I noticed something while working with the column.onCellUpdate event handler. It seems like the event is triggered for all columns in the database, instead of just the one where the cell was updated. Is this intentional or possibly a bug?

I need this to only trigger for the specific column being updated so I can recalculate data properly. Do you think we can adjust the API to make sure the event only fires for the updated column?


https://github.com/toeverything/blocksuite/assets/123532141/3f5b447c-6821-431b-8bd7-8f882a49e6c8

packages/blocks/src/database-block/table/types.ts
```diff
@@ -12,7 +12,42 @@ export interface Column<
   type: ColumnType;
   name: string;
   data: Data;
+  statCalcOp: StatCalcOpType;
```
golok727, line 15: @zzj3720 I need to make a migration right ?

golok727:
@zzj3720 Hey some help🙏🏻, why does datasource.onCellUpdate is inconsistent with different property types? rich text and title is using ytext observe and others are using model.propsChanges in data source. due to this the callback given to onCellUpdate is called for all the columns when a cell is updated.
Is this intended behaviour as I only want to compute the stats for the updated col

packages/blocks/src/database-block/table/types.ts
zzj3720: We'll try to avoid migration now, so this should be an optional value

zzj3720:
> @zzj3720 Hey some help🙏🏻, why does datasource.onCellUpdate is inconsistent with different property types? rich text and title is using ytext observe and others are using model.propsChanges in data source. due to this the callback given to onCellUpdate is called for all the columns when a cell is updated. Also is it not getting called for richtext cells. Is this intended behaviour as I only want to compute the stats for the updated columns

Listening for updates on columns is challenging; I need to think about it.

zzj3720:
@golok727 I think we can temporarily set aside the issue of listening for updates on the title column and fix it later. If the statistics functionality is completed, we can add a feature flag to merge it first and then optimize it. I might be doing some adjustments and refactoring to the directory structure soon, so if this task drags on too long, it might be difficult to merge.

golok727:
> @golok727 I think we can temporarily set aside the issue of listening for updates on the title column and fix it later. If the statistics functionality is completed, we can add a feature flag to merge it first and then optimize it. I might be doing some adjustments and refactoring to the directory structure soon, so if this task drags on too long, it might be difficult to merge.

The core functionality is finished; I'll refine it by making the statCalcOp optional, integrating the feature flag as discussed, and ensuring the PR is ready for review by this evening. Thanks!

Saul-Mirone:
Waiting for @zzj3720 to review this.

zzj3720 (approved):

PR #6563, opened 2024-03-25
fix(database): slow first rendering of the Database Block

zzj3720:
Avoid unnecessary contenteditable assignments

before:
image

after:
image


Flrande (approved):

zkwolf:
This change is so obvious, I hava a large data table and the render time from 12s to 1s.

PR #6622, opened 2024-03-28
fix(edgeless): general shape for triangle

RubaXa:
**Problem**
Now, when you select a `triangle`, a `rect` is drawn under the cursor


RubaXa:
> Good catch! Could you sign the `CLA.md` to start the following landing process? Thanks!

✍️ Done

doodlewind (approved):
Thanks for the fix!

PR #6630, opened 2024-03-28
feat: add draft model for transformer

Saul-Mirone:


packages/framework/store/src/transformer/base.ts
```diff
@@ -38,7 +39,7 @@ export class BaseBlockTransformer<Props extends object = object> {
     ) as Props;
   }
 
-  protected _propsToSnapshot(model: BlockModel) {
+  protected _propsToSnapshot(model: DraftModel) {
```
doodlewind, line 42: What's the difference between `BlockModel` and `DraftModel`?

Saul-Mirone: > What's the difference between `BlockModel` and `DraftModel`?

`BlockModel` is readonly, the changes to it should sync with yjs data. However, `DraftModel` is mutable and it won't be connected with yjs data. Which is useful for temporary transformations in transformer.

doodlewind (approved):

PR #6632, opened 2024-03-28
fix(edgeless): pie menu update shape overlay

golok727:
Updates shape overlays


doodlewind (approved):
Thanks for the update!

Issue #6585, opened 2024-03-27
Use Shift + ToolShortCut to toggle between grouped tools in edgeless


golok727:
Could we add Shift key shortcuts for tool suboptions, similar to Photoshop? For example:
Eg Pressing 'Shift' + 'L' toggles between types of connectors.

PR #6634, opened 2024-03-28
feat(edgeless): shift + s to toggle between shapes

golok727:
Close https://github.com/toeverything/blocksuite/issues/6585

https://github.com/toeverything/blocksuite/assets/123532141/b70d1c8a-ef22-47aa-941f-d4f8b8e50c42


Saul-Mirone (changes requested):
Please add integration test for this feature.

Saul-Mirone (approved):

PR #6672, opened 2024-04-01
refactor(store): add block collection

Saul-Mirone:


doodlewind (commented):
Should this be considered as a notable update, or implementation details?

PR #6675, opened 2024-04-02
feat(edgeless): mindmap gen

doouding:
continue https://github.com/toeverything/blocksuite/pull/6508


package.json
```diff
@@ -52,8 +52,8 @@
     "postinstall": "husky"
   },
   "lint-staged": {
-    "*": "prettier --write --cache --ignore-unknown",
-    "*.{ts,tsx,js,jsx}": "eslint --cache --fix"
+    "*": "pnpm exec prettier --write --cache --ignore-unknown",
+    "!(examples/**/*).{ts,tsx,js,jsx}": "pnpm exec eslint --cache --fix"
```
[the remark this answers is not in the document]
doouding: Example folder has already been configured ignored in the eslint config file but not in lint-staged. So eslint will still check the folder when you try to commit change but it will always fail because of some plugins' missing.

packages/blocks/src/surface-block/mini-mindmap/page-root-block.ts
```diff
@@ -0,0 +1,30 @@
+import { BlockElement } from '@blocksuite/block-std';
+import { html } from 'lit';
+import { customElement } from 'lit/decorators.js';
+
+import type { RootBlockModel } from '../../root-block/root-model.js';
+
+@customElement('mini-mindmap-root-block')
+export class MindmapRootBlock extends BlockElement<RootBlockModel> {
```
Saul-Mirone, line 8: This file should be named as `mindmap-root-block`

packages/presets/tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "outDir": "./dist/",
     "noEmit": false
   },
-  "include": ["./src"],
+  "include": ["src/**/*"],
```
Saul-Mirone, line 8: I think the `include` property are all set to `./src` in all tsconfig files. Why do we need to change for this package?

doouding: Oh, I will delete this.

packages/blocks/src/root-block/widgets/ai-panel/ai-panel.ts
```diff
@@ -24,7 +24,10 @@ import type {
 } from './components/index.js';
 
 export interface AffineAIPanelWidgetConfig {
-  answerRenderer: (answer: string) => TemplateResult<1>;
+  answerRenderer: (
+    answer: string,
+    state: AffineAIPanelState
```
Flrande, line 29: Will it influence action refactor? @pengx17

packages/blocks/src/root-block/widgets/ai-panel/ai-panel.ts
```diff
@@ -129,7 +134,6 @@ export class AffineAIPanelWidget extends WidgetElement {
     assertExists(text);
     assertExists(this.config.generateAnswer);
 
-    this._resetAbortController();
```
Flrande, line 132: Why we delete them (L132 and L147)

packages/blocks/src/root-block/widgets/ai-panel/ai-panel.ts
pengx17: maybe not. the provider does not care about how the answer is rendered

packages/blocks/src/root-block/widgets/ai-panel/ai-panel.ts
doouding: Example folder has already been configured ignored in the eslint config file but not in lint-staged. So eslint will still check the folder when you try to commit change but it will always fail because of some plugins' missing.

doouding: Because this will disable the `autoUpdate`, and the AiPanel would then stop moving along with the reference element.

Flrande (approved):

PR #6686, opened 2024-04-03
fix: placeholder visibility and cursor style in readonly mode

L-Sun:
- In read-only mode, placeholder need to be hidden
- In read-only mode, the cursor style of `inline-editor` need to be `"default"`, and `v-text` need to be `"text"`
- Add readonly check for `RootService.appendParagraph()`.

placeholder_cursor


Flrande (approved):

c0sc0s:
hi~ how to use read-only mode, i can't find id in docs

Flrande:
> hi~ how to use read-only mode, i can't find id in docs

`collection.awarenessStore.setReadonly(doc, true)`

c0sc0s:
thanks

PR #6691, opened 2024-04-04
fix(database): soft enter removes next character

golok727:
https://github.com/toeverything/blocksuite/assets/123532141/b507b3a2-32b9-49de-a228-d63f22b80edb

It was using window.length instead of inlineRange.length

### DIFF
--- a/packages/blocks/src/database-block/common/columns/rich-text/cell-renderer.ts
+++ b/packages/blocks/src/database-block/common/columns/rich-text/cell-renderer.ts
@@ -340,7 +340,7 @@ export class RichTextCellEditing extends BaseCellRenderer<Text> {
 
       this.column.captureSync();
       const text = new Text(this.inlineEditor.yText);
-      text.replace(inlineRange.index, 0, '\n');
+      text.replace(inlineRange.index, inlineRange.length, '\n');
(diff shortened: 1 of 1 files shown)


packages/blocks/src/database-block/common/columns/rich-text/cell-renderer.ts
```diff
@@ -340,7 +340,7 @@ export class RichTextCellEditing extends BaseCellRenderer<Text> {
 
       this.column.captureSync();
       const text = new Text(this.inlineEditor.yText);
-      text.replace(inlineRange.index, length, '\n');
+      text.replace(inlineRange.index, 0, '\n');
```
Flrande, line 343: The situation in the video should have a length of 0, so it may not be the source of the error.

Flrande (changes requested):

golok727:
> How to reproduce it?
> 
>  CleanShot.2024-04-04.at.18.18.54.mp4

It is not working in Chrome. Tested it in Edge and it is working 

https://github.com/toeverything/blocksuite/assets/123532141/7caec720-94bf-43d4-a668-aad2b417ddcd

Flrande (approved):

Issue #6666, opened 2024-04-01
Importing markdown files, the page title always displays untitled


EYHN:
It was fixed before, but seems to break again.

https://github.com/toeverything/AFFiNE/issues/4906

Tzyito:
I think this feature is ambiguous(Does the import support only newly created doc, or does it support insertion into an existing doc?) and I have found the problem. If you are interested, I would like to discuss my views with you

PR #6702, opened 2024-04-08
fix(page): the imported markdown file name is displayed by default

Tzyito:
fix https://github.com/toeverything/blocksuite/issues/6666 & https://github.com/toeverything/AFFiNE/issues/6460

This will make the file as doc default name when uploading markdown files. But I’m not sure if I need to do this when uploading other types of files.

(btw: One thing I'm confused about is whether these two buttons have the same effect. 
image


fourdim:
No worries. CC @doodlewind to merge it manually.

packages/blocks/src/root-block/widgets/linked-doc/import-doc/import-doc.ts
```diff
@@ -242,14 +257,15 @@ export class ImportDoc extends WithDisposable(LitElement) {
     const pageIds: string[] = [];
     for (const file of files) {
       const text = await file.text();
+      const [fileName] = file.name.split('.');
```
CatsJuice, line 260: Maybe `file.name.split('.').slice(0, -1).join('.')` would be more accurate?

Tzyito: thanks for your replay!

Issue #6708, opened 2024-04-09
【bug】When exporting pdf or png, the hover selection style will appear.


Tzyito:
now: 
<img width="1498" alt="image" src="https://github.com/toeverything/blocksuite/assets/37643993/546d9ad0-321f-4a01-83ea-8150d5a0f06f">

As you can see, when I export after selecting it with the mouse, some unnecessary background colors and toolbars will be generated. I think this should be hidden.

This confusion exists in both page and edgeless modes. Influence points: export pdf/png, clipboard export png

PR #6711, opened 2024-04-09
fix(blocks): ignore formart-bar-widget&block-selection tags when exporting png/pdf

Tzyito:
fix https://github.com/toeverything/blocksuite/issues/6708


donteatfriedrice (approved):
Thanks for the fix.

PR #6714, opened 2024-04-09
feat(edgeless): edgeless copilot actions

doouding:
continue https://github.com/toeverything/blocksuite/pull/6620

### Add new actions
- findActions
- brainstormMindmap
- expandMindmap
- brainstormMindmap
- makeItReal
- createSlides
- explain


doouding:
@pengx17 I listed some new added actions in the PR description.

packages/presets/src/ai/config/edgeless-copilot.ts
```diff
@@ -1,317 +1,172 @@
...
-          },
-        ],
-        actions: [],
-      },
-      errorStateConfig: {
-        upgrade: () => {},
-        responses: [],
-      },
-    };
+import type { AIItemGroupConfig } from '@blocksuite/blocks';
```
pengx17, line 1: I think @Flrande mentioned that the edgless config should be located at `ai/entries/edgeless`

Flrande: We can move item components from blocks to preset

Saul-Mirone (approved):

packages/presets/src/ai/config/edgeless-copilot.ts
Flrande: Entries subfolders correspond to setup functions for each ui entry (such as listening for events to trigger panel when a space is entered)

Flrande (approved):

PR #6716, opened 2024-04-09
fix(edgeless): lasso tool should consider rotation and fallback to ray-casting

golok727:
Provides a better selection algorithm for the lasso tool.

https://github.com/toeverything/blocksuite/assets/123532141/50d41fc6-d522-4c8a-be1a-05081bc99b71


golok727:
@doouding PTAL

doodlewind (approved):
Thanks for the fix!

PR #6728, opened 2024-04-10
feat(presets): ai images

fundon:
Related to #6669

### DIFF
--- a/packages/playground/apps/default/specs-examples/ai/provider.ts
+++ b/packages/playground/apps/default/specs-examples/ai/provider.ts
@@ -19,7 +19,7 @@ export function setupAIProvider() {
     Summarize the key points from the following content in a clear and concise manner,
     suitable for a reader who is seeking a quick understanding of the original content.
     Ensure to capture the main ideas and any significant details without unnecessary elaboration:
-  
+
     ${options.input}
     `;
     return textToTextStream({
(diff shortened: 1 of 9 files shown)


packages/presets/src/ai/config/edgeless-copilot.ts
```diff
@@ -308,7 +314,104 @@ const chatWithAI: AIItemConfig = {
...
+        stream: true,
+        docId: host.doc.id,
+        workspaceId: host.doc.collection.id,
+        attachments: [png],
+      });
+      bindEventSource(stream, { update, finish, signal });
+    };
+    // const selectionRect = copilotPanel.selectionModelRect;
+
+    aiPanel.config = {
```
pengx17, line 380: `AffineAIPanelWidgetConfig` is initialized one time at https://github.com/toeverything/blocksuite/pull/6728/files#diff-19bc5e6df24f77683ae5a0429af2241650ac3c24af898f92e78da8b46f826a26R40

I think we need to reconsider how the config can be used in both edgeless and doc mode

fundon: Instead of exposing config, we can provide a method to configure it.

packages/presets/src/ai/ai.ts
```diff
@@ -39,10 +39,12 @@ export function getAISpecs() {
               if (view.component instanceof AffineAIPanelWidget) {
                 view.component.config = buildAIPanelConfig(view.component);
                 setupSpaceEntry(view.component);
+                return;
```
Flrande, line 42: Why return early

Flrande (approved):

pengx17 (approved):

PR #6744, opened 2024-04-12
fix(presets): ai response insert and replace

donteatfriedrice:
- [x] Optimize selected text content passed to ai panel input.
- [x] AI response: insert below and replace selection. 

https://github.com/toeverything/blocksuite/assets/99816898/c3a1dc38-a10e-4cf2-8253-792691971012

https://github.com/toeverything/blocksuite/assets/99816898/bd441817-a952-46dd-8885-c4af7fa717a1


pengx17 (approved):

packages/presets/src/ai/ai-panel.ts
```diff
@@ -10,57 +10,81 @@ import {
...
+      });
+
+      const models = await insertFromMarkdown(
+        host,
+        panel.answer ?? '',
+        firstBlockParent.model.id,
+        firstIndex
+      );
+
+      host.updateComplete
```
Saul-Mirone, line 81: why don't use `await`?

PR #6759, opened 2024-04-15
feat: add slide action

regischen:


packages/presets/src/ai/slides/template.ts
```diff
@@ -20,8 +20,7 @@ const getImageUrlByKeyword =
   (keyword: string) =>
   async (w: number, h: number): Promise<string> => {
     const url = new URL('https://api.unsplash.com/search/photos');
-    const params = new URLSearchParams(window.location.search);
-    const unsplashKey = params.get('unsplashKey');
+    const unsplashKey = 'BONl2qMSdMLIexsUIOmi7Bh2dSO8hBUZQq2sUR0E4Jc'; //params.get('unsplashKey');
```
pengx17, line 23: this should be provided by AIProvider as well

regischen: yeah, it does. now it's just a workaround.

pengx17 (approved):

packages/presets/src/ai/slides/template.ts
pengx17: I will add a follow up fix later today (new service in AIProvider)

PR #6776, opened 2024-04-16
fix(blocks): page movement due to scrolling into view

fundon:
Closes: AFF-841


pengx17:
@EYHN taught me a trick to use `overflow: clip` instead of `hidden` to get rid of unwanted page scrolling. Maybe you can try that instead

pengx17 (approved):

PR #6794, opened 2024-04-18
fix(blocks): optimize copilot tool user experience

fundon:
Closes: AFF-868

* [x] update selection rectangle when selecting 
* [x] keep copilot panel in view and fix duplicate creation issue
* [x] in copilot tool mode, lock whiteboard movement, zoom, disable editing shortcuts, etc.

https://github.com/toeverything/blocksuite/assets/27926/7a7f9edb-b9c0-4d0a-866b-0abe0b469ab8

### DIFF
--- a/packages/blocks/src/root-block/edgeless/edgeless-keyboard.ts
+++ b/packages/blocks/src/root-block/edgeless/edgeless-keyboard.ts
@@ -13,6 +13,7 @@ import {
 } from '../../surface-block/index.js';
 import { EdgelessBlockModel } from '../edgeless/type.js';
 import { PageKeyboardManager } from '../keyboard/keyboard-manager.js';
+import { CopilotSelectionController } from './controllers/tools/copilot-tool.js';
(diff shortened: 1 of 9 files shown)


packages/blocks/src/root-block/edgeless/edgeless-keyboard.ts
```diff
@@ -86,6 +87,7 @@ export class EdgelessPageKeyboardManager extends PageKeyboardManager {
           });
         },
         m: () => {
+          if (this.rootElement.service.tool.locked) return;
```
doouding, line 90: Checking for a lock each time seems a bit inelegant. I'm wondering if we could add a lock attribute to the viewport class, allowing calls to fail silently if it's locked.

fundon: I moved it to `service`.

doouding (approved):

PR #6803, opened 2024-04-18
fix(blocks): switch the file-type to @sgtpooki/file-type

lawvs:
Fix https://github.com/toeverything/AFFiNE/issues/6603

This pull request switches the [file-type](https://github.com/sindresorhus/file-type) to [@sgtpooki/file-type](https://github.com/sgtpooki/file-type).

It also includes other minor changes and updates.

~~The `file-type` is no longer useful since the https://github.com/toeverything/AFFiNE/issues/3245 resolved in https://github.com/toeverything/AFFiNE/issues/5565~~

---

More context see https://github.com/toeverything/AFFiNE/issues/3245 https://github.com/toeverything/blocksuite/pull/3555 https://github.com/toeverything/blocksuite/pull/4818 https://github.com/toeverything/blocksuite/pull/5372 https://github.com/sindresorhus/file-type/issues/578

---

## Caveat

The [@sgtpooki/file-type](https://github.com/sgtpooki/file-type) is an unofficial fork package from the community. It may need to be source code audited before updating.

Also, thanks to sgtpooki for their strong contribution to @sgtpooki/file-type


fourdim (commented):
It still depends on buffer: https://github.com/sindresorhus/file-type/issues/578.
The only solution is to reimplement file-type.

JimmFly:
I also tried upgrading file-type package but got this error in AFFiNE.
```
Module build failed: UnhandledSchemeError: Reading from "node:buffer" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
```

packages/blocks/src/image-block/utils.ts
```diff
@@ -67,30 +67,11 @@ async function getImageBlob(model: ImageBlockModel) {
...
   }
-
   const doc = model.doc;
   const blob = await doc.blob.get(sourceId);
-
   if (!blob) {
     return null;
   }
-
-  if (!blob.type) {
```
EYHN, line 78: I suggest that we not delete file-type, because affine's sync backend does not store `blob.type`. You can expect all `blob.types` to be empty `""`. We can only calculate them through `file-type` when type is needed.

As for the buffer issue of file-type v19 in the browser, affine can workaround by adding `node:buffer` polyfill.

fourdim (approved):
`@sgtpooki/file-type` and it's direct dependencies are audited. No malicious codes found.

EYHN (approved):

PR #6810, opened 2024-04-18
fix(edgeless): optimize ai panel user experience

fundon:
Closes: AFF-902,AFF-901,AFF-908

### DIFF
--- a/packages/blocks/src/root-block/edgeless/components/toolbar/edgeless-toolbar.ts
+++ b/packages/blocks/src/root-block/edgeless/components/toolbar/edgeless-toolbar.ts
@@ -39,7 +39,10 @@ import {
 import type { EdgelessTool } from '../../../../_common/types.js';
 import { stopPropagation } from '../../../../_common/utils/event.js';
 import { getImageFilesFromLocal } from '../../../../_common/utils/filesys.js';
-import type { FrameBlockModel } from '../../../../index.js';
+import {
+  CopilotSelectionController,
+  type FrameBlockModel,
+} from '../../../../index.js';
 import { Bound, clamp } from '../../../../surface-block/index.js';
 import type { EdgelessRootBlockComponent } from '../../edgeless-root-block.js';
 import { isFrameBlock } from '../../utils/query.js';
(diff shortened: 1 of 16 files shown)


fundon:
* **#6810** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/6810?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/blocks/src/root-block/edgeless/components/toolbar/edgeless-toolbar.ts
```diff
@@ -348,6 +349,16 @@ export class EdgelessToolbar extends WithDisposable(LitElement) {
     }
   }
 
+  // The toolbar should be disabled while edgeless AI is in progress.
```
doouding, line 352: I think this part should be moved to the edgeless copilot widget.  Get the reference through `host.view.getWidget` and change its style or something.

doouding (approved):

packages/blocks/src/root-block/edgeless/components/toolbar/shape/shape-tool-button.ts
```diff
@@ -75,10 +75,7 @@ export class EdgelessShapeToolButton extends EdgelessToolButton<
     }
 
     .container-clip {
-      padding: 10000px 10000px 0;
```
CatsJuice, line 78: Leds to some unexpected effects #7139

fundon: A very large layer will be created here.

CatsJuice: Yes, that's expected, to achieve the effect of `overflow: clip` only on the bottom edge. Is there any performance issue or other problem with this approach?

fundon: The layer is always there. 

You can consider trying canvas overlayer, also create a separate element outside(fixed/absolute).

fundon: Screenshot 2024-05-26 at 12 45 37

fundon: > Yes, that's expected, to achieve the effect of `overflow: clip` only on the bottom edge

clip does not work on padding

PR #6818, opened 2024-04-19
fix(std): prevent unexpected touch behavior

EYHN:
Currently, unexpected drags will be triggered on touch screens.

I recommended to ignore touch events until we implement better touch logic.


EYHN:
* **#6818** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/6818?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

 Join @EYHN and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

Saul-Mirone:
My suggestion is that we should not introduce changes before adding support for touch screens.

EYHN:
I think we will eventually need to rewrite a completely different set of operation logic for touch, and before that it is appropriate to completely separate pointer events from touch events.

The result of this change is acceptable to me (can do simple text editing), but it may make some operations that were previously possible with the touch screen unavailable (for example, the toolbar cannot be clicked).

PR #6855, opened 2024-04-23
feat: create zod union

F4lkr4m:
## **Add util function for creating zod union from arrays**

example: _packages/blocks/src/\_common/consts_:
image

The example shows that when creating a zod scheme, there are repetitions of values from the array. Such use leads to an unguaranteed schema change when changing the values of this array.
This can be avoided using createZodUnion utility

+simplification of the code
+ensuring an up-to-date scheme

image

### DIFF
--- a/.github/CLA.md
+++ b/.github/CLA.md
@@ -124,3 +124,4 @@ Example:
 - RhonJoe, @JunIce, 2024/03/04
 - Lebedev Konstantin, @RubaXa, 2024/04/04
 - GambleMeow, @GambleMeow, 2024/04/10
+- Shcherbakov Sergey, @F4lkr4m, 2024/04/24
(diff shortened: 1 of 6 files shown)


packages/framework/global/src/utils/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from './assert.js';
+export * from './createZodUnion.js';
```
doodlewind, line 2: Thanks for the enhancement! Could you move this file to `packages/blocks/_common/utils/zod.ts`? There seems to be a better fit than the `global` package.

F4lkr4m: done!

doodlewind (approved):
Thanks for the feat!

Issue #6631, opened 2024-03-28
Unable to use note block after inserting a table in Edgeless editor.


RajaTheKing826:
I have created a table in a note block, and i tried to insert a new line at the end of the table but i didn't get slash popover to enter element. 

Steps to reproduce.
1. Open edgeless editor
2. Insert a note block by selecting note icon
3. insert a table view
4. Click bottom of the table view to enter new content you cant use it

Please check this.

RajaTheKing826:
@Flrande  can you please check this

Flrande:
We should add a new line when we click the spacing area of the blocks.

RajaTheKing826:
@golok727  Ive missed to add a data point that this issue is coming for youtube, figma, code block, image these kind of blocks

golok727:
> @golok727 Ive missed to add a data point that this issue is coming for youtube, figma, code block, image these kind of blocks

Will close all those in the PR

PR #6870, opened 2024-04-24
fix(edgeless): add a paragraph block when clicking on empty space on the note block

golok727:
close https://github.com/toeverything/blocksuite/issues/6631

### DIFF
--- a/packages/blocks/src/root-block/edgeless/components/block-portal/note/edgeless-note.ts
+++ b/packages/blocks/src/root-block/edgeless/components/block-portal/note/edgeless-note.ts
@@ -1,6 +1,8 @@
 import '../../note-slicer/index.js';
 
+import type { BlockElement } from '@blocksuite/block-std';
 import { ShadowlessElement, WithDisposable } from '@blocksuite/block-std';
+import type { BlockModel } from '@blocksuite/store';
 import { css, html, nothing } from 'lit';
 import { customElement, property, query, state } from 'lit/decorators.js';
 import { classMap } from 'lit/directives/class-map.js';
@@ -12,6 +14,8 @@ import { MoreIndicatorIcon } from '../../../../../_common/icons/edgeless.js';
 import { NoteDisplayMode } from '../../../../../_common/types.js';
 import { almostEqual, clamp } from '../../../../../_common/utils/math.js';
 import { matchFlavours } from '../../../../../_common/utils/model.js';
+import { getClosestBlockElementByPoint } from '../../../../../_common/utils/query.js';
+import { Point } from '../../../../../_common/utils/rect.js';
 import { handleNativeRangeAtPoint } from '../../../../../_common/utils/selection.js';
 import { type NoteBlockModel } from '../../../../../note-block/note-model.js';
 import { Bound, StrokeStyle } from '../../../../../surface-block/index.js';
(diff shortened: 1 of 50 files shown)


packages/blocks/src/root-block/edgeless/components/block-portal/note/edgeless-note.ts
```diff
@@ -188,6 +189,49 @@ export class EdgelessBlockPortalNote extends EdgelessPortalBase<NoteBlockModel>
...
     handleNativeRangeAtPoint(x, y);
+
+    // add an empty paragraph
+    const readonly = this.surface.doc.readonly;
+    const last = this.model.children.at(-1);
+    if (readonly) return;
+    if (
+      !last ||
+      !last.text ||
+      matchFlavours(last, [
```
donteatfriedrice, line 200: Can be changed to check whether it is a paragraph or a list

```
!matchFlavours(parent, ['affine:paragraph', 'affine:list'])
```

packages/blocks/src/root-block/edgeless/components/block-portal/note/edgeless-note.ts
```diff
@@ -188,6 +189,49 @@ export class EdgelessBlockPortalNote extends EdgelessPortalBase<NoteBlockModel>
     const x = clamp(e.x, rect.left + offsetX, rect.right - offsetX);
     const y = clamp(e.y, rect.top + offsetY, rect.bottom - offsetY);
     handleNativeRangeAtPoint(x, y);
+
+    // add an empty paragraph
```
donteatfriedrice, line 193: In this case, when the last block is a non-text block, a paragraph block will always be added at the bottom. The logic here should be to focus on the nearest text block first when clicking. If the nearest block is not text block, and the last block is not a text block, then add an empty block

donteatfriedrice (changes requested):
In this case, when the last block is a non-text block, a paragraph block will always be added at the bottom. The logic here should be to focus on the nearest text block first when clicking. If the nearest block is not text block, and the last block is not a text block, then add an empty block

golok727:
>when the last block is a non-text block, a paragraph block will always be added at the bottom. 

When I tested this, I found that a paragraph block wasn't added at the bottom as described when adding a non-text. We need to click on the empty area in the doc to add a new text block. 

https://github.com/toeverything/blocksuite/assets/123532141/747f81bd-e7b9-4ad9-a709-cc9d9aebae2d

donteatfriedrice:
> >when the last block is a non-text block, a paragraph block will always be added at the bottom. 
> 
> 
> 
> When I tested this, I found that a paragraph block wasn't added at the bottom as described when adding a non-text. We need to click on the empty area in the doc to add a new text block. 
> 
> 
> 
> https://github.com/toeverything/blocksuite/assets/123532141/747f81bd-e7b9-4ad9-a709-cc9d9aebae2d
> 
> 
> 
> 

You can add some paragraph blocks before the table, and click at the top or left or right near the paragraph blocks, in this case, should focus on the paragraph first like in doc mode. In the video you uploaded, there are no text blocks before table,

donteatfriedrice:
> @donteatfriedrice How is this behavior? I think we should consider adding a + button like in notion beside the drag handle
> 
>  Thing2.mp4

Looks great! Thanks! Could you add tests for it ?

donteatfriedrice (approved):

golok727:
> > @donteatfriedrice How is this behavior? I think we should consider adding a + button like in notion beside the drag handle
> > Thing2.mp4
> 
> Looks great! Thanks! Could you add tests for it ?

Sure!

Issue #6832, opened 2024-04-20
​Some bugs about tables


EnterMan123:
### What happened?

**1. The CTRL+V paste function cannot be used in the table expansion box**
![1](https://github.com/toeverything/AFFiNE/assets/107846196/5018c4a7-c288-4b95-9185-733384ac8add)

**2. Unable to use CTRL+A to select all text in the table box**
![2](https://github.com/toeverything/AFFiNE/assets/107846196/9c4f2fb7-9822-4612-bc20-8ef45cf13b4b)

**3. When there are too many column types that exceed the width of the table, there is no horizontal scroll bar below**
![5](https://github.com/toeverything/AFFiNE/assets/107846196/9a07162f-98a3-47c5-9b00-b7eedb2838e4)

**4. When there are too many column types, the expansion box overflows**
![4](https://github.com/toeverything/AFFiNE/assets/107846196/094eece3-6405-4639-b30c-6313eb09ec27)

**5. If the column type is text in the expansion box, any content entered will not be reflected in the table**
![6](https://github.com/toeverything/AFFiNE/assets/107846196/0f610918-eb26-49e2-a1e7-4f4d04277445)

**6. The table expansion box looks very crude and does not match the design style of the main program**

### Distribution version

Windows x64

### What browsers are you seeing the problem on if you're using web version?

Chrome

### Are you self-hosting?

- [ ] Yes

golok727:
(2) is closed in https://github.com/toeverything/blocksuite/pull/6570

(3) Is fixed cant find the PR

https://github.com/toeverything/blocksuite/assets/123532141/708aea20-f4fa-42d5-a0d2-c420c86298b4

golok727:
(4) https://github.com/toeverything/blocksuite/pull/6851

EnterMan123:
@golok727 If the column type is web link or text, the relevant text will not be searched. For example, I have a web link called abcde.com and a text called abcde. I cannot search them. If abcde exists in the title, I can search it.

![1](https://github.com/toeverything/blocksuite/assets/107846196/9c42c318-1f18-4549-92e6-967eaff992d7)
![2](https://github.com/toeverything/blocksuite/assets/107846196/1e872143-3060-4b2d-8bda-ce63fb251546)
![3](https://github.com/toeverything/blocksuite/assets/107846196/78938052-5f46-4765-b979-4b00ae5777f3)
[...]

PR #6858, opened 2024-04-23
refactor(database): side detail title

golok727:
image
image

PR #6875, opened 2024-04-25
fix(database): can't edit database side layout title

golok727:
Related to https://github.com/toeverything/blocksuite/pull/6858
Closes https://github.com/toeverything/blocksuite/issues/6832 (1)


golok727:
@zzj3720 if this is done then we can easily add the placeholder

```ts
 //TODO(@Flrande) wrap placeholder in `rich-text` or inline-editor to make it more developer-friendly
  private _updatePlaceholder = () => {
    if (
      !this._placeholderContainer ||
      !this._richTextElement ||
      !this.inlineEditor
    )
      return;

    if (
      this.doc.readonly ||
      this.inlineEditor.yTextLength > 0 ||
      this.inlineEditor.isComposing ||
      !this.selected ||
      this._isInDatabase()
    ) {
      this._placeholderContainer.classList.remove('visible');
    } else {
      this._placeholderContainer.classList.add('visible');
    }
  };

```

golok727:
This has bugs in edgeless. As it is attached to the root element it scrolls with the viewport in edgeless.

zzj3720:
The root cause of this bug is using `renderModel` to render single block. According to the current technical plan, it should be replaced with `renderSpecPortal`. However, I'm not sure how to use `renderSpecPortal` to only render a block that nested inside a database block. I need to discuss this with @Saul-Mirone.

golok727:
@zzj3720 should I close this for now ?

PR #6937, opened 2024-04-30
feat(sync): add blob engine

### LINKED ISSUE #6294: Document image upload to remote server
I'm currently test functionality of saving and then uploading snapshot of the page and got into the problem of uploading images. When uploading image and then trying to export it, it usually downloads a folder, with images in there. Is there an option to have images uploaded to S3 for example and then using URL from there?

### LINKED ISSUE #6752: blob resource collaboration issues
How to handle the collaboration of blob resources, and when will resource custom URLs be supported?

Flrande:
The blob engine completes a part that has been missing in the sync package, from now on all data synchronization of the blocksuite editor (including yjs data and blob data) can be achieved through the sync package.

### Breaking changes

Removed the [old blob addon](https://github.com/toeverything/blocksuite/pull/6937/files#diff-216478eb70e7b976470182010193d0a29cc3dbbdc01ec5c234976420b176ce63) (other addons will also be gradually removed and refactored in the future)

### Usage of new API

Similar to other engines in the sync package (doc engine and awareness engine), blob engine requires implementing a [blob source interface](https://github.com/toeverything/blocksuite/pull/6937/files#diff-e30ce6f58fc15bf3b61b8b03feaabf73a8a2cf350b423b07c663752915542707) to control the source of blob data. There are two types of data sources: main and shadow, usually one main and multiple shadows, where main is mandatory and shadow is optional. The blob engine will automatically synchronize multiple blob sources, prioritizing data retrieval from the main source.

For example, we implement a blob source based on IndexedDB and use it when initializing DocCollection.

```ts
import { createStore, del, get, keys, set } from 'idb-keyval';

class IndexedDBBlobSource implements BlobSource {
  readonly = false;

  readonly store = createStore(`${this.name}_blob`, 'blob');
  readonly mimeTypeStore = createStore(`${this.name}_blob_mime`, 'blob_mime');

  constructor(readonly name: string) {}

  async get(key: string) {
    const res = await get<ArrayBuffer>(key, this.store);
    if (res) {
      return new Blob([res], {
        type: await get(key, this.mimeTypeStore),
      });
    }
    return null;
  }

  async set(key: string, value: Blob) {
    await set(key, value.arrayBuffer(), this.store);
    await set(key, value.type, this.mimeTypeStore);
    return key;
  }

  async delete(key: string) {
    await del(key, this.store);
    await del(key, this.mimeTypeStore);
  }

  async list() {
    const list = await keys<string>(this.store);
    return list;
  }
}

const options: DocCollectionOptions = {
    // ...
    blobSources: {
      main: new IndexedDBBlobSource('quickEdgeless'),
    },
    // ...
  };
const collection = new DocCollection(options);
```

Another common use case is storing editor's blob data on a remote server. For this purpose, we only need to implement a corresponding blob source. For instance, for testing purposes, we have implemented a [mock server blob source](https://github.com/toeverything/blocksuite/pull/6937/files#diff-df896d35b8905f56255a1af5bbf7a3335843223d86d62dc4f72dd8ffb48d2f83R11).

The sync package has pre-implemented two commonly used blob sources ([IndexedDBBlobSource](https://github.com/toeverything/blocksuite/pull/6937/files#diff-19ad5f5d8dc56ab2e53bf90bd874dfa1cbcec4945797c37901d2ac654ab3606bR5) and [MemoryBlobSource](https://github.com/toeverything/blocksuite/pull/6937/files#diff-1b8dd5308f397ecafe774d4b62fb648171ea7414e71f5f879352919b7c3a38c3R3)), which can be used directly.

Close #6294
Close #6752
Close BS-283

### Migration Guide

```ts
// old
function createBlobStorage(): BlobStorage {
  return {
    crud: {
      set: async (key: string, value: Blob): Promise<string> => {
        await client.insertBlob(key, value);
        return key;
      },
      get: async (key: string): Promise<Blob | null> => {
        return client.getBlob(key);
      },
      delete: async (key: string): Promise<void> => {
        await client.deleteBlob(key);
      },
      list: async (): Promise<string[]> => {
        return client.getAllBlobIds();
      },
    },
  };
}

// new
class ClientBlobSource implements BlobSource {
  readonly = false;
  name = 'client';

  async get(key: string) {
    return client.getBlob(key);
  }

  async set(key: string, value: Blob) {
    return client.insertBlob(key, value);
  }

  async delete(key: string) {
    return client.deleteBlob(key);
  }

  async list() {
    return client.getAllBlobIds();
  }
}
```

### Delete Method

The blob engine does not yet support [the deletion of specific blobs](https://github.com/toeverything/blocksuite/pull/6937/files#diff-65e362c834ceec82ad613ff3adf72ac8b87cca840ad6e197a7e5492fd82cb191R172), as we currently cannot determine which blobs can be safely deleted, because we cannot confirm whether a blob is used by a doc. For this, we need to wait for the implementation of the indexer.


doodlewind (commented):
How does this close #6294? I haven't found implementation for uploading image blobs to remote server endpoints yet.

Will also better to add the migration guide. For example, as downstream user, when syncing blobs to SQLite with legacy `BlobStorage` API (https://github.com/toeverything/blocksuite/blob/master/examples/react-sqlite/src/editor/provider/provider.ts#L17) , the migrated implementation should be hinted.

Flrande:
> How does this close #6294? I haven't found implementation for uploading image blobs to remote server endpoints yet.

This depends on the storage service used; perhaps we can write an example based on R2 or S3.

> Would also be better to add migration guide, aka the comparison between the new API surface and the implementation based on the legacy `BlobStorage`. This is needed for bumping downstream (e.g. https://github.com/toeverything/blocksuite/blob/master/examples/react-sqlite/src/editor/provider/provider.ts#L17)

I will update the migration document, and after merging, I will migrate the old implementation in the example.

packages/framework/sync/src/blob/engine.ts
```diff
@@ -0,0 +1,189 @@
...
+          this.logger.error(`error when sync`, err);
+          continue;
+        }
+
+        const needUpload = mainList.filter(key => !shadowList.includes(key));
+        for (const key of needUpload) {
+          try {
+            const data = await this.main.get(key);
+            if (data) {
+              await shadow.set(key, data);
```
fourdim, line 82: Consider log the error if data does not exist.

packages/framework/sync/src/blob/engine.ts
```diff
@@ -0,0 +1,189 @@
...
+        }
+      }
+
+      const needDownload = shadowList.filter(key => !mainList.includes(key));
+
+      for (const key of needDownload) {
+        try {
+          const data = await shadow.get(key);
+          if (data) {
+            await this.main.set(key, data);
```
fourdim, line 99: Same here.

fourdim (approved):
LGTM, but please wait for other reviewers.

Flrande:
Considering the migration cost, I will merge this PR when AFFiNE releases 0.14. cc @pengx17 @EYHN

EYHN (approved):

packages/framework/sync/src/blob/engine.ts
```diff
@@ -0,0 +1,196 @@
...
+}
+
+/**
+ * # BlobEngine
+ *
+ * sync blobs between storages in background.
+ *
+ * all operations priority use main, then use shadows.
+ */
+export class BlobEngine {
```
Saul-Mirone, line 16: Need unit test for this feature.

Saul-Mirone (changes requested):

Saul-Mirone (approved):

PR #6968, opened 2024-05-06
feat(blocks): add esc tip when iframe is fullscreen

fundon:
Closes: [TOV-843](https://linear.app/affine-design/issue/TOV-843/添加-esc-退出提醒，在-iframe-全屏时)

https://github.com/toeverything/blocksuite/assets/27926/e2a72011-e3b6-4a45-977f-7f0adb2ef227

### DIFF
--- a/packages/blocks/src/embed-html-block/embed-html-block.ts
+++ b/packages/blocks/src/embed-html-block/embed-html-block.ts
@@ -108,6 +108,75 @@ export class EmbedHtmlBlockComponent extends EmbedBlockElement<
     }
   }
 
+  private _fullscreenchange() {
+    const iframe = this._iframe;
+    if (!iframe) return;
+
+    const show =
+      document.fullscreenEnabled && document.fullscreenElement === iframe;
+
+    iframe.contentWindow?.postMessage({ show }, '*');
+  }
+
+  private _tipHtml() {
+    const computed = getComputedStyle(this);
+    const fontFamily = computed.getPropertyValue('--affine-font-sans-family');
+    const fontSize = computed.getPropertyValue('--affine-font-sm');
+    const borderColor = computed.getPropertyValue('--affine-black-10');
+    const backgroundColor = computed.getPropertyValue('--affine-primary-color');
+    const tipColor = computed.getPropertyValue('--affine-white');
+    const color = computed.getPropertyValue('--affine-text-secondary-color');
+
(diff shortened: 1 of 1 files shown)


fundon:
* **#6968** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/6968?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/blocks/src/embed-html-block/embed-html-block.ts
```diff
@@ -108,6 +108,73 @@ export class EmbedHtmlBlockComponent extends EmbedBlockElement<
...
+        border: 1px solid ${borderColor};
+        background: ${backgroundColor};
+      }
+
+      #affine-tip[data-show] {
+        display: flex;
+      }
+      </style>
+      <div id="affine-tip">
+        Press&nbsp;<span class="key">ESC</span>&nbsp;to&nbsp;close
```
pengx17, line 168: The ESC prompt should also be a button that exits fullscreen on click

packages/blocks/src/embed-html-block/embed-html-block.ts
```diff
@@ -108,6 +108,73 @@ export class EmbedHtmlBlockComponent extends EmbedBlockElement<
...
+      </div>
+      <script>
+      window.onmessage = function(event){
+        document.getElementById('affine-tip').toggleAttribute('data-show', event.data && event.data.show);
+      };
```
pengx17, line 173: Can we listen to `fullscreenchange` event inside of the iframe so that we do not need to rely on message passing? Message maybe be polluted by browser extensions.

fundon: it does not work.

packages/blocks/src/embed-html-block/embed-html-block.ts
fundon: Only if `allow-same-origin` is enabled in the sandbox.

pengx17: 1. may use message to tell the top frame to exit fullscreen
2. or perhaps draw the tip overlay outside of the iframe so that no message passing is needed

pengx17 (approved):

PR #6971, opened 2024-05-07
fix(presets): limit ai text rendering

donteatfriedrice:
https://github.com/toeverything/blocksuite/assets/99816898/508752d6-68a1-408f-806c-bc50201827ef


Flrande (approved):

doodlewind (approved):

packages/presets/src/ai/messages/text.ts
```diff
@@ -56,51 +52,58 @@ export class AIAnswerTextPreview extends WithDisposable(LitElement) {
   @property({ attribute: false })
   answer!: string;
 
-  private _previewDoc: Doc | null = null;
+  private _doc!: Doc;
+  private _answers: string[] = [];
```
EYHN, line 56: It looks like you don't need an array here since you only care about the last item

PR #6982, opened 2024-05-07
fix: chat scroll down

regischen:


doodlewind (changes requested):
Should fire event on re-render, instead of using hard-coded timeout to ensure timing sequence.

PR #6990, opened 2024-05-08
fix: should not inherit last text element hasMaxWidth prop

donteatfriedrice:


Flrande (approved):

Flrande:
It seems more fundamentally to avoid this field being cached directly to solve the problem

https://github.com/toeverything/blocksuite/blob/0b499bcf02f89b05f7a8d6c20fbe46327ae4f645/packages/blocks/src/surface-block/managers/edit-session.ts#L101

PR #6992, opened 2024-05-08
perf(edgeless): slice first block rendering to multiple parts

doouding:
### Change
- split first blocks rendering into multiple parts
- move edgeless label update logic into `edgeless-index-label` component
- remove unecessary code from portal
- simplify layer `zIndex` property

### DIFF
--- a/packages/blocks/src/_common/utils/event.ts
+++ b/packages/blocks/src/_common/utils/event.ts
@@ -168,20 +168,20 @@ export function requestConnectedFrame(
 /**
  * A wrapper around `requestConnectedFrame` that only calls at most once in one frame
  */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function batchToAnimationFrame<T extends (...args: any[]) => any>(
-  func: T,
-  element?: HTMLElement
-): T {
+export function requestThrottledConnectFrame<
+  T extends (...args: unknown[]) => void,
+>(func: T, element?: HTMLElement): T {
   let raqId: number | undefined = undefined;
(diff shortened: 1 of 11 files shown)


packages/blocks/src/root-block/edgeless/components/block-portal/edgeless-block-portal.ts
```diff
@@ -56,6 +52,7 @@ const portalMap = new Map<EdgelessBlockType | RegExp, string>([
 export class EdgelessBlockPortalContainer extends WithDisposable(
   ShadowlessElement
 ) {
+  static RENDER_STEP = 5;
```
doodlewind, line 55: Comment needed to elaborate the step-based rendering strategy

packages/blocks/src/_common/utils/event.ts
```diff
@@ -168,20 +168,21 @@ export function requestConnectedFrame(
 /**
  * A wrapper around `requestConnectedFrame` that only calls at most once in one frame
  */
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function batchToAnimationFrame<T extends (...args: any[]) => any>(
-  func: T,
-  element?: HTMLElement
-): T {
+export function requestThrottledConnectFrame<
+  // eslint-disable-next-line
```
Saul-Mirone, line 172: Instead disable eslint, you should disable a specific rule

doouding:
fix: use unknown

Saul-Mirone (approved):

PR #6998, opened 2024-05-09
fix(blocks): support overlapping of inline styles

### LINKED ISSUE #6986: code markup does not take care of link
- if a text is marked as code, than link, it will lose the code format
- if a text is marked as link, than code, it will lose the blue color to suggest it is clickable, which makes some inconsistent; also the font size would change

Flrande:
Close #6986

https://github.com/toeverything/blocksuite/assets/50035259/a425bc24-b87e-45c3-90e9-cf02d339c61f

### DIFF
--- a/packages/blocks/src/_common/inline/presets/nodes/link-node/affine-link.ts
+++ b/packages/blocks/src/_common/inline/presets/nodes/link-node/affine-link.ts
@@ -62,21 +62,8 @@ export class AffineLink extends ShadowlessElement {
   }
 
   static override styles = css`
-    affine-link > a {
-      white-space: nowrap;
-      word-break: break-word;
-      color: var(--affine-link-color);
-      fill: var(--affine-link-color);
-      text-decoration: none;
-      cursor: pointer;
-    }
-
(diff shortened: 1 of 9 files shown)


packages/blocks/src/_common/inline/presets/nodes/link-node/affine-link.ts
```diff
@@ -63,20 +63,20 @@ export class AffineLink extends ShadowlessElement {
 
   static override styles = css`
     affine-link > a {
-      white-space: nowrap;
-      word-break: break-word;
-      color: var(--affine-link-color);
-      fill: var(--affine-link-color);
-      text-decoration: none;
-      cursor: pointer;
+      white-space: nowrap !important;
```
doodlewind, line 66: Is it possible avoiding using `!important`?

fourdim: https://developer.mozilla.org/en-US/docs/Web/CSS/@layer
I don't know whether this can be implemented. Just for reference.

Saul-Mirone (approved):

PR #7002, opened 2024-05-09
fix: avoid selection change when ai-panel display

donteatfriedrice:
To fixed [TOV-834](https://linear.app/affine-design/issue/TOV-834/激活-ai-panel-后需要将-focus-保持在-ai-panel)

- [x] Add an affine-viewport-overlay-widget to avoid selection change when some panel/dialog/modal opened. By preventing the dragStart, pointerDown, and click event to dispatch when viewport widget is locked.

Before, when ai-panel opened and has generated some content, if user click or drag outside of the ai panel, the selection will be changed, then the response actions will be wired, for example, insert below or replace the wrong blocks.

https://github.com/toeverything/blocksuite/assets/99816898/a8c1d89b-0058-4bd8-85ca-75cd624dcca4

Now can lock the viewport when ai-panel opened by calling affineViewportOverlayWidget.lock().

https://github.com/toeverything/blocksuite/assets/99816898/958ab7b7-07a9-440c-8129-3fcf192881a3

### DIFF
--- a/packages/blocks/src/root-block/widgets/ai-panel/ai-panel.ts
+++ b/packages/blocks/src/root-block/widgets/ai-panel/ai-panel.ts
@@ -14,6 +14,7 @@ import { choose } from 'lit/directives/choose.js';
 
 import type { AIError } from '../../../_common/components/index.js';
 import { stopPropagation } from '../../../_common/utils/event.js';
+import type { AffineViewportOverlayWidget } from '../viewport-overlay/viewport-overlay.js';
 import type { AIPanelDiscardModal } from './components/discard-modal.js';
 import { toggleDiscardModal } from './components/discard-modal.js';
 import type { AffineAIPanelState, AffineAIPanelWidgetConfig } from './type.js';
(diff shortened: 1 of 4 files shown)


packages/blocks/src/root-block/widgets/ai-panel/ai-panel.ts
```diff
@@ -77,6 +78,16 @@ export class AffineAIPanelWidget extends WidgetElement {
   @query('.mock-selection-container')
   mockSelectionContainer!: HTMLDivElement;
 
+  get viewportOverlayWidget() {
+    const rootId = this.host.doc.root?.id;
+    return rootId
+      ? (this.host.view.getWidget(
+          'affine-viewport-overlay-widget',
```
Flrande, line 85: You can use `AFFINE_VIEWPORT_OVERLAY_WIDGET` here

Flrande (approved):

packages/blocks/src/root-block/widgets/ai-panel/ai-panel.ts
```diff
@@ -325,19 +338,28 @@ export class AffineAIPanelWidget extends WidgetElement {
...
         : null;
 
       if (formatBar) {
         formatBar.requestUpdate();
       }
     }
+
+    if (this.state !== 'hidden') {
+      this.viewportOverlayWidget?.lock();
+    } else {
```
fundon, line 360: The order can be swapped here. 
`!==` no need.

fundon (approved):

PR #7039, opened 2024-05-14
refactor: cleanup block specs

Flrande:


doodlewind (commented):
I'm not sure why it's needed to split `paragraph-block/paragraph-spec.ts` and `paragraph-block/index.ts`. Why not simply compose the spec in `index.ts`?

Flrande:
> I'm not sure why it's needed to split `paragraph-block/paragraph-spec.ts` and `paragraph-block/index.ts`. Why not simply compose the spec in `index.ts`?

I have no particular preference for this, but some blocks may have several different specs. For example, the specs for surface in edgeless mode and page mode are different.

Saul-Mirone (approved):

PR #7043, opened 2024-05-14
fix: browser window resize causes image deformation

akumatus:
### TL;DR

This pull request removes the height property from the image css style. No matter how the browser window size changes, images can keep in shape with their original aspect ratio.

Close affine-design/issue/BS-304.

| Before      | After |
| ----------- | ----------- |
| 截屏2024-05-14 17 10 31| 截屏2024-05-14 17 10 43|

### What changed?

The height property was removed from two different places in the code base:
1. In the `ImageBlockPageComponent` class, it was removed from the `style` of  resizable image container div.
2. In the `ImageResizeManager` class, it was removed from the `style` of dragging image container.

### How to test?

Incorporate these changes into your local codebase, and try to resize the browser window. You should now see that the height is automatically adjusted according to the width.

### Why make this change?

This change helps the images maintain their aspect ratio when being resized. By removing the explicit height, the browser's default behaviour of maintaining the aspect ratio kicks in and thus helps prevent any distortions during resizing operation.

---

### DIFF
--- a/packages/blocks/src/image-block/components/page-image-block.ts
+++ b/packages/blocks/src/image-block/components/page-image-block.ts
@@ -251,7 +251,6 @@ export class ImageBlockPageComponent extends WithDisposable(ShadowlessElement) {
     if (this._isDragging && this.resizeImg) {
       return {
         width: this.resizeImg.style.width,
-        height: this.resizeImg.style.height,
       };
     }
 
@@ -265,7 +264,6 @@ export class ImageBlockPageComponent extends WithDisposable(ShadowlessElement) {
 
     return {
       width: `${width}px`,
-      height: `${height}px`,
     };
   }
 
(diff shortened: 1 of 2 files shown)


akumatus:
* **#7043** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7043?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

doodlewind (approved):
Good catch, thanks!

doodlewind:
Would be better to add the `close AFF-xxx` desc, so that we can link and close corresponding linear issue!

akumatus:
> Would be better to add the `close AFF-xxx` desc, so that we can link and close corresponding linear issue!

Thanks for your advice! I have updated this PR description and linked this PR to the linear issue.

PR #7044, opened 2024-05-14
perf: avoid recalculate style

doouding:
### Change
- Move `<style>` to static property to avoid style recalculation on each instance.
- Set initial `contenteditable` to the inline editor container to avoid later recalculation

### DIFF
--- a/packages/blocks/src/list-block/list-block.ts
+++ b/packages/blocks/src/list-block/list-block.ts
@@ -25,6 +25,7 @@ export class ListBlockComponent extends BlockElement<
   ListBlockModel,
   ListBlockService
 > {
+  static override styles = listBlockStyles;
   get inlineManager() {
     const inlineManager = this.service?.inlineManager;
     assertExists(inlineManager);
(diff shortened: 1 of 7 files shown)


doouding:
* **#7044** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7044?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/blocks/src/list-block/list-block.ts
```diff
@@ -190,9 +191,6 @@ export class ListBlockComponent extends BlockElement<
 
     return html`
       <div class=${'affine-list-block-container'}>
-        <style>
```
Flrande, line 193: This may cause the loss of styles in the AI text renderer, and we may need to disable the shadow DOM involving the AI text renderer. cc @donteatfriedrice

Saul-Mirone (approved):

packages/blocks/src/list-block/list-block.ts
donteatfriedrice: I will check again.

PR #7061, opened 2024-05-15
fix(blocks): export block spec

Flrande:


L-Sun (approved):

packages/blocks/src/specs/index.ts
```diff
@@ -7,8 +31,40 @@ SpecProvider.getInstance().addSpec('page', PageEditorBlockSpecs);
 SpecProvider.getInstance().addSpec('edgeless', EdgelessEditorBlockSpecs);
 SpecProvider.getInstance().addSpec('preview', PreviewEditorBlockSpecs);
 
+// specs preset
 export * from './edgeless-specs.js';
 export * from './page-specs.js';
 export * from './preview-specs.js';
 export * from './utils/spec-builder.js';
 export * from './utils/spec-provider.js';
+
```
Saul-Mirone, line 40: You can use

```ts
const commonSpec = {};
const edgelessSpec = {};

export {
  ...commonSpec,
  ...edgelessSpec
}
```

instead of disable eslint rules.

PR #7062, opened 2024-05-15
fix: the code block wrap state shoud be persisted in database

akumatus:
### TL;DR

This Pull Request moves the `wrap` property from temporary state to persistent data model. When user refreshes the web browser, the code block warp style keeps.

Fix affine-design/issue/BS-297.

### What changed?

- In `code-block.ts`, the `_wrap` private state variable has been removed, and all references to it have instead been directed to `this.model.wrap`.
- A `setWrap` function has been added to handle updating the `wrap` property in the block model, which can be persisted in database.

### How to test?

Test this change by creating a code block in document and toggle the `wrap` icon. Check if this modifies the wrap state in the model and updates the UI accordingly. Refresh the browser, the code block warp styles keeps.

截屏2024-05-15 16 03 20

截屏2024-05-15 16 03 02

---

### DIFF
--- a/packages/blocks/src/code-block/code-block.ts
+++ b/packages/blocks/src/code-block/code-block.ts
@@ -48,9 +48,6 @@ import { getHighLighter } from './utils/high-lighter.js';
 export class CodeBlockComponent extends BlockElement<CodeBlockModel> {
   static override styles = codeBlockStyles;
 
-  @state()
-  private _wrap = false;
-
   @query('.lang-button')
   private _langButton!: HTMLButtonElement;
 
@@ -176,9 +173,9 @@ export class CodeBlockComponent extends BlockElement<CodeBlockModel> {
         CodeOptionTemplate({
           anchor: this,
           model: this.model,
-          wrap: this._wrap,
-          onClickWrap: () => {
-            this._wrap = !this._wrap;
+          wrap: this.model.wrap,
+          toggleWrap: () => {
+            this.setWrap(!this.model.wrap);
             updatePortal();
           },
           abortController,
@@ -429,6 +426,10 @@ export class CodeBlockComponent extends BlockElement<CodeBlockModel> {
     });
   }
 
+  setWrap(wrap: boolean) {
+    this.doc.updateBlock(this.model, { wrap });
+  }
+
   private _onClickLangBtn() {
     if (this.readonly) return;
     if (this._langListAbortController) return;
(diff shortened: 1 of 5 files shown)


akumatus:
* **#7062** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7062?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

Flrande:
Can you add a test for this case?

doodlewind (commented):
Need to test that for legacy data without the `wrap` field, whether the backward compatibility could be supported.

doodlewind (approved):

akumatus:
> Can you add a test for this case?

Thanks for your comments! I have added several tests in `tests/code.spec.ts`.

akumatus:
> Need to test that for legacy data without the `wrap` field, whether the backward compatibility could be supported.

https://github.com/toeverything/blocksuite/blob/0cbbba69fe3fabbd61d9c8f9774ae14e6502e89e/packages/framework/store/src/store/doc/block.ts#L216

I suppose that legacy data will filled with default `wrap:false` value, because of the above code. But I don't know how to write the test in that case, could you give me a hint?

Flrande (approved):

Flrande:
> > Need to test that for legacy data without the `wrap` field, whether the backward compatibility could be supported.
> 
> https://github.com/toeverything/blocksuite/blob/0cbbba69fe3fabbd61d9c8f9774ae14e6502e89e/packages/framework/store/src/store/doc/block.ts#L216
> 
> @doodlewind I suppose that legacy data will filled with default `wrap:false` value, because of the above code. But I don't know how to write the test in that case, could you give me a hint?

You can test by importing an old snapshot.

akumatus:
> Need to test that for legacy data without the `wrap` field, whether the backward compatibility could be supported.

I have tested both import old snapshot and paste old code block cases, and everything works fine!

donteatfriedrice:
If we move the wrap property to the model, we may need to disable the wrap code button in code toolbar as well when readonly mode, because it will not work in readonly mode.

doodlewind (approved):
LGTM, for the feedback in https://github.com/toeverything/blocksuite/pull/7062#issuecomment-2113080652, it could be implemented in a standalone PR or simply marked help wanted!

akumatus:
> If we move the wrap property to the model, we may need to disable the wrap code button in code toolbar when readonly mode.
> 
>  Screen.Recording.2024-05-16.at.01.29.05.mov

Got it! Thanks for your comment, it helps a lot.

Saul-Mirone (approved):

PR #7065, opened 2024-05-15
refactor(presets): export ai block spec instead of patch function

Flrande:


packages/presets/src/ai/ai-spec.ts
```diff
@@ -0,0 +1,126 @@
...
+import { buildAIPanelConfig } from './ai-panel.js';
+import {
+  setupEdgelessCopilot,
+  setupEdgelessElementToolbarEntry,
+} from './entries/edgeless/index.js';
+import { setupFormatBarEntry } from './entries/format-bar/setup-format-bar.js';
+import { setupSlashMenuEntry } from './entries/slash-menu/setup-slash-menu.js';
+import { setupSpaceEntry } from './entries/space/setup-space.js';
+
+export const AIPageRootBlockSpec: BlockSpec = {
```
pengx17, line 27: I still think it is better to use patch function instead. Imagine we have different types of provided specs, then there is no way to use them together in this way.

Flrande: Reasonable, but I will change the patch function to receive a specific single block spec instead of an spec array, because now we encourage AFFiNE to assemble the required block specs themselves.

pengx17: I'm not so sure what if certain feature like AI requires patching multiple specs. Individual patch function may not work at all on its own, unless different specs are patched all together.

pengx17 (approved):

PR #7089, opened 2024-05-17
fix: selecting multiple blocks and dragging will show multiple placeholders

akumatus:
### What changed?
- the placeholder will be visible, only if the selection is text selection and collapsed
- add e2e test

Fix affine-design/issue/BS-312.

|  Before  | After |
|  ----  | ----  |
|  | |

### How to test?

Test the paragraph block in various states like editing, selecting, etc. Ensure the placeholder visibility is as expected.

---

### DIFF
--- a/packages/blocks/src/paragraph-block/paragraph-block.ts
+++ b/packages/blocks/src/paragraph-block/paragraph-block.ts
@@ -103,11 +103,14 @@ export class ParagraphBlockComponent extends BlockElement<
     )
       return;
 
+    const selection = this.host.selection.find('text');
+    const isCollapsed = selection?.isCollapsed() ?? false;
     if (
       this.doc.readonly ||
       this.inlineEditor.yTextLength > 0 ||
       this.inlineEditor.isComposing ||
       !this.selected ||
+      !isCollapsed ||
       this._isInDatabase()
     ) {
       this._placeholderContainer.classList.remove('visible');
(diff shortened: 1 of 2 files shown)


akumatus:
* **#7089** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7089?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/blocks/src/paragraph-block/paragraph-block.ts
```diff
@@ -103,11 +107,15 @@ export class ParagraphBlockComponent extends BlockElement<
     )
       return;
 
+    const selection = this.host.selection.value[0];
```
Saul-Mirone, line 110: Better to use `selection.find('text')`

packages/blocks/src/paragraph-block/paragraph-block.ts
```diff
@@ -103,11 +107,15 @@ export class ParagraphBlockComponent extends BlockElement<
     )
       return;
 
+    const selection = this.host.selection.find('text');
+    const isCollapsed =
+      selection instanceof TextSelection ? selection.isCollapsed : false;
```
Saul-Mirone, line 112: If you use `find` method, you won't need to check the type of it, so it can be replaced by `selection?.isCollapsed ?? false`.

Saul-Mirone (approved):

packages/blocks/src/paragraph-block/paragraph-block.ts
akumatus: Got it! I will update this change in another PR, thanks for your advice.

PR #7113, opened 2024-05-22
refactor(blocks): move edgeless type and remove useless type

Flrande:
Move from surface block to edgeless root block.


doouding (approved):

doouding:
Adding some description would be better.

Flrande:
> Adding some description would be better.

done

PR #7130, opened 2024-05-23
fix: safari compatibility

fundon:


fundon:
* **#7130** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7130?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/framework/inline/src/services/range.ts
```diff
@@ -73,9 +74,10 @@ export class RangeService<TextAttributes extends BaseTextAttributes> {
       // range change may happen before the editor is prepared
       await this.editor.waitForUpdate();
       // improve performance
-      requestIdleCallback(() => {
-        this.editor.requestUpdate(false);
-      });
+      REQUEST_IDLE_CALLBACK_ENABLED &&
```
Saul-Mirone, line 77: If requestIdleCallback is not supported, maybe we should call this with setTimeout or Promise.resolve. @Flrande what do you think?

Flrande: `Promise.resolve` look good to me

fundon: Change to:

```ts
      if (REQUEST_IDLE_CALLBACK_ENABLED) {
        requestIdleCallback(() => {
          this.editor.requestUpdate(false);
        });
      } else {
        await Promise.resolve(this.editor.requestUpdate(false));
      }
```

PR #7144, opened 2024-05-27
feat(edgeless): add edgeless text

Flrande:
https://blocksuite-git-flrande-feat-edgeless-text-0527-toeverything.vercel.app/?enable_edgeless_text=true


packages/blocks/src/_common/utils/iterable.ts
```diff
@@ -13,11 +13,12 @@
  */
 export function countBy<T>(
   items: T[],
-  key: (item: T) => string | number
+  key: (item: T) => string | number | null
 ): Record<string, number> {
   const count: Record<string, number> = {};
   items.forEach(item => {
     const k = key(item);
```
Saul-Mirone, line 20: what if k is 0

Saul-Mirone:
https://github.com/toeverything/blocksuite/assets/10047788/afca18f2-b8db-42bb-807f-8e7d3ccd0bcc

Also the keyboard navigation with `Arrow-Up` and `Arrow-Down` is totally not working.

Saul-Mirone:
Keyboard navigation not working.

Saul-Mirone:
cannot select multiple lines.

Saul-Mirone:
https://github.com/toeverything/blocksuite/assets/10047788/c9be9dca-4d82-403a-a483-ec13bb3f88ce

Seems still not working.

Flrande:
I am fixing

packages/blocks/src/surface-block/element-model/text.ts
```diff
@@ -15,14 +21,15 @@ import { yfield } from './decorators.js';
...
-  fontFamily: FontFamily;
-  fontWeight?: FontWeight;
-  fontStyle?: FontStyle;
-  textAlign: TextAlign;
+  // color: string;
+  // fontSize: number;
+  // fontFamily: FontFamily;
+  // fontWeight?: FontWeight;
+  // fontStyle?: FontStyle;
+  // textAlign: TextAlign;
```
doouding, line 29: Should we delete thoese?

doouding (commented):
Well, I would suggest not to change the API or property on common used module if it's not necessary. 

Because this would make a PR super large, and it required lots of ongoing PRs to adapt those change.

Flrande:
> Well, I would suggest not to change the API or property on common used module if it's not necessary.
> 
> Because this would make a PR super large, and it required lots of ongoing PRs to adapt those change.

Historical burdens won't disappear, they will only accumulate more and more.

Saul-Mirone (approved):
There's a lot of bugs need to be fixed. Approve for not blocking.

PR #7166, opened 2024-05-29
feat: use view type in selector for block rendering

Saul-Mirone:


doodlewind (commented):
What is the block selector feature documented? If there's an API update, the docs should also be updated correspondingly.

Saul-Mirone:
> What is the block selector feature documented? If there's an API update, the docs should also be updated correspondingly.

The selector feature is still WIP and I think there's no documentation for it now...(which means we don't need to updated anything....)

PR #7178, opened 2024-05-30
refactor(blocks): extract common components from blocks

golok727:
Extracts common components
Extend from this class `block-selection` or `block-caption-editor` is needed.
```ts
export class BlockComponent<
  Model extends BlockModel = BlockModel,
  Service extends BlockService = BlockService,
  WidgetName extends string = string,
> extends BlockElement<Model, Service, WidgetName> {

  protected accessor useCaptionEditor = false;

  protected accessor blockContainerStyles: StyleInfo | undefined = undefined;
}

```

Closes: [BS-361](https://linear.app/affine-design/issue/BS-361/extract-common-logic-from-blocks)

### DIFF
--- a/packages/blocks/src/_common/components/block-component.ts
+++ b/packages/blocks/src/_common/components/block-component.ts
@@ -1,0 +1,48 @@
+import { BlockElement, type BlockService } from '@blocksuite/block-std';
+import { type BlockModel } from '@blocksuite/store';
+import { html, nothing } from 'lit';
+import { query } from 'lit/decorators.js';
+import { type StyleInfo, styleMap } from 'lit/directives/style-map.js';
+
+import type { BlockCaptionEditor } from './block-caption.js';
+
+export class BlockComponent<
+  Model extends BlockModel = BlockModel,
+  Service extends BlockService = BlockService,
+  WidgetName extends string = string,
+> extends BlockElement<Model, Service, WidgetName> {
+  protected accessor useCaptionEditor = false;
+
+  protected accessor blockContainerStyles: StyleInfo | undefined = undefined;
+
+  @query('.affine-block-component > block-caption-editor')
+  private accessor _captionEditor!: BlockCaptionEditor | null;
+
+  constructor() {
+    super();
+    this.addRenderer(this._renderWithWidget);
+  }
+
+  get captionEditor() {
+    if (!this.useCaptionEditor || !this._captionEditor)
+      throw new Error(
+        'Oops! Please enable useCaptionEditor before accessing captionEditor'
+      );
+    return this._captionEditor;
+  }
+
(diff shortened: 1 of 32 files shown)


golok727:
* **#7178** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7178?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

 Join @golok727 and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/blocks/src/_common/components/block-component.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  get captionEditor() {
+    if (!this.useCaptionEditor || !this._captionEditor)
+      throw new Error(
+        'Oops! Please enable useCaptionEditor before accessing captionEditor'
+      );
+    return this._captionEditor;
+  }
+
+  override render() {
```
Saul-Mirone, line 29: Instead of override `render`, you should override `renderBlock`

golok727: But the subclass will override the `renderBlock` method.

Saul-Mirone: The problem of this code is that, it will always try to render `affine-block-selection` even if the block is [in `Hide` or `Bypass` mode](https://github.com/toeverything/blocksuite/blob/master/packages/framework/block-std/src/view/element/block-element.ts#L301C5-L320C8). And even wroth, when the component is in error mode, thoses components will still be rendered.

golok727: > The problem of this code is that, it will always try to render `affine-block-selection` even if the block is [in `Hide` or `Bypass` mode](https://github.com/toeverything/blocksuite/blob/master/packages/framework/block-std/src/view/element/block-element.ts#L301C5-L320C8). And even wroth, when the component is in error mode, thoses components will still be rendered.

Yeah, I noticed it. PTAL of the new version.

Saul-Mirone: This will be hard to maintain if we add new view types. I'll add a new solution here in your branch.

fundon:
It might be better to add a state to the block caption to control the edit state.
Use `position-event` CSS property to control it.


https://github.com/toeverything/blocksuite/assets/27926/57d5a017-2c0f-473b-a75e-0abf7ad6b4ca

packages/blocks/src/_common/components/block-component.ts
Saul-Mirone: Done

fundon (approved):

PR #7219, opened 2024-06-05
fix(edgeless): missing rerender during stacked canvases updates

oliseviche:
When there are a mix of canvas and block elements on the board, the stacked canvases didn't rerender themselves after top object is being deleted. This leads to the state when all underlying layer elements visually dissapear.

https://github.com/toeverything/blocksuite/assets/8721360/55444532-3a63-41d9-a833-440f70be7ceb


doouding:
Good catch! Thanks for the fix.

doouding (approved):

PR #7243, opened 2024-06-07
refactor: add docsService for doc mode get, set, toogle and observe change

akumatus:
### What changed?
- add docsService for doc mode get, set, toogle and observe change
- set doc mode to edgeless after linked doc created

[Related PR in AFFiNE](https://github.com/toeverything/AFFiNE/pull/7174)

---

### DIFF
--- a/packages/blocks/src/root-block/edgeless/edgeless-root-service.ts
+++ b/packages/blocks/src/root-block/edgeless/edgeless-root-service.ts
@@ -82,8 +82,6 @@ export class EdgelessRootService extends RootService {
       blockId?: string;
     }>(),
     tagClicked: new Slot<{ tagId: string }>(),
-    editorModeSwitch: new Slot<'edgeless' | 'page'>(),
-
     toolbarLocked: new Slot<boolean>(),
   };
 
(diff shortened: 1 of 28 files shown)


akumatus:
* **#7243** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7243?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/blocks/src/root-block/edgeless/edgeless-root-service.ts
```diff
@@ -83,8 +84,8 @@ export class EdgelessRootService extends RootService {
       blockId?: string;
     }>(),
     tagClicked: new Slot<{ tagId: string }>(),
-    editorModeSwitch: new Slot<'edgeless' | 'page'>(),
-
+    editorModeSwitch: new Slot<DocMode>(),
+    setEditorMode: new Slot<{ id: string; mode: DocMode }>(),
```
pengx17, line 88: I think `slots.editorModeSwitch` in base `RootService` should also be amended.
BTW IMO we can reuse & extend `editorModeSwitch`'s event so that there is no need to introduce another event slot.

akumatus: I used to think it's a good idea to separate set doc mode into two parts, set current doc and set others, because there are lots of listeners observe current doc mode change.

akumatus: Let me check the releated logic.

PR #7245, opened 2024-06-07
fix(blocks): refine tooltip visual effect

L-Sun:
Close BS-327

- Refine tooltip visual effect

https://github.com/toeverything/blocksuite/assets/20479050/dc078ea2-475b-4560-9935-4dbec57f5b2f

https://github.com/toeverything/blocksuite/assets/20479050/6dd62885-b28b-4b64-95c3-a3e4dd301827

- Fix edgeless tool bar button size (40x43.33 -> 40x40)
image
image

PR #7246, opened 2024-06-07
refactor(blocks): hover controller

golok727:
related to https://github.com/toeverything/blocksuite/pull/7245

https://github.com/toeverything/blocksuite/assets/123532141/593769b5-8a9d-4003-bdae-e071ed728f33

### DIFF
--- a/packages/blocks/src/_common/components/hover/controller.ts
+++ b/packages/blocks/src/_common/components/hover/controller.ts
@@ -119,6 +119,7 @@ export class HoverController implements ReactiveController {
       // If some problems arise when aborting the previous hover,
       // consider fixing the transition related issues and return void here
       this._abortController?.abort();
+
       this._abortController = new AbortController();
       this._abortController.signal.addEventListener('abort', () => {
         this._abortController = undefined;
--- a/packages/blocks/src/_common/components/filterable-list/index.ts
+++ b/packages/blocks/src/_common/components/filterable-list/index.ts
@@ -99,7 +99,9 @@ export class FilterableListComponent<Props = unknown> extends WithDisposable(
             'filterable-item': true,
             focussed,
           })}
+          @mouseover=${() => (this._curFocusIndex = idx)}
           @click=${() => this._select(item)}
+          hover=${focussed}
           width="100%"
           height="32px"
         >
--- a/packages/blocks/src/_common/components/tooltip/tooltip.ts
+++ b/packages/blocks/src/_common/components/tooltip/tooltip.ts
@@ -199,21 +199,15 @@ export class Tooltip extends LitElement {
               updatePortal();
             });
 
-            const slot = this.shadowRoot?.querySelector('slot');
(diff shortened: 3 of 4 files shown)


golok727:
* **#7246** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7246?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

 Join @golok727 and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/blocks/src/_common/components/hover/controller.ts
```diff
@@ -95,17 +95,13 @@ export class HoverController implements ReactiveController {
         }
         // Transition out
         Object.assign(this._portal.style, this._hoverOptions.transition.out);
-        this._portal.addEventListener(
-          'transitionend',
-          () => {
-            abortController.abort();
-          },
-          { signal: abortController.signal }
-        );
```
lawvs, line 104: The transition works as expected, If there is no need the leave transition in most cases, try removing the `DEFAULT_HOVER_OPTIONS.transition.out` directly.

lawvs (dismissed):

packages/blocks/src/_common/components/hover/controller.ts
```diff
@@ -105,7 +105,8 @@ export class HoverController implements ReactiveController {
         return;
       }
 
-      this._abortController?.abort();
+      if (this._abortController) return;
```
L-Sun, line 108: This line of code change will sometimes make the tooltip unable to display. You can see the tooltip for `underline` 
is no longer showing.

https://github.com/toeverything/blocksuite/assets/20479050/fab68156-6dcf-40f7-8517-a8ee862d542c

golok727: > Cloud you let me know what happens occur if the abortController abort?
It is the abortController for 
https://github.com/toeverything/blocksuite/blob/c3dda83b8dc5ac74527cef3899e7754be1bdcb49/packages/blocks/src/_common/components/portal.ts#L120
Which on abort will remove the portal. 
The hover controller renders the `template` into the `portal`

golok727: > This line of code change will sometimes make the tooltip unable to display. You can see the tooltip for underline
is no longer showing.

Both the List and More menus utilize the `HoverController`. For example, when you click the More button, it opens the More menu portal. However, when you move the mouse away from the button, the whenHover function is triggered with hover = false. At this point, if `abortController.abort()` is called, it will close the portal, causing issues with the code block. The hover controller might be removed based on https://github.com/toeverything/blocksuite/pull/7248, but the existing logic should be maintained.

The problem arises because abortController.abort() is not being triggered in the transition end event due to rapid mouse movements, which should set the abortController to undefined. Removing the transition may resolve this issue.

L-Sun (approved):

PR #7248, opened 2024-06-07
fix: code-language-list widget disappear

lawvs:
Related to https://github.com/toeverything/blocksuite/pull/7245 https://github.com/toeverything/blocksuite/pull/7246

The `HoverController` is unsuitable here since the lang list should not disappear when the user hovers out.

### DIFF
--- a/packages/blocks/src/_common/components/hover/controller.ts
+++ b/packages/blocks/src/_common/components/hover/controller.ts
@@ -95,8 +95,19 @@ export class HoverController implements ReactiveController {
         }
         // Transition out
         Object.assign(this._portal.style, this._hoverOptions.transition.out);
+
+        // The transition event is not reliable,
+        // consider adding explicit duration to the transition options in the future
+        // See also https://github.com/w3c/csswg-drafts/issues/3043 https://github.com/toeverything/blocksuite/pull/7248/files#r1631375330
         this._portal.addEventListener(
           'transitionend',
+          () => {
+            abortController.abort();
+          },
+          { signal: abortController.signal }
+        );
+        this._portal.addEventListener(
+          'transitioncancel',
           () => {
             abortController.abort();
           },
@@ -105,6 +116,8 @@ export class HoverController implements ReactiveController {
         return;
       }
 
+      // If some problems arise when aborting the previous hover,
+      // consider fixing the transition related issues and return void here
       this._abortController?.abort();
       this._abortController = new AbortController();
       this._abortController.signal.addEventListener('abort', () => {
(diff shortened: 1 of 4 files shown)


packages/blocks/src/_common/components/hover/controller.ts
```diff
@@ -131,7 +133,10 @@ export class HoverController implements ReactiveController {
 
       const transition = this._hoverOptions.transition;
       if (transition) {
-        Object.assign(this._portal.style, transition.in);
+        setTimeout(() => {
+          assertExists(this._portal);
+          Object.assign(this._portal.style, transition.in);
+        });
```
lawvs, line 139: Revert the `setTimeout` to prevent sometimes the tooltip not displayed.

packages/blocks/src/_common/components/hover/controller.ts
```diff
@@ -105,7 +105,9 @@ export class HoverController implements ReactiveController {
         return;
       }
 
-      this._abortController?.abort();
```
lawvs, line 108: We should not abort it every time since `whenHover` may run its callback multiple times, which may cause the tooltip to flicker.


https://github.com/toeverything/blocksuite/blob/8032acf83e8e7dd22470801f2e18dbd3c434070b/packages/blocks/src/_common/components/hover/when-hover.ts#L10

golok727:
@lawvs Should the more menu for the code block share the same behavior?

packages/blocks/src/_common/components/hover/controller.ts
golok727: https://github.com/toeverything/blocksuite/assets/123532141/0d092e03-3318-4469-a328-9d4f354952de

I guess the problem arises because abortController.abort() is not being triggered in the transition end event due to rapid mouse movements. Which worked when I removed the transition in my PR.

L-Sun: This is becase the `transitionend` listener was remove since `transitioncancel` event happend, which make the `abortController` never abort, and  the opacity of portal is always zero.

https://developer.mozilla.org/en-US/docs/Web/API/Element/transitionend_event

image

lawvs:
> Should the more menu for the code block share the same behavior?

Ok, I will update it.

lawvs:
Since this PR is relatively complete, I will open a new PR to update the 'code-toolbar' and 'image-toolbar'.

L-Sun (approved):

golok727 (approved):

PR #7286, opened 2024-06-12
fix(presets): get image blob from selection

donteatfriedrice:


pengx17 (approved):

packages/presets/src/ai/utils/selection-utils.ts
```diff
@@ -250,9 +250,8 @@ export const getSelectedImagesAsBlobs = async (host: EditorHost) => {
     .run();
 
   const blobs = await Promise.all(
-    data.currentBlockSelections?.map(async s => {
-      const sourceId = (host.doc.getBlock(s.blockId)?.model as ImageBlockModel)
-        ?.sourceId;
+    data.selectedBlocks?.map(async b => {
```
pengx17, line 253: what is the difference between selectedBlocks & currentBlockSelections?

donteatfriedrice: selectedBlocks includes currentBlockSelections and currentImageSelections, current implementation would return null when there are only image selections.

PR #7288, opened 2024-06-12
refactor: add DocModeService for doc mode get, set, toogle and observe change

akumatus:
### TL;DR

This refactor introduces a new `DocModeService` to get document mode, set or toggle mode and observe mode changes. You can also get or set sepcific document mode by pass through id, not only the current document mode.

### Why make this change?
- The old `editorModeSwitch` slot only support manipulating current doc, can not change other doucment mode.
- The old `editorModeSwitch.on` only react to `editorModeSwitch.emit` which is not strict equal to mode change.
- Explicitly calling the `setMode` function should be more intuitive than emitting an event.

### What changed?
- Introduced `DocModeService` to manage document mode, which can be overwritten by upper business logic (eg. in playground or affine)
```typeacript
interface DocModeService {
  setMode: (mode: DocMode, docId?: string) => void;
  getMode: (docId?: string) => DocMode;
  toggleMode: (docId?: string) => DocMode;
  onModeChange: (
    handler: (mode: DocMode) => void,
    docId?: string
  ) => Disposable;
}
```
- Add a default implement of doc mode service in blocksuites
- Mock `DocModeService` with localstorage in playground
- Overwritte `DocModeService`  in [AFFiNE](https://github.com/toeverything/AFFiNE/pull/7200)
- Updated all references from `pageService.slots.editorModeSwitch` to `pageService.docModeService`
- Updated all references from `pageService.getEditorMode` to `pageService.docModeService.getMode`
- Set doc mode to edgeless after linked doc created from edgeless elements

### How to test?
1.  Lots of e2e tests include mode switch case.
2. Toggling between 'edgeless' and 'page' modes in the document editor.

### Problems
I think all the services should be injected into the constructor instead of overwritting public function during the `slots.mounted` lifecycle, which might have hidden asynchronous problems. Hope there are better solutions for this case. @Saul-Mirone 

---

### DIFF
--- a/packages/blocks/src/_common/components/ai-item/types.ts
+++ b/packages/blocks/src/_common/components/ai-item/types.ts
@@ -1,7 +1,7 @@
 import type { Chain, EditorHost, InitCommandCtx } from '@blocksuite/block-std';
 import type { TemplateResult } from 'lit';
 
-import type { EditorMode } from '../../utils/index.js';
+import type { DocMode } from '../../utils/index.js';
 
 export interface AIItemGroupConfig {
   name?: string;
@@ -13,7 +13,7 @@ export interface AIItemConfig {
   icon: TemplateResult | (() => HTMLElement);
   showWhen?: (
     chain: Chain<InitCommandCtx>,
-    editorMode: EditorMode,
+    editorMode: DocMode,
     host: EditorHost
   ) => boolean;
   subItem?: AISubItemConfig[];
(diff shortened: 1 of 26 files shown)


akumatus:
* **#7288** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7288?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

Saul-Mirone (approved):

PR #7343, opened 2024-06-17
fix: set page to page list will flush page list

Saul-Mirone:


packages/framework/store/src/store/meta.ts
```diff
@@ -162,6 +162,12 @@ export class DocCollectionMeta {
     });
   };
 
+  initialize() {
```
pengx17, line 165: might add call this in create doc is explicitly called in blocksuite?

Saul-Mirone: Done

EYHN (approved):

pengx17 (approved):

packages/framework/store/src/store/meta.ts
Saul-Mirone: Roll back after discuss with @EYHN

EYHN (approved):

PR #7351, opened 2024-06-17
fix: ignore composing event in slash menu

lawvs:
This pull request fixes an issue where the slash menu was not ignoring composing events.

Before

https://github.com/toeverything/blocksuite/assets/18554747/a5da4c83-3412-4e04-b02e-30971277dc86

After

https://github.com/toeverything/blocksuite/assets/18554747/def918aa-2649-4d30-a178-04789fef7cee

### DIFF
--- a/packages/blocks/src/root-block/widgets/slash-menu/slash-menu-popover.ts
+++ b/packages/blocks/src/root-block/widgets/slash-menu/slash-menu-popover.ts
@@ -455,6 +455,7 @@ export class InnerSlashMenu extends WithDisposable(LitElement) {
       'keydown',
       event => {
         if (this._currentSubMenu) return;
+        if (event.isComposing) return;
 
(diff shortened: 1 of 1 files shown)


L-Sun:
Could you provide a video explaining the differences before and after the changes? On my device, whether the IME is activated or not, the slash key always inputs "/".

lawvs:
> Could you provide more details explaining the differences before and after the changes? On my device, whether the IME is activated or not, the slash key always inputs '/'. Maybe other IMEs input different characters with the slash key, but we plan to support this case to make it easier to use.

The slash menu has a query function. This PR mainly fixes the user experience when using IME query. I have updated a comparison video in the PR description.

L-Sun:
> The slash menu has a query function. This PR mainly fixes the user experience when using IME query. I have updated a comparison video in the PR description.

Good catch, and it makes sense.

L-Sun (approved):

PR #7355, opened 2024-06-18
fix: keep connectors after note converted to linked doc

akumatus:
Use `doc.deleteBlock` instead of `deleteElements` to delete selected note element.

### Why make this change?
In function `deleteElements`,  the value get immediately from `service.surface.getConnectors` is outdated after modify the connector data. Because the `_elementToConnector` map is updated in the event thrown by Y.js, which is a  asynchronous process. 
@doouding @Saul-Mirone Maybe we should have a better way to update the `_elementToConnector` map.

### DIFF
--- a/packages/blocks/src/root-block/widgets/element-toolbar/more-button.ts
+++ b/packages/blocks/src/root-block/widgets/element-toolbar/more-button.ts
@@ -393,9 +393,9 @@ export class EdgelessMoreButton extends WithDisposable(LitElement) {
         other: 'new doc',
       });
       moveConnectors(element.id, cardId, this.edgeless.service);
-      // delete selected elements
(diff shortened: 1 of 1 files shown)


akumatus:
* **#7355** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7355?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

Saul-Mirone (approved):

Saul-Mirone:
> Maybe we should have a better way to update the _elementToConnector map.
Can't agree more.

doouding:
Any way to reproduce this issue? The yjs event should be emitted synchronously if the action happend in local.

PR #7384, opened 2024-06-20
feat(edgeless): support edgeless text for new toolbar

Flrande:


packages/blocks/src/root-block/edgeless/components/toolbar/mindmap/basket-elements.ts
```diff
@@ -152,24 +152,32 @@ export const textRender: DraggableTool['render'] = (
...
+      x: bound.x,
+      y: vCenter - h / 2,
+    });
+  } else {
+    const id = service.addElement(CanvasElementType.TEXT, {
+      xywh: new Bound(bound.x, vCenter - h / 2, w, h).serialize(),
+      text: new DocCollection.Y.Text(),
+    });
+
+    service.telemetryService?.track('CanvasElementAdded', {
```
CatsJuice, line 170: I think this should be moved out of if-else condition?

PR #7386, opened 2024-06-20
fix(blocks): preserve block ids when pasting

### LINKED ISSUE #6366: Block id mutates when pasting text

fourdim:
This closes #6366, closes BS-139.

This PR merges part of snapshot to the `fromPointState.model`. The first snapshot is deleted after being merged to the `fromPointState.model`, so the snapshot could probably be an empty array.
We delete snapshot from snapshots instead of block from blocks mainly because that if blocks are deleted immediately after they are created will cause errors as operations in `connectcallback` does not have any corresponding blocks attached.

### DIFF
--- a/packages/blocks/src/root-block/clipboard/middlewares/paste.ts
+++ b/packages/blocks/src/root-block/clipboard/middlewares/paste.ts
@@ -17,8 +17,8 @@ import {
 } from '@blocksuite/store';
 
 import { matchFlavours } from '../../../_common/utils/index.js';
-import type { CodeBlockModel } from '../../../code-block/index.js';
 import type { ParagraphBlockModel } from '../../../paragraph-block/index.js';
+import type { QuickSearchService } from '../../root-service.js';
 
 const findLast = (snapshot: BlockSnapshot): BlockSnapshot => {
   if (snapshot.children && snapshot.children.length > 0) {
@@ -61,11 +61,16 @@ class PasteTr {
 
   private readonly to: TextRangePoint | null;
 
-  private readonly firstSnapshot: BlockSnapshot;
+  private firstSnapshot?: BlockSnapshot;
 
-  private lastSnapshot: BlockSnapshot;
+  private lastSnapshot?: BlockSnapshot;
 
   private readonly firstSnapshotIsPlainText: boolean;
+
+  // The model that the cursor should focus on after pasting
+  private pasteStartModel: BlockModel | null = null;
+
+  private pasteStartModelChildrenCount = 0;
 
   constructor(
     readonly std: EditorHost['std'],
@@ -92,7 +97,23 @@ class PasteTr {
       temp.set('text', text.yText);
       this.lastIndex = text.length;
     } else {
-      this.lastIndex = this.endPointState.text.length - end.index - end.length;
+      this.lastIndex =
+        this.fromPointState.point.index +
+        this.snapshot.content
+          .map(snapshot =>
+            this._textFromSnapshot(snapshot)
+              .delta.map(op => {
+                if (op.insert) {
+                  return op.insert.length;
+                } else if (op.delete) {
+                  return -op.delete;
+                } else {
+                  return 0;
+                }
+              })
+              .reduce((a, b) => a + b, 0)
+          )
+          .reduce((a, b) => a + b + 1, -1);
     }
     this.firstSnapshotIsPlainText =
       this.firstSnapshot.flavour === 'affine:paragraph' &&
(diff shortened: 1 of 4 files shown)


fourdim:
* **#7386** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7386?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

 Join @fourdim and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

Saul-Mirone:
Can you please also re-check the `clipboard.spec.ts`? we have several `test.skip` now and I think it's time to decide if we should fix or remove them.

packages/blocks/src/root-block/clipboard/middlewares/paste.ts
```diff
@@ -127,14 +151,16 @@ class PasteTr {
...
-    if (toLanguage !== 'Plain Text') {
-      this.firstSnapshot.props.language = toLanguage;
-    }
-    const deltas: DeltaOperation[] = [...fromDelta];
+    const { toDelta } = this._getDeltas();
+    const deltas: DeltaOperation[] = [
+      { retain: this.fromPointState.point.index },
+      this.fromPointState.text.length - this.fromPointState.point.index
+        ? {
+            delete:
```
Saul-Mirone, line 159: :+1:

packages/blocks/src/root-block/clipboard/middlewares/paste.ts
```diff
@@ -148,62 +174,139 @@ class PasteTr {
...
+    );
+  };
+
+  private async _transformLinkDelta(
+    delta: DeltaOperation[],
+    linkToDocId: Map<string, string | null>,
+    quickSearchService: QuickSearchService
+  ) {
+    const needToConvert = new Map<DeltaOperation, string>();
+    for (const op of delta) {
```
Saul-Mirone, line 254: Let's use `forEach` here.

packages/blocks/src/root-block/clipboard/middlewares/paste.ts
```diff
@@ -219,50 +322,73 @@ class PasteTr {
...
 
-    const lastModel = this.std.doc.getBlockById(this.lastSnapshot.id);
-    assertExists(lastModel);
-
-    if (!this.to) {
-      this.std.doc.deleteBlock(this.fromPointState.model, {
-        bringChildrenTo: lastModel,
-      });
+    if (this.to) {
+      const [_, context] = this.std.command
```
Saul-Mirone, line 326: Can be shorten with:

```ts
this.std.command.exec('getSelectedModels', { types: ['text'] })
```

packages/blocks/src/root-block/clipboard/middlewares/paste.ts
```diff
@@ -407,10 +499,21 @@ export const pasteMiddleware = (std: EditorHost['std']): JobMiddleware => {
       }
     });
     slots.afterImport.on(payload => {
+      if (payload.type === 'slice') {
+        const [_, context] = std.command
```
Saul-Mirone, line 503: Same here

tests/clipboard.spec.ts
```diff
@@ -1373,7 +1373,9 @@ test(scoped`clipboard copy multi selection`, async ({ page }) => {
   await focusRichText(page, 1);
   await pasteByKeyboard(page);
   await waitNextFrame(page);
-  await assertRichTexts(page, ['abc', 'defbc', 'd']);
+  await type(page, 'cursor');
+  await waitNextFrame(page);
+  await assertRichTexts(page, ['abc', 'defbc', 'dcursor']);
 });
```
Saul-Mirone, line 1380: I think we need to add more test cases to prevent furthur changes break the current behavior.

tests/clipboard.spec.ts
```diff
@@ -1373,7 +1373,9 @@ test(scoped`clipboard copy multi selection`, async ({ page }) => {
...
 
 test.skip(scoped`clipboard copy nested items`, async ({ page }) => {
```
Saul-Mirone, line 1381: Can we recheck the skipped cases? Either re-enable them, remove them or fix them.

Saul-Mirone (approved):

packages/blocks/src/root-block/clipboard/middlewares/paste.ts
fourdim: According to [`no-misused-promises`](https://typescript-eslint.io/rules/no-misused-promises/):
```ts
// incorrect
[1, 2, 3].forEach(async value => {
  await fetch(`/${value}`);
});

// correct
for (const value of [1, 2, 3]) {
  await doSomething(value);
}
```

Saul-Mirone: Didn't notice the await here, sorry

fourdim:
More test cases will be added in a separate PR.

PR #7395, opened 2024-06-22
fix: add meta init function

Flrande:


Saul-Mirone (approved):

doodlewind (commented):
This may require being documented.

PR #7400, opened 2024-06-24
fix(edgeless): can not undo block moving

Flrande:
Close BS-630

### DIFF
--- a/packages/framework/store/src/store/doc/block.ts
+++ b/packages/framework/store/src/store/doc/block.ts
@@ -269,15 +269,7 @@ export class Block {
           }
 
           const yValue = native2Y(value);
-
-          if (this.doc) {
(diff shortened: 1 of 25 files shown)


packages/framework/store/src/store/doc/block.ts
```diff
@@ -285,7 +293,13 @@ export class Block {
           typeof p === 'string' &&
           model.keys.includes(p)
         ) {
-          this.yBlock.delete(`prop:${p}`);
+          if (this.doc) {
```
Saul-Mirone, line 296: What if a transaction need to include multiple operations? This doesn't make sense.

Saul-Mirone (changes requested):

Saul-Mirone (approved):

PR #7431, opened 2024-06-26
fix(edgeless): disable sync doc in edgeless text

Flrande:
Close BS-629

### DIFF
--- a/packages/blocks/src/_common/inline/presets/nodes/reference-node/reference-popup.ts
+++ b/packages/blocks/src/_common/inline/presets/nodes/reference-node/reference-popup.ts
@@ -10,6 +10,7 @@ import { html, LitElement, nothing } from 'lit';
 import { customElement, property, query } from 'lit/decorators.js';
 
 import type { RootBlockComponent } from '../../../../../root-block/types.js';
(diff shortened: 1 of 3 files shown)


packages/blocks/src/_common/inline/presets/nodes/reference-node/reference-popup.ts
```diff
@@ -10,6 +10,7 @@ import { html, LitElement, nothing } from 'lit';
 import { customElement, property, query } from 'lit/decorators.js';
 
 import type { RootBlockComponent } from '../../../../../root-block/types.js';
+import { insideEdgelessText } from '../../../../../root-block/widgets/slash-menu/utils.js';
```
L-Sun, line 13: I think use `isInsideBlockByFlavour` from `../../../../utils/index.js` is better than use `inseideEdgelessText` from `slash-menu/utils.js`

L-Sun (approved):

Saul-Mirone:
Need a test

Saul-Mirone (approved):

Flrande:
I will improve in the next PR

PR #7436, opened 2024-06-26
fix(blocks): edgeless element toolbar more button error

donteatfriedrice:


donteatfriedrice:
Error no reproduction.

pengx17 (approved):

PR #7440, opened 2024-06-27
fix(edgeless): edgeless portal does not render new added blocks

doouding:
Fixes [BS-685](https://linear.app/affine-design/issue/BS-685/edgeless-copy-paste-error)


doouding:
* **#7440** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7440?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

Saul-Mirone:
Can we add a test case for this?

L-Sun:
I feel that the code logic before and after the change is almost the same. What caused this bug? 🤔

doouding:
> I feel that the code logic before and after the change is almost the same. What caused this bug? 🤔

I think lit cannot handle the `repeat` rendering in nested loop correctly. So I flat the array first.

doouding:
> Can we add a test case for this?

Done.

Saul-Mirone (approved):

PR #7443, opened 2024-06-27
fix: edgeless selection may disappear when typing on edgeless text

Flrande:
Close BS-652

https://github.com/toeverything/blocksuite/assets/50035259/3e663b76-49bb-4c06-b50f-5bac382dc3e3

### DIFF
--- a/packages/framework/block-std/src/view/utils/range-binding.ts
+++ b/packages/framework/block-std/src/view/utils/range-binding.ts
@@ -219,7 +219,7 @@ export class RangeBinding {
       },
       to: null,
     });
-    this.selectionManager.set([newSelection]);
+    this.selectionManager.setGroup('note', [newSelection]);
   };
 
   private _onCompositionStart = () => {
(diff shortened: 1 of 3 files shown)


packages/framework/block-std/src/view/utils/range-binding.ts
```diff
@@ -298,7 +298,7 @@ export class RangeBinding {
         },
         to: null,
       });
-      this.host.selection.set([selection]);
+      this.host.selection.setGroup('note', [selection]);
```
Saul-Mirone, line 301: Do we have better solution? I want to move the `note` group from std to blocks. Not sure if it's possible.

Saul-Mirone (approved):

packages/framework/block-std/src/view/utils/range-binding.ts
Flrande: I think it's fine, but before that, we need to improve the type of the group first, now it's all hard-coded.

PR #7461, opened 2024-07-01
refactor(edgeless): refine edgeless block definition

doouding:
## Edgeless block usage guide

To define a edgeless block, you need to use `selectable` and `EdgelessBlockElement` to define you model and rendering element:
```typescript
import { BlockModel, defineBlockSchema } from '@blocksuite/store';
import { EdgelessBlockElement, BlockElement, selectable, SerializedXYWH } from '@blocksuite/block-std';

// define your model schema first
type ExampleProps = {
     xywh: SerializedXYWH;
     index: string
}

defineBlockSchema({
     flavour: 'affine:example',
     props: () => ({
        xywh: '[0,0,0,0]',
        index: 'a0'
     }),
     toModel: () => new EdgelessExampleBlock()
})

// edgeless block model should extend the `BlockModel` wrapped with `selectable`
class ExampleBlockModel extends selectable<ExampleProps>(BlockModel) {}



/**
 * Define how to render that block in dom.
 *
 * RootService is a temporary solution that author may need to access some API in edgeless which haven't been migrated to std. It can provide service of other block. Access through `affine:root`;
 */
export ExampleBlock extends EdgelessBlockElement<EdgelessRootService, ExampleBlockModel> {
    // tell the block which blocks' service is used at rootService
    rootServiceFlavour!: 'affine:page';

    xxxx() {
        this.rootService.xxxx;
    }

    /**
     * Usually the block will parse model's `xywh` and `index` during the rendering process.
     * If you need to return xywh other than model's xywh property. You can override this method.
     */
    override getRenderingRect() {}

    /**
     * Define how your block should be rendered
     */
    override renderEdgelessBlock() {
      return html`<div class="example-block"></div>`
    }
}
```

### May change in the future
- [ ] new package `edgeless-std`


doouding:
* **#7461** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7461?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/framework/block-std/src/index.ts
```diff
@@ -1,5 +1,6 @@
 export * from './clipboard/index.js';
 export * from './command/index.js';
+export * from './edgeless/index.js';
```
doodlewind, line 3: I'd recommend only exporting the necessary types here, since there are types like `IHitTestOptions` that are not supposed to be reused by `block-std` users.

packages/framework/block-std/src/edgeless/bound.ts
```diff
@@ -0,0 +1,298 @@
...
+  EPSILON,
+  getBoundsFromPoints,
+  lineIntersects,
+  polygonPointDistance,
+} from '../utils/math.js';
+import { deserializeXYWH, serializeXYWH } from '../utils/xywh.js';
+import type { IBound, SerializedXYWH, XYWH } from './types.js';
+import type { IVec } from './vec.js';
+
+export class Bound implements IBound {
```
Saul-Mirone, line 11: Do you want to use `signal` to build this? properties like points seems can be auto computed.

doouding: I would like to, but not in this PR.

doodlewind (approved):
Thanks for the feat!

packages/framework/block-std/src/view/element/edgeless-block-element.ts
```diff
@@ -0,0 +1,172 @@
...
+      throw new Error('Edgeless block should have at least `xywh` property.');
+    }
+
+    const [x, y, w, h] = JSON.parse(xywh);
+
+    return { x, y, w, h, zIndex: this.toZIndex() };
+  }
+
+  override renderBlock() {
+    const { xywh, index } = this.model as BlockModel<{
```
Saul-Mirone, line 42: You can use: 
```ts
const { xywh$, index$ } = this.model;

if (!xywh$.value || !index$.value) { ... }
```

This reactive prop will be updated automatically without the subscription to `propsUpdated`.

doouding: Got it. I will refine these part in the next PRs.

packages/framework/block-std/src/edgeless/model.ts
```diff
@@ -0,0 +1,205 @@
...
+  get externalXYWH(): SerializedXYWH | undefined {
+    return this._externalXYWH;
+  }
+
+  set externalXYWH(xywh: SerializedXYWH | undefined) {
+    this._externalXYWH = xywh;
+  }
+
+  get group(): IEdgelessElement | null {
+    const surface = this.doc.getBlocksByFlavour('affine:surface')[0]?.model;
```
Saul-Mirone, line 156: This is very strange. If you really need to do something like this, you should add a flag on the block model instead of get a specific model flavour.

Saul-Mirone:
```ts
class SurfaceModel extends BlockModel {
  [edgelessSurfaceRoot]: true
}

const surface = Object.values(this.doc.getBlocks()).find(...)
```

packages/framework/block-std/src/edgeless/model.ts
```diff
@@ -0,0 +1,205 @@
...
+
+    if (!surface) return null;
+
+    // FIXME: make surface a official supported block
+    // @ts-ignore
+    return surface.getGroup(this.id) ?? null;
+  }
+
+  get groups(): IEdgelessElement[] {
+    const surface = this.doc.getBlocksByFlavour('affine:surface')[0]?.model;
```
Saul-Mirone, line 166: Same here.

PR #7473, opened 2024-07-02
docs: switch to stackblitz for online examples

doodlewind:
Remaining examples that are not compatible with StackBlitz yet:

* `react-basic-next`
* `angular-basic`


sikaco:
StackBlitz seems couldn't be visit now.

PR #7476, opened 2024-07-02
feat(edgeless): support using last props for edgeless text

Flrande:
Support memory for style changes to edgeless text and optimize some naming in `EditPropsStore`.

Before:

https://github.com/toeverything/blocksuite/assets/50035259/cd9656c9-b79c-427a-98db-dc0484ec4295

After:

https://github.com/toeverything/blocksuite/assets/50035259/cbfb6c45-7768-4f73-9952-88714c25d89b

### DIFF
--- a/packages/blocks/src/edgeless-text/edgeless-text-service.ts
+++ b/packages/blocks/src/edgeless-text/edgeless-text-service.ts
@@ -1,9 +1,7 @@
 import { BlockService } from '@blocksuite/block-std';
 
 import { asyncFocusRichText } from '../_common/utils/selection.js';
-import { GET_DEFAULT_TEXT_COLOR } from '../root-block/edgeless/components/panel/color-panel.js';
 import type { EdgelessRootBlockComponent } from '../root-block/index.js';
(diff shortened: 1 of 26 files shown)


doouding (approved):
I think the previous API would be neater because calling the method in editPropsStore already implies the context.

Flrande:
> I think the previous API would be neater because calling the method in `editPropsStore` already implies the context.

I made this change because when I first used this API, I didn't know what the behaviours of `record` and `apply` were, nor did I understand what `item` and `storage` referred to. It was only after reading the original code that I figured out how to use it. Maybe we should separate `storage` and `lastProps`?🤔

doouding:
> > I think the previous API would be neater because calling the method in `editPropsStore` already implies the context.
> 
> I made this change because when I first used this API, I didn't know what the behaviours of `record` and `apply` were, nor did I understand what `item` and `storage` referred to. It was only after reading the original code that I figured out how to use it. Maybe we should separate `storage` and `lastProps`?🤔

Fair enough, this justifies the change.

PR #7478, opened 2024-07-02
fix(blocks): slash menu not closable in peek view

pengx17:


pengx17:
* **#7478** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7478?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

 Join @pengx17 and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/blocks/src/root-block/widgets/slash-menu/styles.ts
```diff
@@ -5,6 +5,7 @@ import { scrollbarStyle } from '../../../_common/components/utils.js';
 
 export const styles = css`
   .overlay-mask {
+    pointer-events: auto;
```
L-Sun, line 8: because the peek view container has `pointer-events: none;` ?

L-Sun (approved):

packages/blocks/src/root-block/widgets/slash-menu/styles.ts
pengx17: Yes. I think the issue is caused by this line:
https://github.com/radix-ui/primitives/blob/main/packages/react/dismissable-layer/src/DismissableLayer.tsx#L117

When radixui dialog opens, it will applies pointer-events: none to body. All clickable elements should be explicitly marked as pointer-events: auto

PR #7486, opened 2024-07-03
fix(edgeless): edgeless text shake when dragging

Flrande:
Close BS-730
Close BS-722

Before:

https://github.com/toeverything/blocksuite/assets/50035259/c5a59ecf-3534-451b-8206-c86b750ccb1f


Saul-Mirone (approved):
Is it possible to add test for this fix?

Flrande:
I will add it later.

PR #7493, opened 2024-07-04
feat: use signal for block model children

Saul-Mirone:


packages/framework/store/src/schema/base.ts
```diff
@@ -206,6 +213,35 @@ export class BlockModel<
...
+    super();
+    this._onCreated = this.created.once(() => {
+      this._children.value = this.yBlock.get('sys:children').toArray();
+      this.yBlock.get('sys:children').observe(event => {
+        this._children.value = event.target.toArray();
+      });
+    });
+    this._onDeleted = this.deleted.once(() => {
+      this._onCreated.dispose();
+    });
```
pengx17, line 237: I think these two may be initialized in class properties as well instead of being syntactically in the constructor.

packages/framework/store/src/schema/base.ts
```diff
@@ -214,33 +250,18 @@ export class BlockModel<
...
-    }
-
-    const children: BlockModel[] = [];
-    block.forEach(id => {
-      const child = this.doc.getBlockById(id);
-      if (!child) {
-        return;
-      }
-      children.push(child);
+    const computedChildren = computed(() => {
```
pengx17, line 254: it seems this is a memleak. A new watcher will be created on every `children` get.

packages/blocks/src/_common/utils/query.ts
```diff
@@ -822,7 +822,7 @@ export function getDropRectByPoint(
     const table = getDatabaseBlockTableElement(element);
     assertExists(table);
     let bounds = table.getBoundingClientRect();
-    if (model.isEmpty()) {
+    if (model.isEmpty.value) {
```
pengx17, line 825: from preact/signal doc, it looks like the recommended way of reading values is via `.peek()`

packages/framework/store/src/schema/base.ts
Saul-Mirone: Is there any special reason for this?

packages/framework/store/src/schema/base.ts
Saul-Mirone: At first I use one signal that computed from `this._children`, however I found a problem that when y events emitted, the block instance haven't been created so the `if` in line 258 will get an `undefined`. I think we need to refactor the data flow to fix this problem. But I have no idea how to do that.

packages/blocks/src/_common/utils/query.ts
Saul-Mirone: Done

packages/framework/store/src/schema/base.ts
pengx17: I think .value actually subscribes to the signal in one doc that I read. Let me double check.

packages/framework/store/src/schema/base.ts
pengx17: two ideas:
1.

Saul-Mirone: Not really, once a block subscribe to the children, the future changes will notify the subscription. I think that's how the new `block-element` update the children without listener to `childrenUpdated` slot.

pengx17: Are you sure about that? getter like `get children` will only evaluate when it is being read, then it is done. There will be no way to make it an reactive attribute that allows it to be reactive to changes in the internal signals.
I guess it is being revaluated because this value is being read again that caused by some other effects.

Saul-Mirone: Not really sure, my guessing is that it will be subscribed when read because I subscribe it in `blockElement` with a [`SignalWatcher`](https://www.npmjs.com/package/@lit-labs/preact-signals#signalwatcher). I need to confirm the behavior.

Saul-Mirone: Based on our offline discussion. It's working in lit and still be a signal subscribed by the lit element. But the computed timing decides whether it can get the correct children list. It's a major discovery and we'll discuss more in dev committee.

Saul-Mirone:
* **#7500** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7500?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7498** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7498?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7495** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7495?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7494** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7494?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7493** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7493?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

doodlewind (approved):

PR #7500, opened 2024-07-05
feat: signal based placeholder

Saul-Mirone:


Saul-Mirone:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7500?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#7500** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7500?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#7498** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7498?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7495** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7495?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7494** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7494?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7493** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7493?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

packages/framework/block-std/src/selection/manager.ts
```diff
@@ -177,24 +177,28 @@ export class SelectionManager {
...
+  }
 
-    return computedFind.value;
+  find<T extends BlockSuite.SelectionType>(type: T) {
+    return this.find$(type).value;
   }
 
-  filter<T extends BlockSuite.SelectionType>(type: T) {
-    const computedFilter = computed(() =>
+  filter$<T extends BlockSuite.SelectionType>(type: T) {
```
Flrande, line 192: Will this naming convention become a code standard in the future?

Flrande (approved):

packages/framework/block-std/src/selection/manager.ts
Saul-Mirone: Not really, if we decide to move to signal. We will use signal compute to replace the old `filter` and `find`.

Flrande: LGTM, do we need to mark the old method as deprecated?

Saul-Mirone: Not yet, I'm still not sure if we will use signal in the end. Let's discuss in dev committee.

Flrande: OK

packages/blocks/src/paragraph-block/paragraph-block.ts
```diff
@@ -73,36 +74,10 @@ export class ParagraphBlockComponent extends BlockComponent<
   @query('.affine-paragraph-placeholder')
   private accessor _placeholderContainer: HTMLElement | null = null;
 
-  private _currentTextSelection: TextSelection | undefined = undefined;
+  private _displayPlaceholder!: ReadonlySignal<boolean>;
 
   override accessor blockContainerStyles = { margin: '10px 0' };
 
-  //TODO(@Flrande) wrap placeholder in `rich-text` or inline-editor to make it more developer-friendly
-  private _updatePlaceholder = () => {
```
doodlewind, line 81: So the benefit of using signal here should be removing more imperative `_updateSomething` methods across the code base?

Also looks great to avoid `updateComplete`! Not sure if this is something can be refactored easily, or requires case by case debugging in time sequence?

doodlewind (approved):

packages/blocks/src/paragraph-block/paragraph-block.ts
Saul-Mirone: With the selection signal. Blocks can compute the correct state based on the selection and has mainly 3 benefits:
1. Developers won't need to subscribe to selection state. The compute based pattern will make more straight forward. 
2. Developers won't need to cared about component update triggered by changes (@lit-labs/preact-signal will handle that)
3. The performance will be better because compute is lazy.

However, this cannot be auto refactored by something like a codemod. Although it's not hard to refactor, it still needs us to refactor the code progressively.

PR #7530, opened 2024-07-09
fix: hover style of icon button

akumatus:
Fix issue [BS-721](https://linear.app/affine-design/issue/BS-721).

Cause we have css style like:
```
:host([hover='false']) {
  background: transparent;
}
```
If the `hover` is used as boolean attribute binding, the attribute will be added or removed based on the boolean value. Which means `host([hover='false'])` style will never happened to work.

### What changed?
- Make the `hover` attribute treated as a string
- Use standard attribute binding instead of boolean attribute binding or property binding

### DIFF
--- a/packages/blocks/src/_common/components/button.ts
+++ b/packages/blocks/src/_common/components/button.ts
@@ -58,7 +58,7 @@ export class IconButton extends LitElement {
     }
 
     /* You can add a 'hover' attribute to the button to show the hover style */
-    :host([hover]) {
+    :host([hover='true']) {
       background: var(--affine-hover-color);
     }
     :host([hover='false']) {
@@ -128,8 +128,8 @@ export class IconButton extends LitElement {
   @property({ attribute: true, type: Boolean })
   accessor active: boolean = false;
 
-  @property({ attribute: true, type: Boolean })
-  accessor hover: boolean | undefined = undefined;
+  @property({ attribute: true, type: String })
+  accessor hover: 'true' | 'false' | undefined = undefined;
 
   // Do not add `{ attribute: false }` option here, otherwise the `disabled` styles will not work
   @property({ attribute: true, type: Boolean })
--- a/packages/blocks/src/_common/inline/presets/nodes/reference-node/reference-popup.ts
+++ b/packages/blocks/src/_common/inline/presets/nodes/reference-node/reference-popup.ts
@@ -274,7 +274,7 @@ export class ReferencePopup extends WithDisposable(LitElement) {
               <icon-button
                 size="24px"
                 class="affine-reference-popover-view-selector-button link current-view"
-                .hover=${false}
+                hover=${false}
               >
                 ${LinkIcon}
                 <affine-tooltip .offset=${12}>${'Inline view'}</affine-tooltip>
@@ -283,7 +283,7 @@ export class ReferencePopup extends WithDisposable(LitElement) {
               <icon-button
                 size="24px"
                 class="affine-reference-popover-view-selector-button embed card-view"
-                .hover=${false}
+                hover=${false}
                 @click=${() => this._convertToCardView()}
               >
                 ${BookmarkIcon}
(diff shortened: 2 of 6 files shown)


akumatus:
* **#7530** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7530?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/blocks/src/_common/inline/presets/nodes/reference-node/reference-popup.ts
```diff
@@ -283,7 +283,7 @@ export class ReferencePopup extends WithDisposable(LitElement) {
               <icon-button
                 size="24px"
                 class="affine-reference-popover-view-selector-button embed card-view"
-                .hover=${false}
+                hover=${false}
```
Saul-Mirone, line 286: Not a lit expert, what's the difference between `hover`, `.hover` and `?hover` ?

akumatus: In Lit, you can bind properties, attributes, and boolean attributes in templates using specific syntax. The differences between `.hover=${}`, `?hover=${}`, and `hover=${}` are related to how these bindings are applied to the elements in your template.

### `hover=${}`

This is a standard attribute binding. It sets the attribute on the element to the given value.

#### Example
```javascript
import { LitElement, html, css, customElement } from 'lit';

@customElement('my-element')
class MyElement extends LitElement {
  render() {
    const hoverValue = 'hovered';
    return html`<div hover=${hoverValue}>Content</div>`;
  }
}
```

In this example:
- The `hover` attribute is set to the value of `hoverValue` (`'hovered'`).
- This results in `<div hover="hovered">Content</div>` being rendered.

### `.hover=${}`

This syntax binds a property of the element to the given value, rather than setting an attribute. This is used when you want to directly set a property on the DOM element.

#### Example
```javascript
import { LitElement, html, css, customElement } from 'lit';

@customElement('my-element')
class MyElement extends LitElement {
  render() {
    const hoverValue = true;
    return html`<div .hover=${hoverValue}>Content</div>`;
  }
}
```

In this example:
- The `hover` property of the `div` element is set to `true`.
- This does not reflect in the HTML as an attribute but sets the DOM property directly.

### `?hover=${}`

This syntax is used for boolean attributes. If the value is true, the attribute is added; if false, the attribute is removed.

#### Example
```javascript
import { LitElement, html, css, customElement } from 'lit';

@customElement('my-element')
class MyElement extends LitElement {
  render() {
    const isHovered = true;
    return html`<div ?hover=${isHovered}>Content</div>`;
  }
}
```

In this example:
- If `isHovered` is `true`, the `hover` attribute is present on the `div` element (`<div hover>Content</div>`).
- If `isHovered` is `false`, the `hover` attribute is not present on the `div` element.

### Summary

- **`hover=${}`**: Standard attribute binding. Sets the attribute to the given value as a string.
- **`.hover=${}`**: Property binding. Sets the property on the element to the given value.
- **`?hover=${}`**: Boolean attribute binding. Adds or removes the attribute based on the boolean value.

Each of these bindings serves different purposes and allows you to interact with elements in specific ways. Understanding these differences helps you effectively manage how attributes and properties are applied in your Lit templates.

akumatus: I'm not a lit expert either, this is what I learned from chatgpt ;)

golok727 (approved):

packages/blocks/src/_common/components/button.ts
```diff
@@ -128,8 +128,8 @@ export class IconButton extends LitElement {
   @property({ attribute: true, type: Boolean })
   accessor active: boolean = false;
 
-  @property({ attribute: true, type: Boolean })
-  accessor hover: boolean | undefined = undefined;
+  @property({ attribute: true, type: String })
+  accessor hover: 'true' | 'false' | undefined = undefined;
```
fundon, line 132: Or adds fields:
```
reflect: true,
converter: {
  fromAttribute,
  toAttriubute,
}
```

golok727: Goood

golok727: Many parts of the codebase use hover=... instead of ?hover as a workaround. We should change that as well. cc @akumatus

PR #7535, opened 2024-07-10
fix(database): readonly mode

zzj3720:
fix: BS-764


golok727:
I think we should hide the drag handle. Notion does the same. 
image

Saul-Mirone (approved):

packages/blocks/src/database-block/data-view/view/presets/table/components/row.ts
```diff
@@ -225,7 +228,8 @@ export class TableRow extends WithDisposable(ShadowlessElement) {
               >
               </affine-database-cell-container>
             </div>
-            ${column.dataViewManager.header.titleColumn === column.id
+            ${!column.readonly &&
```
Saul-Mirone, line 231: Maybe we can have some tests for this case.

PR #7565, opened 2024-07-12
feat(store): add signal props for model

Saul-Mirone:
Add signal props for model.

This pr automatically add a signal version property for every prop

```ts
declare const model: BlockModel<{ name: string }>;
assert(typeof model.name$.value === 'string'); // true
 
model.name = 'hello';
assert(model.name$.value === 'hello'); // true
assert(model.yBlock.get('prop:name') === 'hello'); // true

model.name$.value = 'world';
assert(model.name === 'world'); // true
assert(model.yBlock.get('prop:name') === 'world'); // true

model.yBlock.set('prop:name', 'blocksuite');
assert(model.name$.value === 'blocksuite'); // true
```

### DIFF
--- a/packages/blocks/src/paragraph-block/paragraph-block.ts
+++ b/packages/blocks/src/paragraph-block/paragraph-block.ts
@@ -73,7 +73,7 @@ export class ParagraphBlockComponent extends BlockComponent<
   }
 
   override renderBlock(): TemplateResult<1> {
-    const { type } = this.model;
+    const { type$ } = this.model;
     const children = html`<div
       class="affine-block-children-container"
       style="padding-left: ${BLOCK_CHILDREN_CONTAINER_PADDING_LEFT}px"
@@ -83,7 +83,7 @@ export class ParagraphBlockComponent extends BlockComponent<
 
     return html`
       <div class="affine-paragraph-block-container">
-        <div class="affine-paragraph-rich-text-wrapper ${type}">
+        <div class="affine-paragraph-rich-text-wrapper ${type$.value}">
           <rich-text
             .yText=${this.model.text.yText}
             .inlineEventSource=${this.topContenteditableElement ?? nothing}
--- a/packages/framework/store/src/__tests__/block.unit.spec.ts
+++ b/packages/framework/store/src/__tests__/block.unit.spec.ts
@@ -1,4 +1,5 @@
-import { expect, test } from 'vitest';
+import { computed, effect } from '@preact/signals-core';
+import { describe, expect, test, vi } from 'vitest';
 import * as Y from 'yjs';
 
 import {
@@ -14,6 +15,7 @@ const pageSchema = defineBlockSchema({
   props: internal => ({
     title: internal.Text(),
     count: 0,
+    toggle: false,
     style: {} as Record<string, unknown>,
   }),
   metadata: {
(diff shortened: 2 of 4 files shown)


Saul-Mirone:
* **#7565** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7565?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/framework/store/src/__tests__/block.unit.spec.ts
```diff
@@ -55,3 +57,142 @@ test('init block without props should add default props', () => {
...
+    yBlock.set('sys:flavour', 'page');
+    yBlock.set('sys:children', new Y.Array());
+
+    const block = new Block(doc.schema, yBlock, doc);
+    const model = block.model as RootModel;
+
+    const isOdd = computed(() => model.count$.value % 2 === 1);
+
+    expect(model.count$.value).toBe(0);
+    expect(isOdd.peek()).toBe(false);
```
Flrande, line 76: Just a question, why do we need `peek` here.

Flrande (approved):
Will we replace proxy with signal in the future?

packages/framework/store/src/__tests__/block.unit.spec.ts
Saul-Mirone: We don't. I just test all the functionalities.

Saul-Mirone:
> Will we replace proxy with signal in the future?

I think we won't because proxy can be updated partially. But we'll deprecate `propsUpdated` and `childrenUpdated`.

packages/blocks/src/paragraph-block/paragraph-block.ts
```diff
@@ -83,7 +83,7 @@ export class ParagraphBlockComponent extends BlockComponent<
 
     return html`
       <div class="affine-paragraph-block-container">
-        <div class="affine-paragraph-rich-text-wrapper ${type}">
+        <div class="affine-paragraph-rich-text-wrapper ${type$.value}">
```
L-Sun, line 86: Is `type$.value` not same as `type` if I want to make render reactive?

Saul-Mirone: type$.value will be automatically subscirbed by [SignalWatcher](https://www.npmjs.com/package/@lit-labs/preact-signals#signalwatcher). So you won't need to subscribe to `model.propsUpdated` slot. That's the difference.

L-Sun: So my understanding is that although the data of `model.prop`, `model.prop$`, and `model.Y` are synchronized, using model.prop directly in render does not allow for reactive updates. This is because `model.prop` is not a getter of  `model.prop$.value`.

Saul-Mirone: Yep, model.prop won't be a reactive value and you'll need to re-get it manually when props updated. However, things will be different if `model.prop` is an `object` or `array`, for example:

```ts
interface ModelStyle {
  light: {
    color: string
  },
  dark: {
    color: string
  }
}

// This will trigger a update and the result will be synced to `model.$style` and yjs data
model.style.light.color = 'black';
```

PR #7567, opened 2024-07-12
fix: placeholder after image

Saul-Mirone:


packages/framework/block-std/src/view/element/block-element.ts
```diff
@@ -22,6 +22,18 @@ export class BlockElement<
...
+      return selection.blockId === this.model.id;
+    });
+
+    if (!selection) {
+      return null;
+    }
+
+    return selection;
+  });
+
```
L-Sun, line 36: I have a question, when was this signal first calculated?

L-Sun (approved):

packages/framework/block-std/src/view/element/block-element.ts
L-Sun: Got it. In the first access its value

PR #7591, opened 2024-07-16
fix: cannot drag embed block to edgeless surface

donteatfriedrice:
[BS-868](https://linear.app/affine-design/issue/BS-868/无法拖拽-embed-block-到白板)


Saul-Mirone (approved):
LGTM

PR #7597, opened 2024-07-17
feat: ai chat block peek view

donteatfriedrice:
[BS-565](https://linear.app/affine-design/issue/BS-565/支持-ai-chat-block-在-center-peek-中展示)

### DIFF
--- a/packages/presets/src/blocks/_common/utils.ts
+++ b/packages/presets/src/blocks/_common/utils.ts
@@ -26,3 +26,30 @@ export async function markDownToDoc(host: EditorHost, answer: string) {
   }
   return doc as Doc;
 }
+
+export function formatDate(dateStr: string): string {
+  const date = new Date(dateStr);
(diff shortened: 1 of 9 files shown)


packages/presets/src/blocks/_common/utils.ts
```diff
@@ -26,3 +26,30 @@ export async function markDownToDoc(host: EditorHost, answer: string) {
   }
   return doc as Doc;
 }
+
+export function formatDate(dateStr: string): string {
```
pengx17, line 30: maybe use i18nTime from '@affine/i18n';

pengx17 (approved):

packages/presets/src/blocks/_common/utils.ts
donteatfriedrice: Get it. I will update it when move these AI Chat Block related components to AFFiNE.

PR #7598, opened 2024-07-17
feat(blocks): improve order role for list

Flrande:


packages/blocks/src/list-block/list-model.ts
```diff
@@ -28,6 +36,29 @@ export const ListBlockSchema = defineBlockSchema({
...
-export type ListBlockModel = SchemaToModel<typeof ListBlockSchema>;
+export class ListBlockModel extends BlockModel<ListProps> {
+  override flavour!: 'affine:list';
+
+  override text!: Text;
+
+  constructor() {
+    super();
+
+    this.propsUpdated.on(({ key }) => {
```
Saul-Mirone, line 50: instead of use `this.propsUpdated`, use

```ts
const dispose = effect(() => {
  const type = this.model.$type.value;
  handleType(type);
})
this.deleted.once(dispose);
```

Saul-Mirone (approved):

packages/blocks/src/list-block/list-block.ts
```diff
@@ -104,6 +106,19 @@ export class ListBlockComponent extends CaptionedBlockComponent<
     bindContainerHotkey(this);
 
     this._inlineRangeProvider = getInlineRangeProvider(this);
+
+    effect(() => {
```
Saul-Mirone, line 110: Should we add this to disposable group?

PR #7634, opened 2024-07-18
refactor(database): entering edit mode require two clicks

zzj3720:
close: BS-876


L-Sun (commented):
It would be better to add a test.

zzj3720:
> It would be better to add a test.

There are tests in place now, so after making changes, the tests failed

L-Sun (approved):

PR #7686, opened 2024-07-18
fix: edgeless text can not add max width

Flrande:
Close BS-887


packages/framework/block-std/src/view/element/edgeless-block-element.ts
```diff
@@ -41,11 +49,11 @@ export abstract class EdgelessBlockElement<
   override renderBlock() {
     const { x, y, w, h, zIndex } = this.getRenderingRect();
 
-    this.style.left = `${x}px`;
-    this.style.top = `${y}px`;
-    this.style.width = `${w}px`;
-    this.style.height = `${h}px`;
-    this.style.zIndex = zIndex;
+    if (x) this.style.left = `${x}px`;
+    if (y) this.style.top = `${y}px`;
```
doouding, line 53: x and y might be zero.

doouding (changes requested):

doouding (approved):

Saul-Mirone (approved):

PR #7687, opened 2024-07-19
test: add schema snapshot

doodlewind:
As a starting point of schema checking, this PR leverages vitest for collecting snapshot, making sure newly added props would be properly reviewed via the code owner config.

Suggest to implement in further PRs:

- Field-level block schema definition using zod.
- Formalize the snapshot format, the current format is experimental.


packages/blocks/src/__tests__/schema/__snapshots__/schema.unit.spec.ts.snap
```diff
@@ -0,0 +1,794 @@
...
+    "language": {
+      "type": "string",
+      "default": "Plain Text"
+    },
+    "caption": {
+      "type": "string",
+      "default": ""
+    }
+  }
+}",
```
Saul-Mirone, line 31: The indent seems not correct

doodlewind: It's auto generated by vitest, I don't think the prettier config applies here.

Saul-Mirone:
I think having schema for blocks is definitely a good idea. However, in favor of https://github.com/toeverything/blocksuite/pull/7852. We'll have a `@blocksuite/affine-model` package and we should put the schema with snapshots there.

PR #7705, opened 2024-07-22
refactor(std): add block-std/edgeless scope

doodlewind:
Edgeless related entities should be exported under the `block-std/edgeless` scope.

### DIFF
--- a/packages/framework/block-std/package.json
+++ b/packages/framework/block-std/package.json
@@ -35,7 +35,8 @@
     "@blocksuite/store": "workspace:*"
   },
   "exports": {
-    ".": "./src/index.ts"
+    ".": "./src/index.ts",
+    "./edgeless": "./src/edgeless/index.ts"
   },
   "publishConfig": {
     "access": "public",
@@ -43,8 +44,14 @@
     "types": "dist/index.d.ts",
     "exports": {
       ".": {
+        "types": "./dist/index.d.ts",
         "module": "./dist/index.js",
         "import": "./dist/index.js"
+      },
+      "./edgeless": {
+        "types": "./dist/edgeless/index.d.ts",
+        "module": "./dist/edgeless/index.js",
+        "import": "./dist/edgeless/index.js"
       }
     }
   },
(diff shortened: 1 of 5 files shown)


packages/framework/block-std/package.json
```diff
@@ -45,7 +46,8 @@
       ".": {
         "module": "./dist/index.js",
         "import": "./dist/index.js"
-      }
+      },
+      "./edgeless": "./dist/edgeless/index.js"
```
Saul-Mirone, line 50: You'll also need to add 'types' property here

packages/framework/block-std/package.json
```diff
@@ -45,7 +46,8 @@
       ".": {
         "module": "./dist/index.js",
```
Saul-Mirone, line 47: We should align the configs for each entry points.

Saul-Mirone (changes requested):

packages/framework/block-std/package.json
doodlewind: I checked the docs and seems that if the config between `module` and `import` are the same, we don't have to repeat them. So I suggest making the config simpler.

packages/framework/block-std/package.json
```diff
@@ -35,17 +35,16 @@
...
   "publishConfig": {
     "access": "public",
     "main": "dist/index.js",
     "types": "dist/index.d.ts",
     "exports": {
-      ".": {
-        "module": "./dist/index.js",
-        "import": "./dist/index.js"
-      }
+      ".": "./dist/index.js",
```
Saul-Mirone: You removed the `types` property here, will make users won't be able to find the type declaration when they use this package with typescript `bundler` module resolution.

Saul-Mirone (changes requested):

packages/framework/block-std/package.json
Saul-Mirone: Please check: https://github.com/Milkdown/milkdown/blob/main/packages/milkdown/package.json#L68

doodlewind: I think this was missing in current codebase, just added to keep the config consistent.

Saul-Mirone (approved):

PR #7718, opened 2024-07-24
fix(edgeless): discard group like model from calculating bound

fourdim:
This closes [BS-824](https://linear.app/affine-design/issue/BS-824).

PR #7722, opened 2024-07-24
fix(edgeless): fix group insertion from clipboard

oliseviche:
During insertion of grouped objects the position being calculated on elements which forms the group itself.
The elements includes the group model and by definition of a group it doesn't have its own x-y-w-h values but serialize them by default as '[0,0,0,0]' xywh.
As of it, evaluating the position of insertion always will have zeroed x and y edge in `edgelessElementsBound` call and will always insert the group relative to center of the board and min/max edge of selection rather than center of the last pointer position.

https://github.com/user-attachments/assets/2412edde-b2b4-49af-9db5-aa952fab74b5

To overcome the issue, I suppose just to filter the group elements from insert collection passed to `edgelessElementsBound`

### DIFF
--- a/packages/blocks/src/root-block/edgeless/controllers/clipboard.ts
+++ b/packages/blocks/src/root-block/edgeless/controllers/clipboard.ts
@@ -47,7 +47,10 @@ import {
   type SerializedElement,
   SurfaceGroupLikeModel,
 } from '../../../surface-block/element-model/base.js';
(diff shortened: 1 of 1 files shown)


fourdim:
Dup #7718

fourdim:
Hi. I'm so sorry but we already have #7718 which fixes the same issue and currently merged to the mainline.
I have to close this PR since they are duplicate. Thanks for your contribution again.

PR #7740, opened 2024-07-27
export src/gfx from block-std

### LINKED ISSUE #7739: export GfxBlockElementModel from block-std package
I'm trying to create an edgeless compatible `BlockModel` based on those from the `@blocksuite/blocks` package. I'm running into issues with the compiler failing to resolve type details of [`EdgelessBlockModel`](https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/root-block/edgeless/edgeless-block-model.ts).

The issue is that the models in `@blocksuite/blocks` extend [`EdgelessBlockModel `](https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/root-block/edgeless/edgeless-block-model.ts) which is a re-export of [`GfxBlockElementModel`](https://github.com/toeverything/blocksuite/blob/master/packages/framework/block-std/src/gfx/model.ts) from package `@blocksuite/blocks`. `GfxBlockElementModel` however is not exported from that library. The `@blocksuite/blocks` package is able to build because of the [`workspace` package dependency](https://github.com/toeverything/blocksuite/blob/bbdf12e3d70adc5d4cb846e01accf1862003ca3c/packages/blocks/package.json#L18), but no other external consumers can resolve this type:

I believe the solution is to simply export `src/gfx` from package`@blocksuite/block-std`.

goldsam:
closes #7739

### DIFF
--- a/packages/framework/block-std/src/index.ts
+++ b/packages/framework/block-std/src/index.ts
@@ -1,6 +1,7 @@
 export * from './clipboard/index.js';
 export * from './command/index.js';
 export * from './event/index.js';
+export * from './gfx/index.js';
 export * from './scope/index.js';
 export * from './selection/index.js';
 export * from './service/index.js';


doodlewind (commented):
Thanks for your contribution! However this basic graphics lib in the framework is still evolving, and we tend to use `@blocksuite/block-std/gfx` instead of exposing everything in top level of the `block-std` package, which may create too much noise in its API surface.

PR #7747, opened 2024-07-29
fix(edgeless): layer issue

doouding:
### Fixes
Fixes [BS-899](https://linear.app/affine-design/issue/BS-899/edgeless-图片渲染层级错误)
Fixes [BS-926](https://linear.app/affine-design/issue/BS-926/edgeless-图层显示不正常：edgeless-text-没有参与排序)
Fixes [BS-966](https://linear.app/affine-design/issue/BS-966/group-整体-send-to-back-后仍位于-note-之上)

### Other changes
Added a new mechanism to sync element properties that are stored in yjs to a local map. This way, even if an element is unloaded, you can still access its properties, preventing the issue of getting undefined values in many cases.

This mechanism will migrate to use the same infra of the block in the coming iteration.


doouding:
* **#7747** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7747?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/blocks/src/surface-block/managers/layer-manager.ts
```diff
@@ -528,19 +541,15 @@ export class LayerManager {
       insertToOrderedArray(array, element);
     };
 
+    if (shouldUpdateGroupChildren) {
+      this._init();
```
doodlewind, line 545: This seems to be a `_reset()` method? This runs more than once during the lifecycle of `LayerManager`.

packages/blocks/src/root-block/edgeless/edgeless-root-block.ts
```diff
@@ -200,6 +203,23 @@ export class EdgelessRootBlockComponent extends BlockComponent<
       .catch(console.error);
   }
 
+  private _initLayerUpdateEffect() {
+    const updateLayers = requestThrottledConnectedFrame(() => {
+      const blocks =
+        this.renderRoot?.querySelectorAll<GfxBlockComponent>(
```
doodlewind, line 209: So instead of using `requestUpdate`, here we switch to querying all GFX blocks by DOM and update their index on every layer update. Are there some potential DOM querying performance concerns?

If it's just a temp workaround, it's still acceptable for fixing fatal issues. But in this case, maybe we need to mark it with comments.

Saul-Mirone (approved):
LGTM

packages/blocks/src/image-block/image-block.ts
```diff
@@ -158,6 +159,14 @@ export class ImageBlockComponent extends CaptionedBlockComponent<
     `;
   }
 
+  toZIndex() {
+    return this.rootService?.layer.getZIndex(this.model) ?? 1;
+  }
+
+  updateZIndex() {
```
doodlewind, line 166: It looks suboptimal that we need to hard-code the same default methods for these block types. Inheriting from `GfxBlockComponent` avoids the hardcoding, but seems to be conflict with `CaptionedBlockComponent`.

Do you think we need a better block-element relationship modeling (probably not in this PR) to avoid this @Saul-Mirone?

packages/blocks/src/root-block/edgeless/edgeless-root-block.ts
doouding: A better approach is to implement this logic within the `GfxBlockComponent` so that elements can update themselves. But we can't do this because some block does not extends the `GfxBlockComponent`. 

This is a temp workaround. I'll put a label on this.

packages/blocks/src/image-block/image-block.ts
doouding: This is a temporary workaround. Because we need to render both the page mode image and the edgeless mode image at the same time, the image block cannot extend the GfxBlockComponent. I am adding those methods manually to make it renderable in edgeless mode.

We need to support defining different views for the same type of block depending on its parent type. In this case, an image in the note block should use the "page image block," while an image in the surface block should use the "edgeless image block."

doodlewind (approved):

PR #7753, opened 2024-07-29
fix(edgeless): toolbar nav next button's size should be 20*20

CatsJuice:
previous:

### DIFF
--- a/packages/blocks/src/root-block/edgeless/components/toolbar/edgeless-toolbar.ts
+++ b/packages/blocks/src/root-block/edgeless/components/toolbar/edgeless-toolbar.ts
@@ -240,6 +240,10 @@ export class EdgelessToolbar extends WithDisposable(LitElement) {
     .senior-nav-button {
       padding: 0;
     }
+    .senior-nav-button svg {
+      width: 20px;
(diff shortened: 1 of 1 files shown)


CatsJuice:
* **#7753** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7753?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

 Join @CatsJuice and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

fundon:
It may be caused by this: https://github.com/toeverything/blocksuite/blob/3b62f7a91e89e364342764c7f9d941ccda8ac331/packages/blocks/src/_common/components/button.ts#L103-L104

But it is not reasonable to set svg itself to 32x32.

L-Sun (approved):

L-Sun:
I think the reason is that the icon is changed

CleanShot 2024-08-01 at 12 06 11@2x


CleanShot 2024-08-01 at 12 05 30@2x

PR #7755, opened 2024-07-29
feat: replace selector with query object

Saul-Mirone:


Saul-Mirone:
* **#7755** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7755?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

doouding:
Is there any doc on how to use the new query object?

Saul-Mirone:
> Is there any doc on how to use the new query object?

I plan to write something in next pr.

packages/framework/store/src/store/doc/query.ts
```diff
@@ -0,0 +1,82 @@
...
+import type { BlockModel } from '../../schema/index.js';
+import type { Block } from './block/index.js';
+
+import { BlockViewType } from './doc.js';
+
+export type QueryMatch = {
+  id?: string;
+  flavour?: string;
+  props?: Record<string, unknown>;
+  viewType: BlockViewType;
```
EYHN, line 12: just wonder why we need to define viewtype in query. If there are only include and not_include, the design here will be simpler.

Saul-Mirone: by pass is needed for render a block without render the view of its ancestors

PR #7776, opened 2024-07-31
chore(blocks): improve color picker

fundon:
* no need to use `styleMap` wrapper in `computed`
* improve wheel event


Saul-Mirone (approved):
LGTM

PR #7778, opened 2024-07-31
fix(blocks): show filter-bar when an filter added

lawvs:
When adding a filter, `SingleViewBase` is initialized multiple times, causing it to be unable to set the status through `this.view.filterSetVisible(true);`.

Before

https://github.com/user-attachments/assets/e2fc4a4c-e0c5-4557-9a22-56dfbba6f4d3

https://github.com/user-attachments/assets/dfdc4f51-8278-4911-87bb-39c9af56331b

### DIFF
--- a/packages/blocks/src/database-block/data-view/view-manager/single-view.ts
+++ b/packages/blocks/src/database-block/data-view/view-manager/single-view.ts
@@ -152,7 +152,7 @@ export abstract class SingleViewBase<
   ViewData extends DataViewDataType = DataViewDataType,
 > implements SingleView<ViewData>
 {
-  private _filterVisible$ = signal(false);
+  private _filterVisible$ = signal<boolean | undefined>(undefined);
 
   private searchString = signal('');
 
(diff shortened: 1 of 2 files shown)


zzj3720 (approved):

lawvs:
I removed the `filterVisible` since it is not stored in the model, which will be reset once the `prop:views` is updated (e.g., when updating columns/headers). The filter will always be displayed when there is any filter in this PR. 

Additionally, there are other solutions: saving `filterVisible` in the ydoc.

@zzj3720 

Screenshot 2024-08-18 at 20 21 09

zzj3720:
> I removed the `filterVisible` since it is not stored in the model, which will be reset once the `prop:views` is updated (e.g., when updating columns/headers). The filter will always be displayed when there is any filter now.
> 
> Additionally, there are other solutions: saving `filterVisible` in the ydoc.
> 
> @zzj3720
> 
> Screenshot 2024-08-18 at 20 21 09

This bug is mainly caused by the unexpected recreation of the SingleView instance. It should not be recreated when the ydoc is updated. It's a bit troublesome to write this logic correctly. I think it can be fixed like this for the time being.

zzj3720 (approved):

PR #7801, opened 2024-08-01
feat: add requiredProperties decorator for runtime LitComponent props type check

akumatus:
### What Changed?
- Add `@requiredProperties` decorator for runtime LitComponent property type checking.
- Add some usage examples of this decorator.

### Some principles？
- Do not use the `!` non-null assertion operator unless necessary. 
  - If the property type is `boolean` or `Array`, you can fill in the default value of `false` or `[]`.
- Be sure to add `@requiredProperties` for property end with `!`.
- Do not use the `!` operator after property with `@query` decorator. The query result for a DOM element probably be null.

### What's Next?
If the view layer is refactored by Vue, maybe we don't need this `@requiredProperties`.

### DIFF
--- a/packages/framework/block-std/src/view/element/lit-host.ts
+++ b/packages/framework/block-std/src/view/element/lit-host.ts
@@ -6,7 +6,8 @@ import {
   handleError,
 } from '@blocksuite/global/exceptions';
 import { Slot } from '@blocksuite/global/utils';
-import { type BlockModel, BlockViewType, type Doc } from '@blocksuite/store';
+import { Doc } from '@blocksuite/store';
+import { type BlockModel, BlockViewType } from '@blocksuite/store';
 import { createContext, provide } from '@lit/context';
 import { SignalWatcher } from '@lit-labs/preact-signals';
 import {
@@ -26,7 +27,8 @@ import type { SelectionManager } from '../../selection/index.js';
 import type { BlockSpec, SpecStore } from '../../spec/index.js';
 import type { ViewStore } from '../view-store.js';
 
-import { BlockStdScope } from '../../scope/index.js';
+import { BlockStdScope } from '../../scope/block-std-scope.js';
+import { PropTypes, requiredProperties } from '../decorators/required.js';
 import { RangeManager } from '../utils/range-manager.js';
 import { WithDisposable } from '../utils/with-disposable.js';
 import { ShadowlessElement } from './shadowless-element.js';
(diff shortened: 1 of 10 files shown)


akumatus:
* **#7801** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7801?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

doouding (commented):
I saw you put the required decorator in the `std/gfx/surface/decorators` folder which is not so appropriate, because the decorators under this fold are mostly used in edgeless related scenario.

packages/framework/block-std/src/gfx/surface/decorators/required.ts
```diff
@@ -0,0 +1,58 @@
+import type { Constructor } from '@blocksuite/global/utils';
```
Saul-Mirone, line 1: This should be in the `block-std/src/view/decorators` folder

akumatus:
> I saw you put the required decorator in the `std/gfx/surface/decorators` folder which is not so appropriate, because the decorators under this fold are mostly used in edgeless related scenario.

Thanks for your advice! what a mistake.

packages/framework/block-std/src/gfx/surface/decorators/required.ts
akumatus: Good idea! By the way, there was a lot of use of `!` in LitComponent now, I'm not sure if this `@requiredProperties` refactoring is necessary.

Saul-Mirone (approved):
LGTM

packages/framework/block-std/src/view/element/lit-host.ts
```diff
@@ -27,13 +29,17 @@ import type { BlockSpec, SpecStore } from '../../spec/index.js';
...
+import { PropTypes, requiredProperties } from '../decorators/required.js';
 import { RangeManager } from '../utils/range-manager.js';
 import { WithDisposable } from '../utils/with-disposable.js';
 import { ShadowlessElement } from './shadowless-element.js';
 
 export const docContext = createContext<Doc>('doc');
 export const stdContext = createContext<BlockStdScope>('std');
 
+@requiredProperties({
+  specs: PropTypes.arrayOf(PropTypes.object),
```
Saul-Mirone, line 41: What about doc and std?

doodlewind (approved):

L-Sun:
Is it possible to implement the `requiredProperty` as property decorator like:
```
@required()
@property()
accessor myProperty : number;
```

Due to our use of auto-sort linter, the current approach requires jumping from the end of the class to the beginning of the class, which feels a bit troublesome.

packages/framework/block-std/src/view/element/lit-host.ts
akumatus: Done. Fixed circular dependencies between `view.js` and `scope.js`.

akumatus:
I've had this thought before. But we have to check the Lit property in `connectedCallback` life cycle, otherwise 
 the value is undefined before connected.

L-Sun (approved):

PR #7811, opened 2024-08-01
fix: optimize text renderer and update ai chat block model

donteatfriedrice:
fix: 
[BS-1005](https://linear.app/affine-design/issue/BS-1005/chat-block无法被copy-paste到别的文档中，duplicate全篇文档后，可以center-peek)
[BS-990](https://linear.app/affine-design/issue/BS-990/避免centerpeek中发起的会话，等待ai返回时，页面失去响应)

### DIFF
--- a/packages/presets/src/blocks/ai-chat-block/ai-chat-model.ts
+++ b/packages/presets/src/blocks/ai-chat-block/ai-chat-model.ts
@@ -8,7 +8,9 @@ type AIChatProps = {
   index: string;
   scale: number;
   messages: string; // JSON string of ChatMessage[]
-  sessionId: string;
+  sessionId: string; // forked session id
+  rootWorkspaceId: string; // workspace id of root chat session
+  rootDocId: string; // doc id of root chat session
 };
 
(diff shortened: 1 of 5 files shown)


packages/presets/src/blocks/ai-chat-block/ai-chat-model.ts
```diff
@@ -8,7 +8,9 @@ type AIChatProps = {
   index: string;
   scale: number;
   messages: string; // JSON string of ChatMessage[]
-  sessionId: string;
+  sessionId: string; // forked session id
+  rootWorkspaceId: string; // workspace id of root chat session
+  rootDocId: string; // doc id of root chat session
```
Saul-Mirone, line 13: Why do we need this two property?

donteatfriedrice: The AI Chat block can be copied and pasted in different doc, and should support chatting in different doc. When chatting and querying the forked session histories, when calling the backend api, the root session doc id and workspace id is needed. 

related issue

doodlewind (approved):
LGTM

PR #7856, opened 2024-08-06
fix(edgeless): code block line number break in wrap mode

Flrande:


packages/framework/inline/src/inline-editor.ts
```diff
@@ -237,9 +241,11 @@ export class InlineEditor<
       isEmbed = () => false,
       hooks = {},
       inlineRangeProvider = null,
+      renderVLine = null,
     } = ops;
     this._yText = yText;
     this.isEmbed = isEmbed;
+    this.renderVLine = renderVLine;
```
Saul-Mirone, line 248: should be vLineRenderer

Saul-Mirone (approved):

PR #7869, opened 2024-08-07
fix(edgeless): can not add bookmark in empty edgeless text

Flrande:
Before:

https://github.com/user-attachments/assets/e02b2490-7b14-4047-adb0-07da8615c1c1

### DIFF
--- a/packages/blocks/src/edgeless-text/edgeless-text-service.ts
+++ b/packages/blocks/src/edgeless-text/edgeless-text-service.ts
@@ -54,10 +54,11 @@ export class EdgelessTextBlockService extends BlockService<EdgelessTextBlockMode
             const abortController = new AbortController();
             edgelessText.addEventListener(
               'focusout',
-              () => {
+              e => {
                 if (
                   !paragraph.model.text ||
-                  paragraph.model.text.length === 0
+                  (paragraph.model.text.length === 0 &&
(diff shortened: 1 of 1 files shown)


packages/blocks/src/edgeless-text/edgeless-text-service.ts
```diff
@@ -54,10 +54,11 @@ export class EdgelessTextBlockService extends BlockService<EdgelessTextBlockMode
             const abortController = new AbortController();
             edgelessText.addEventListener(
               'focusout',
-              () => {
+              e => {
                 if (
                   !paragraph.model.text ||
-                  paragraph.model.text.length === 0
+                  (paragraph.model.text.length === 0 &&
+                    e.relatedTarget !== null)
```
L-Sun, line 61: If the bookmark input modal is closed without any content, the empty `edgeless-text` will be retained.

Flrande: Compared to the complete inability to add the bookmark before the fix, the issue you pointed out is still acceptable, so I would like to merge this fix first. Additionally, I feel that the edgeless text is limited in this scenario because it does not restrict the interaction methods of the internal Block.

L-Sun (approved):

Saul-Mirone (approved):

PR #7881, opened 2024-08-08
feat: the view component in block spec could be a function

doouding:
### Changed
- Allow to define the `view.component` in spec as a function which receive a model and return a literal template

### DIFF
--- a/packages/framework/block-std/src/view/element/lit-host.ts
+++ b/packages/framework/block-std/src/view/element/lit-host.ts
@@ -19,7 +19,7 @@ import {
 } from 'lit';
 import { customElement, property, state } from 'lit/decorators.js';
 import { repeat } from 'lit/directives/repeat.js';
-import { html, unsafeStatic } from 'lit/static-html.js';
+import { type StaticValue, html, unsafeStatic } from 'lit/static-html.js';
 
 import type { CommandManager } from '../../command/index.js';
 import type { UIEventDispatcher } from '../../event/index.js';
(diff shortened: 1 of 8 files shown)


doouding:
* **#7881** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7881?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/framework/block-std/src/view/element/lit-host.ts
```diff
@@ -58,7 +58,10 @@ export class EditorHost extends SignalWatcher(
       return html`${nothing}`;
     }
 
-    const tag = view.component;
+    const tag =
+      typeof view.component === 'function'
```
Saul-Mirone, line 62: Will this change after the editor initialized?

For example:
```ts
component: (model) => model.type === 'h1' ? 'affine-h1' : 'affine-paragraph'
```

If I change the model.type from `paragraph` to `h1`, what will happen?

doouding: The correct behavior should rerender the model with `affine-h1`. But I'm not sure whether or not the editor will do so. Let me write a test.

Saul-Mirone: Let's verify this before merging the pr.

doouding: It works. But you should use `model.type$.value` instead.

Saul-Mirone: Cool

Saul-Mirone (approved):
Great work 🥇

Issue #7880, opened 2024-08-08
Circular dependencies


siyou:
There is currently a problem of circular dependency in the code, which can cause some issues. I think it needs to be fixed.

siyou:
I'll commence doing this

fourdim:
Could you please provide more details?

![image](https://github.com/user-attachments/assets/10106c86-2075-47c1-88a6-a6c88206bc48)

siyou:
> Could you please provide more details?
> 
> ![image](https://private-user-images.githubusercontent.com/59462000/356313320-10106c86-2075-47c1-88a6-a6c88206bc48.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjMxNjc3OTcsIm5iZiI6MTcyMzE2NzQ5NywicGF0aCI6Ii81OTQ2MjAwMC8zNTYzMTMzMjAtMTAxMDZjODYtMjA3NS00N2MxLTg4YTYtYTZjODgyMDZiYzQ4LnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MDklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODA5VDAxMzgxN1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWJjMDhjNTU3ODQ0MzI5N2YyZDY5M2UwNGRmMDM3ZDUzMjU4ZTA1NjA1ZGM0NDYzMjAwMjk0ODUyYTBlMThiM2MmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.Um5obTZsLDQ3tr6dfyEJ0SCusYqxGzuEdXJrQHuo-D0)

You can run the script [here](https://github.com/toeverything/blocksuite/pull/7883) and have a look

[...]

PR #7883, opened 2024-08-08
fix: circular dependencies

siyou:


siyou:
#7880

fourdim:
Nice catch! 👍

siyou:
#7889

PR #7912, opened 2024-08-10
feat(blocks): print pdf

EYHN:


EYHN:
* **#7912** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7912?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

 Join @EYHN and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

fourdim (commented):
Why are we having too many important! in CSS?
If you are putting media print after the original CSS, I think it's possible to override it.

packages/blocks/src/_common/print-manager/print-manager.ts
```diff
@@ -0,0 +1,124 @@
+export class PrintManager {
```
Saul-Mirone, line 1: I think we should put this in `@blocksuite/affine-shared/utils`.

Saul-Mirone (approved):
LGTM \w little comments

Saul-Mirone (approved):
LGTM

packages/blocks/src/_common/print-manager/print-manager.ts
EYHN: done

EYHN:
> Why are we having too many important! in CSS?
> If you are putting media print after the original CSS, I think it's possible to override it.

`!important` has been removed as much as possible

PR #7934, opened 2024-08-12
fix: 0.16.1 patch

doouding:
### Fixes

- Edgeless text infinite loop update width @doouding 
- Optimize first viewport element rendering @doouding
- Edgeless content shift on first entering edgeless mode @doouding
- Optimize canvas renderer to avoid looply calling rAf @doouding


Saul-Mirone (approved):
LGTM

PR #7937, opened 2024-08-12
chore(edgeless): adjust selected rect border width

L-Sun:
Close [BS-1096](https://linear.app/affine-design/issue/BS-1096/[frame-and-group-improvement]-frame-和-group-进入选中态-border-改为-2px)

### DIFF
--- a/packages/blocks/src/root-block/edgeless/components/rects/edgeless-selected-rect.ts
+++ b/packages/blocks/src/root-block/edgeless/components/rects/edgeless-selected-rect.ts
@@ -486,7 +486,7 @@ export class EdgelessSelectedRect extends WithDisposable(LitElement) {
       top,
       rotate,
       borderStyle: 'solid',
-      borderWidth: selection.editing ? 2 : 1,
+      borderWidth: 2,
     };
   }, this);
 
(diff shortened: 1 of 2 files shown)


fundon:
https://github.com/toeverything/blocksuite/blob/405ae9b62e8381f4675e2ac55d9259f3c2cb0a88/packages/blocks/src/root-block/edgeless/components/rects/edgeless-selected-rect.ts#L1363
https://github.com/toeverything/blocksuite/blob/405ae9b62e8381f4675e2ac55d9259f3c2cb0a88/packages/blocks/src/root-block/edgeless/components/rects/edgeless-selected-rect.ts#L1299

It may be necessary to determine whether to select single selection or multiple selections.

fundon (commented):
https://github.com/toeverything/blocksuite/pull/7937#issuecomment-2283647749

L-Sun:
Good catch. I've updated the code for the first case. 

For the second case (see the border of inner element in the figure), this PR is make the border width same as resize handler, so there is no need to change the code for that case.

CleanShot 2024-08-12 at 19 49 35@2x

L-Sun:
Need to update test

fundon (approved):

PR #7939, opened 2024-08-12
refactor: move commands to block spec declarations

akumatus:
### What changed?
- Move commands to block spec declarations
- Separate global declares

### What's next?
- Support `InlineSpec` customization
- Separate `@blocksuite/affine-shared/commands`
- Remove `slots` in RootService

### DIFF
--- a/packages/blocks/src/root-block/root-service.ts
+++ b/packages/blocks/src/root-block/root-service.ts
@@ -1,18 +1,14 @@
+import type { PeekViewService } from '@blocksuite/affine-components/peek';
+import type { RefNodeSlots } from '@blocksuite/affine-components/rich-text';
 import type { RootBlockModel } from '@blocksuite/affine-model';
 import type { BlockComponent } from '@blocksuite/block-std';
 import type { BlockModel } from '@blocksuite/store';
 
 import {
-  type PeekViewService,
-  getSelectedPeekableBlocksCommand,
-  peekSelectedBlockCommand,
-} from '@blocksuite/affine-components/peek';
-import {
   type EmbedCardStyle,
   type NoteBlockModel,
   NoteDisplayMode,
 } from '@blocksuite/affine-model';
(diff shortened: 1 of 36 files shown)


akumatus:
* **#7939** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7939?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

Saul-Mirone (approved):

packages/blocks/src/root-block/root-service.ts
```diff
@@ -411,4 +401,22 @@ export class RootService extends BlockService<RootBlockModel> {
     assertExists(viewportElement);
     return viewportElement;
   }
+
+  abstract slots: {
```
Saul-Mirone, line 405: We have a `RefNodeSlots` type in `@blocksuite/affine-components/rich-text`

packages/presets/src/index.ts
```diff
@@ -1,3 +1,5 @@
+import '@blocksuite/affine-shared/commands';
```
Saul-Mirone, line 1: What's this used for?

akumatus: Apply global type augmentations of commands like:
```
declare global {
  namespace BlockSuite {
    interface Commands {
      copySelectedModels: typeof copySelectedModelsCommand;
    }
  }
}
```
Otherwise there will be some TypeScript Error like `Property 'copySelectedModels' does not exist on type 'Chain<InitCommandCtx & CommandKeyToData<never>>'` in package build stage.

L-Sun (approved):

packages/blocks/src/root-block/root-service.ts
akumatus: Got it. Thanks for your comments!

Issue #7949, opened 2024-08-13
feature request: user can create a paragraph block through click between two not-paragraph-block


raintoway:
show as follow:
![1](https://github.com/user-attachments/assets/802cff8a-1670-4857-993f-28c21cc203c5)

PR #7950, opened 2024-08-13
feat(blocks): add a paragraph block through clicking between two not-paragraph block

raintoway:
this [issue](https://github.com/toeverything/blocksuite/issues/7949),

feature as follow:


Saul-Mirone:
Hi @raintoway. Thanks for your contribution and we think this change makes sense and improves the user experience.

However, before merging this PR. I think there's something needs to improve:

1. We also have blocks named `affine-embed-xxx`. Like `Github`, `Figma`, etc... They're very similar to the bookmark block.  You can find the code here: [embed-block-element](https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/_common/embed-block-helper/embed-block-element.ts#L31). I think you may also want to set the flags for them.
2. It would be great if you can write some E2E tests in `tests/`.

packages/blocks/src/_common/components/block-zero-width.ts
```diff
@@ -0,0 +1,49 @@
...
+      z-index: 1;
+    }
+  `;
+
+  override connectedCallback(): void {
+    super.connectedCallback();
+  }
+
+  handleClick() {
+    if (this.block.doc.readonly) return;
```
Saul-Mirone, line 25: A `stopPropagation` is needed.

hwangdev97 (approved):

PR #7958, opened 2024-08-13
fix(blocks): add link in code block

Flrande:


packages/blocks/src/code-block/highlight/affine-code-unit.ts
```diff
@@ -40,28 +40,45 @@ export class AffineCodeUnit extends ShadowlessElement {
...
+        return html`<v-text
+          .str=${token.content}
+          style=${styleMap({
+            color: token.color,
+          })}
+        ></v-text>`;
+      });
 
-    return html`<span>${vTexts}</span>`;
+      return html`<span>${vTexts}</span>`;
```
Saul-Mirone, line 80: We should add a test case for this conner case.

Saul-Mirone (approved):
LGTM 💯

PR #7978, opened 2024-08-14
chore(playground): cleanup legacy examples

doodlewind:
In the playground package:

- The `provider` example is removed, it's already replaced by the `/examples/vanilla-indexeddb` example.
- The `store` example is migrated to the new `/examples/custom-store` entry, so that more examples about store customization can be added in the future.

All remaining `/playground/examples` entries are now only for internal testing.


Saul-Mirone:
Your pr destroy the ci.

doodlewind:
Then it shouldn't have passed all the tests 🙂‍↕️

PR #7995, opened 2024-08-16
fix(presets): remove editor runtime type check from outline viewer

L-Sun:
This PR remove runtime type check of `AffineEditorContainer`.

Since `AffineEditorContainer` of AFFiNE side is not same as blocksuite site, the following code will throw error in AFFiNE side
```ts
@requiredProperties({
  editor: PropTypes.instanceOf(AffineEditorContainer),
})
```

### DIFF
--- a/packages/presets/src/fragments/outline/outline-viewer.ts
+++ b/packages/presets/src/fragments/outline/outline-viewer.ts
@@ -10,7 +10,8 @@ import { customElement, property, query, state } from 'lit/decorators.js';
 import { classMap } from 'lit/directives/class-map.js';
 import { repeat } from 'lit/directives/repeat.js';
 
-import { AffineEditorContainer } from '../../editors/editor-container.js';
+import type { AffineEditorContainer } from '../../editors/editor-container.js';
+
 import { TocIcon } from '../_common/icons.js';
 import { getHeadingBlocksFromDoc } from './utils/query.js';
 import {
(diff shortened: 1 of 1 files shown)


L-Sun:
* **#7995** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/7995?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

 Join @L-Sun and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

packages/presets/src/fragments/outline/outline-viewer.ts
```diff
@@ -20,9 +17,6 @@ import {
 
 export const AFFINE_OUTLINE_VIEWER = 'affine-outline-viewer';
 
-@requiredProperties({
-  editor: PropTypes.instanceOf(AffineEditorContainer),
-})
 @customElement(AFFINE_OUTLINE_VIEWER)
```
Saul-Mirone, line 20: Instead of remove this, can we replace with some other check?

L-Sun: After discussing with @akumatus , we can use `PropTypes.object` to cover null and undefined check which is the most common case. 

```ts
@requiredProperties({
  editor: PropTypes.object,
})
```

Saul-Mirone (approved):
LGTM

PR #7997, opened 2024-08-16
fix(selection/text): it's also collapsed while cursor between two par…

ydyoto:
…agraph


Saul-Mirone:
@Flrande PTAL

Flrande:
I don't think this case is collapsed

Flrande:
Meets expectations and no updates, just close it

PR #8019, opened 2024-08-19
fix: edgeless pressShiftKeyUpdated slot never trigger

donteatfriedrice:
[BS-1226](https://linear.app/affine-design/issue/BS-1226/edgeless-下无法触发-shift-keydown-事件)

edgeless pressShiftKeyUpdated slot never triger, which will cause some functions that depend on this slot to not work properly, such as holding down shift to scale certain elements proportionally, like note and embed synced doc.


Saul-Mirone (approved):
Nice catch 💯

doouding (approved):

PR #8020, opened 2024-08-19
refactor(blocks): code-block get supported langs in config

Flrande:
Close [BS-1218](https://linear.app/affine-design/issue/BS-1218/code-block在affine里炸了)


Saul-Mirone (approved):
LGTM

PR #8024, opened 2024-08-21
feat(blocks): inline latex

Flrande:


packages/affine/block-list/src/list-service.ts
```diff
@@ -38,6 +38,7 @@ export class ListBlockService<
     const inlineSpecs = getAffineInlineSpecsWithReference(
       this.referenceNodeConfig
     );
+    // @ts-expect-error
```
Saul-Mirone, line 41: Why?

Saul-Mirone (approved):

packages/blocks/src/data-view-block/database-service.ts
```diff
@@ -27,6 +27,7 @@ export class DataViewBlockService<
     const inlineSpecs = getAffineInlineSpecsWithReference(
       this.referenceNodeConfig
     );
+    // @ts-expect-error
```
Saul-Mirone, line 30: This can also be removed like in `paragraph-service`.

packages/blocks/src/database-block/database-service.ts
```diff
@@ -56,6 +56,7 @@ export class DatabaseBlockService<
     const inlineSpecs = getAffineInlineSpecsWithReference(
       this.referenceNodeConfig
     );
+    // @ts-expect-error
```
Saul-Mirone, line 59: Same here

PR #8027, opened 2024-08-21
refactor: add affine palette colors and styles enum

akumatus:
### What changed?
- Add `enum` type for palette colors and styles
- Move file to `affine-model` consts folder
- Remove duplicate defined TypeScript types
- Update References

### Why make this change?
Enums are easy to read and use.

```typescript
// before
mode === 'dark' ? LINE_COLORS[10] : LINE_COLORS[8];
// after
mode === 'dark' ? LineColor.White : LineColor.Black;

### CODE REVIEW

**Comment by akumatus** [comment]
* **#8027** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/8027?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

File: packages/affine/model/src/consts/line.ts [hunk 1]
```diff
@@ -0,0 +1,27 @@
...
+  Green = '--affine-palette-line-green',
+  Grey = '--affine-palette-line-grey',
+  Magenta = '--affine-palette-line-magenta',
+  Orange = '--affine-palette-line-orange',
+  Purple = '--affine-palette-line-purple',
+  Red = '--affine-palette-line-red',
+  Teal = '--affine-palette-line-teal',
+  White = '--affine-palette-line-white',
+  Yellow = '--affine-palette-line-yellow',
+}
```
fundon, line 17: Need to ensure order in color palette.

Screenshot 2024-08-21 at 21 53 09

akumatus: Thanks for your comments! I have changed the code and ensured the order of color list.

packages/affine/model/src/consts/line.ts
```diff
@@ -0,0 +1,37 @@
+import { createZodUnion } from '../utils/zod.js';
+
+export enum LineColor {
```
Saul-Mirone, line 3: Should we use `@toeverything/theme` and `cssVarV2` as mentioned in the weekly meeting?

akumatus: I'm afraid not. This `--affine-xxx-xxx` is saved into data model.

akumatus: I think it's better to save `yellow` rather than `--affine-palette-line-yellow` in datastrcuture.

fundon: He suggested storing real values. /cc @EYHN

fundon (approved):

fundon:
LGTM

packages/affine/model/src/consts/line.ts
akumatus: @fundon Maybe we can use a color system like `@toeverything/theme` and turn `--affine-palette-shape-yellow`, `--affine-palette-line-yellow` and `--affine-palette-shape-yellow`in to the same `theme-yellow` variable. But legacy data is a barrier to this refactoring. @Saul-Mirone

fundon: It makes sense to name them separately, so that we can adjust the shape's border and background, as well as the note's background, separately.

If we migrate to real color values in the future, we can avoid the problems associated with theme upgrades.

PR #8041, opened 2024-08-25
fix: multiple editor event dispatcher active status

donteatfriedrice:
1. Introduction of a static variable for active Event Dispatcher management
- Added a static variable _activeDispatcher to identify the currently active event dispatcher.
- Ensures that only one event dispatcher is active at any given time.

2. More accurate determination of the current active element's state
 - _isActiveElementOutsideHost and _isEditableElementActive method for better recognition of editable elements outside of current editor host.
- Enhanced conditions for setting the active state in the _setActive method, considering more edge cases

3. Remove deprecated slots.

Fix: 
[BS-713](https://linear.app/affine-design/issue/BS-713/cursor-游离在-edgeless-面板会跟-ai-chat-输入冲突)
[BS-1261](https://linear.app/affine-design/issue/BS-1261/cmd-k-面板搜索时，即使光标不在白板，也会触发白板快捷键)
[BS-1262](https://linear.app/affine-design/issue/BS-1262/全选-block-时，光标移动到编辑区域外，再次-cmda，会选中编辑区域外的文字)

### DIFF
--- a/packages/framework/block-std/src/event/dispatcher.ts
+++ b/packages/framework/block-std/src/event/dispatcher.ts
@@ -1,5 +1,5 @@
 import { BlockSuiteError, ErrorCode } from '@blocksuite/global/exceptions';
-import { DisposableGroup, Slot } from '@blocksuite/global/utils';
+import { DisposableGroup } from '@blocksuite/global/utils';
 
 import type { BlockComponent } from '../view/index.js';
 
@@ -71,6 +71,8 @@ export type EventHandlerRunner = {
 export class UIEventDispatcher {
   private _active = false;
 
+  private static _activeDispatcher: UIEventDispatcher | null = null;
+
   private _clipboardControl: ClipboardControl;
 
   private _handlersMap = Object.fromEntries(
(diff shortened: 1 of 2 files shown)


donteatfriedrice:
* **#8041** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/8041?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

Saul-Mirone (approved):
LGTM 🥇 

Can you please also add some tests?

packages/framework/block-std/src/event/dispatcher.ts
```diff
@@ -227,6 +240,43 @@ export class UIEventDispatcher {
...
+    if (!element) return false;
+    return (
+      element instanceof HTMLInputElement ||
+      element instanceof HTMLTextAreaElement ||
+      (element as HTMLElement).isContentEditable
+    );
+  }
+
+  private _setActive(active: boolean) {
+    if (this.host.doc.readonly) return;
```
L-Sun, line 262: This line will make edgeless canvas can not be moved by touch in mobile readonly mode. So what are the considerations behind this line of change?

donteatfriedrice: The consideration here is that in addition to the main editable editor, a web page may also have some read-only preview editors embedded in the sidebar or main editor, and we don't want these read-only editors to be in the active state. However, if this will cause the mobile terminal to be unable to move the viewport when it is read-only, then some modifications may need to be made.

donteatfriedrice: Since some checks have been done in certain scenarios：

```
  private _isEditableElementActive(element: Element | null): boolean {
    if (!element) return false;
    return (
      element instanceof HTMLInputElement ||
      element instanceof HTMLTextAreaElement ||
      (element as HTMLElement).isContentEditable
    );
  }
```

And different scenarios have been tested, this line of code can be removed.

PR #8076, opened 2024-08-28
feat(std): ioc and DocModeService

Saul-Mirone:
Add di container for std

### DIFF
--- a/packages/blocks/src/image-block/image-resize-manager.ts
+++ b/packages/blocks/src/image-block/image-resize-manager.ts
@@ -1,5 +1,6 @@
 import type { BlockComponent, PointerEventState } from '@blocksuite/block-std';
 
+import { getMode } from '@blocksuite/affine-shared/services';
 import {
   getClosestBlockComponentByElement,
   getModelByElement,
@@ -75,10 +76,7 @@ export class ImageResizeManager {
     ) as BlockComponent;
 
     const rootComponent = getClosestRootBlockComponent(this._activeComponent);
-    if (
-      rootComponent &&
-      rootComponent.service.docModeService.getMode() === 'edgeless'
-    ) {
+    if (rootComponent && getMode(rootComponent.service.std)) {
       this._zoom = (
         rootComponent as EdgelessRootBlockComponent
       ).service.viewport.zoom;
--- a/packages/blocks/src/image-block/image-resize-manager.ts
+++ b/packages/blocks/src/image-block/image-resize-manager.ts
@@ -1,5 +1,6 @@
 import type { BlockComponent, PointerEventState } from '@blocksuite/block-std';
 
+import { DocMode } from '@blocksuite/affine-model';
 import { getMode } from '@blocksuite/affine-shared/services';
 import {
   getClosestBlockComponentByElement,
@@ -76,7 +77,10 @@ export class ImageResizeManager {
     ) as BlockComponent;
 
     const rootComponent = getClosestRootBlockComponent(this._activeComponent);
-    if (rootComponent && getMode(rootComponent.service.std)) {
+    if (
+      rootComponent &&
+      getMode(rootComponent.service.std) === DocMode.Edgeless
+    ) {
       this._zoom = (
         rootComponent as EdgelessRootBlockComponent
       ).service.viewport.zoom;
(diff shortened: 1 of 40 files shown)


Saul-Mirone:
* **#8076** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/8076?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/framework/block-std/src/scope/block-std-scope.ts
```diff
@@ -47,6 +51,8 @@ export class BlockStdScope {
     this.spec = new SpecStore(this);
     this.view = new ViewStore(this);
     this.clipboard = new Clipboard(this);
+    this.container = new Container();
+    this.container.addFactory(BlockStdScope, () => this);
```
doodlewind, line 55: Why using the internal `addFactory` instead of `addImpl` here? Not quite sure about what's the proper concept for the DI users to call.

Saul-Mirone: To keep the instance is the same.

doodlewind (commented):
Here's my basic understanding to this DI design when compared to inversify.

In inversify:

``` ts
import { inject, injectable } from "inversify";

@injectable()
class Engine {
    start() {
        return "vroom!";
    }
}

@injectable()
class Car {
    constructor(@inject("PowerSource") private engine: Engine) {}
    drive() { return this.engine.start(); }
}
```

In BlockSuite DI:

``` ts
const container = new Container();

interface PowerSource {
    start(): string;
}

const PowerSource = createIdentifier<PowerSource>('PowerSource');

class Engine implements PowerSource {
    start() {
        return "vroom!";
    }
}

class Car {
    constructor(private powerSource: PowerSource) {}
    drive() { return this.powerSource.start(); }
}

container.addImpl(PowerSource, Engine);
container.add(Car, [PowerSource]);

const provider = container.provider();
const car = provider.get(Car);
car.drive();
```

Subjective feedback on naming based on the understanding:

* `rootService.std.container` -> `rootService.std.di`? The concept of _container_ is well defined under the context of IoC, but a _container_ in the editor framework may be ambiguous with UI stuffs.
* `container.add` -> `container.addService`? This could distinguish better with `container.addImpl`.

packages/blocks/src/root-block/root-service.ts
```diff
@@ -140,6 +135,8 @@ export abstract class RootService extends BlockService<RootBlockModel> {
   }
 
   override mounted() {
+    DocModeService.add(this.std);
```
doouding, line 138: So, should the dependencies be injected during the root service mount?

packages/blocks/src/image-block/image-resize-manager.ts
```diff
@@ -77,7 +79,7 @@ export class ImageResizeManager {
     const rootComponent = getClosestRootBlockComponent(this._activeComponent);
     if (
       rootComponent &&
-      rootComponent.service.docModeService.getMode() === 'edgeless'
+      getMode(rootComponent.service.std) === DocMode.Edgeless
```
akumatus, line 82: I prefer to have APIs like `modeServcie.getMode()` rather than `getMode(rootComponent.service.std)`. The second one looks like a utility function.

packages/blocks/src/root-block/root-service.ts
```diff
@@ -140,6 +135,8 @@ export abstract class RootService extends BlockService<RootBlockModel> {
...
+
```
akumatus, line 139: May be we need to refactor the initialize logic of `std` first. I want to remove the dependence on component mounted.

packages/playground/apps/default/utils/editor.ts
```diff
@@ -121,8 +129,10 @@ export async function mountDefaultDocEditor(collection: DocCollection) {
               return Promise.resolve();
             },
           };
-          pageRootService.docModeService = mockDocModeService(
-            pageRootService.doc.id
+          pageRootService.std.container.override(
+            DocModeFeature,
+            MockDocModeService,
+            [BlockStdScope]
```
akumatus, line 135: Why does override need to explicitly declare the `[BlockStdScope]` it depends on?

akumatus:
I still think that ioc service is a data layer thing and should not be coupled to views. The rich-text editor is a view component, the ioc service should not depend on the editor being successfully mounted.

Therefore, I think it is reasonable for ioc to have the following scopes:
- Singleton Scope
- Doc Scope
- Transient Scope
Maybe we do not need a `Editor Scope` in ioc, cause editor is a view component.

If we want to have some thing which is a singleton in editor scope, curent `BlockSpecService` is enough to use. You can encapsulate a editor scope `getMode` in BlockSpecService from a singleton scope `DocModeService` in ioc container.

@Saul-Mirone @doodlewind

PR #8083, opened 2024-08-29
feat: ioc and DocModeService

Saul-Mirone:


Saul-Mirone:
* **#8083** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/8083?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/blocks/src/root-block/edgeless/edgeless-root-spec.ts
```diff
@@ -96,6 +100,9 @@ export const EdgelessRootBlockSpec: EdgelessRootBlockSpecType = {
     },
   },
   commands,
+  setup: (_slots, _disposableGroup, di) => {
+    di.addImpl(DocModeFeature, DocModeService, [BlockStdScope]);
```
akumatus, line 104: I think it's not a good idea to do this `addImpl` side effect in EdgelessRootBlockSpec `setup` hook.

The setup hook is asynchronous and coupling the DocModeService with the EdgelessRootBlockSpec.

Saul-Mirone: After discuss with @EYHN.  Setup is the only reasonable place for now, to setup the container. DI need two stages:

1. Config. In this stage we put everything into the container.
2. Runtime. In this stage, the container won't be modified anymore and we'll use `provider` to get the value in the container.

Setup function is not asynchronous. Setup will be like the entry point of blocks so we can do:

```ts
  di.addBlockRenderer
  di.addShortKey
  di.addLifecycleListener
  di.addConfig
```

In the future. It will be very similar to vsc extension.

EYHN: and we need a callback after config step, let affine
to override service

EYHN (approved):

packages/blocks/src/root-block/edgeless/edgeless-root-spec.ts
akumatus: Adding this boilerplate code for every dependent service is difficult to maintain, and it is easy to forget to add `di.impl` here.

Saul-Mirone: I defined the extension API in next pr. Basically it can be used like:
```ts
const spec = {
  extensions: [
    blockFlavourExtension,
    blockServiceExtension,
    blockConfigExtension
  ]
}
```

PR #8084, opened 2024-08-29
fix(edgeless): empty group names should not break copy-paste actions

oliseviche:
User can edit group name removing the title totally, leaving the input field blank.
Copy-Paste action would deserialize the field into the empty string, not the Y.Text instance and the call of toJson over string type variable would result in "call on undefined" error.
Additionally, the initial fromYProps implementation should consider transforming of the zero length string to the Y.Text instance, because the inline editor expects it to function properly.

image

https://github.com/user-attachments/assets/d1f84b47-f68f-426b-880e-b5500542c04b


doodlewind (approved):
Thanks for the fix!

PR #8096, opened 2024-08-29
feat: each editor instance has its own independent local selection

donteatfriedrice:
[BS-1079](https://linear.app/affine-design/issue/BS-1079/同一个页面中的不同-editor-需要区分选区)

1. selection manager adds a unique id identifier. Due to the need to identify selection operations on the same document, a composite ID is required.

   ```
    selectionManager._id = `${this.std.blockCollection.id}:${nanoid()}`;
   ```

2. The logic for distinguishing between local selection and remote selection needs to be changed. Only the selection in selectionV2 that corresponds to the current editor selection manager ID is considered local selection, while selections from other clients are remote selections. Selections managed by different selection managers of the same client, such as in a split view, are neither local nor remote. However, if multiple windows exist with selections, they will be synchronized together to other windows and treated as remote selections for those windows.

    The awareness store also needs to be modified, passing the selectionManagerId when calling getLocalSelection and setLocalSelection.

    ```
    getLocalSelection(
        selectionManagerId: string
     ): ReadonlyArray<Record<string, unknown>> {
        return (
      (this.awareness.getLocalState()?.selectionV2 ?? {})[selectionManagerId] ??
          []
        );
      }

      setLocalSelection(selectionManagerId: string, selection: UserSelection) {
        const oldSelection = this.awareness.getLocalState()?.selectionV2 ?? {};
        this.awareness.setLocalStateField('selectionV2', {
          ...oldSelection,
          [selectionManagerId]: selection,
        });
      }
    ```

3. In some places where the document selectionChange event is directly listened to, checks will be performed, and the selection of editors whose event dispatcher status is not active will not be updated.

4.  Collaboration between selection managers of the same version will not cause any issues. However, when collaborating across different versions, the new version will function normally, but the old version may not be able to properly display the selection from the new version.

    Previously, selectionV2 used blockCollectionId as the key.
Now it uses ${this.std.blockCollection.id}:${nanoid()}.

    Old version selection managers don't have an ID. When setting selection, the key is blockCollectionId.

    New version selection managers have an ID. Key is blockCollectionId:nanoId. Attempts to distinguish user instances by selectionManagerId.

    When old and new clients collaborate on the same blockCollection,
the awareness.state.selectionV2 data might look like this:

    ```
    {
      blockCollectionId: selectionJson,
      blockCollectionId:selection1: selectionJson,
      blockCollectionId:selection2: selectionJson,
    }
    ```

    For old clients, they will only get the selection with the key blockCollectionId under the same client in state selectionV2. Selections with different clientIds but the same blockCollectionId are treated as remoteSelections. In this case, since the new version's key is blockCollectionId:selection1, the old version cannot obtain the remote selection from the new version. 

    For the new version, using selectionManagerId as the key, selections with keys matching selectionManagerId under the same client in state selectionV2 are considered local. Selections from different clients but with keys starting with the same blockCollectionId, are considered remote selections. In this case, new version clients can normally obtain selections from old versions.

### DIFF
--- a/packages/framework/block-std/src/range/range-binding.ts
+++ b/packages/framework/block-std/src/range/range-binding.ts
@@ -171,17 +171,14 @@ export class RangeBinding {
       return;
     }
     const range = selection.rangeCount > 0 ? selection.getRangeAt(0) : null;
-    const isRangeReversed =
-      !!selection.anchorNode &&
-      !!selection.focusNode &&
-      (selection.anchorNode === selection.focusNode
-        ? selection.anchorOffset > selection.focusOffset
-        : selection.anchorNode.compareDocumentPosition(selection.focusNode) ===
-          Node.DOCUMENT_POSITION_PRECEDING);
 
     if (!range) {
       this._prevTextSelection = null;
       this.selectionManager.clear(['text']);
+      return;
+    }
+
+    if (!this.host.contains(range.commonAncestorContainer)) {
       return;
     }
 
@@ -214,6 +211,13 @@ export class RangeBinding {
     );
     if (inlineEditor?.isComposing) return;
 
+    const isRangeReversed =
+      !!selection.anchorNode &&
+      !!selection.focusNode &&
+      (selection.anchorNode === selection.focusNode
+        ? selection.anchorOffset > selection.focusOffset
+        : selection.anchorNode.compareDocumentPosition(selection.focusNode) ===
+          Node.DOCUMENT_POSITION_PRECEDING);
     const textSelection = this.rangeManager?.rangeToTextSelection(
       range,
       isRangeReversed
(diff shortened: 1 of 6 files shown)


donteatfriedrice:
* **#8096** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/8096?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

Flrande (approved):

packages/framework/block-std/src/range/range-binding.ts
```diff
@@ -245,20 +249,21 @@ export class RangeBinding {
...
-        const eq =
-          text && this._prevTextSelection && path
-            ? text.equals(this._prevTextSelection.selection) &&
-              path.join('') === this._prevTextSelection.path.join('')
-            : false;
+        if (this.host.event.active) {
+          const eq =
+            text && this._prevTextSelection && path
+              ? text.equals(this._prevTextSelection.selection) &&
+                path.join('') === this._prevTextSelection.path.join('')
```
Saul-Mirone, line 262: please also remove path here, use block id instead

Saul-Mirone (changes requested):

packages/framework/block-std/src/range/range-binding.ts
donteatfriedrice: The pr has been automatically merged by graphite, I will open another pr to remove path.

PR #8103, opened 2024-08-30
feat(blocks): block level latex

Flrande:


Flrande:
https://github.com/toeverything/blocksuite/pull/8177

PR #8110, opened 2024-09-02
chore: add editor mode related service on doc mode service

donteatfriedrice:
Fix: [BS-1292](https://linear.app/affine-design/issue/BS-1292/无法通过-docmodeservicesetmode-切换-mode)

rename:

before:

```
  setMode: (mode: DocMode, docId?: string) => void;
  getMode: (docId?: string) => DocMode;
  toggleMode: (docId?: string) => DocMode;
  onModeChange: (
    handler: (mode: DocMode) => void,
    docId?: string
  ) => Disposable;
```

after:

```
  /**
   * Set the primary mode of the doc.
   * This would not affect the current editor mode.
   * If you want to switch the editor mode, use `setEditorMode` instead.
   * @param mode - The mode to set.
   * @param docId - The id of the doc.
   */
  setPrimaryMode: (mode: DocMode, docId: string) => void;
  /**
   * Get the primary mode of the doc.
   * Normally, it would be used to query the mode of other doc.
   * @param docId - The id of the doc.
   * @returns The primary mode of the document.
   */
  getPrimaryMode: (docId: string) => DocMode;
  /**
   * Toggle the primary mode of the doc.
   * @param docId - The id of the doc.
   * @returns The new primary mode of the doc.
   */
  togglePrimaryMode: (docId: string) => DocMode;
  /**
   * Subscribe to changes in the primary mode of the doc.
   * For example:
   * Embed-linked-doc-block will subscribe to the primary mode of the linked doc,
   * and will display different UI according to the primary mode of the linked doc.
   * @param handler - The handler to call when the primary mode of certain doc changes.
   * @param docId - The id of the doc.
   * @returns A disposable to stop the subscription.
   */
  onPrimaryModeChange: (
    handler: (mode: DocMode) => void,
    docId: string
  ) => Disposable;
```

add:

```
  /**
   * Set the editor mode. Normally, it would be used to set the mode of the current editor.
   * When patch or override the doc mode service, can pass a callback to set the editor mode.
   * @param mode - The mode to set.
   */
  setEditorMode: (mode: DocMode) => void;
  /**
   * Get current editor mode.
   * @returns The editor mode.
   */
  getEditorMode: () => DocMode | null;
```

mock docModeService like this

```
export function mockDocModeService(
  getEditorModeCallback: () => DocMode,
  setEditorModeCallback: (mode: DocMode) => void
) {
  const docModeService: DocModeProvider = {
    getPrimaryMode: (docId: string) => {
      try {
        const modeMap = getModeFromStorage();
        return modeMap.get(docId) ?? DEFAULT_MODE;
      } catch (_e) {
        return DEFAULT_MODE;
      }
    },
    onPrimaryModeChange: (handler: (mode: DocMode) => void, docId: string) => {
      if (!slotMap.get(docId)) {
        slotMap.set(docId, new Slot());
      }
      return slotMap.get(docId)!.on(handler);
    },
    getEditorMode: () => {
      return getEditorModeCallback();
    },
    setEditorMode: (mode: DocMode) => {
      setEditorModeCallback(mode);
    },
    setPrimaryMode: (mode: DocMode, docId: string) => {
      const modeMap = getModeFromStorage();
      modeMap.set(docId, mode);
      saveModeToStorage(modeMap);
      slotMap.get(docId)?.emit(mode);
    },
    togglePrimaryMode: (docId: string) => {
      const mode =
        docModeService.getPrimaryMode(docId) === 'page' ? 'edgeless' : 'page';
      docModeService.setPrimaryMode(mode, docId);
      return mode;
    },
  };
  return docModeService;
}
```

And override docModeService implementation

```
  function patchPageRootSpec(spec: ExtensionType[]) {
    class PatchPageServiceWatcher extends BlockServiceWatcher {
      static override readonly flavour = 'affine:page';

    ...

    const setEditorModeCallBack = editor.switchEditor.bind(editor);
    const getEditorModeCallback = () => editor.mode;
    const newSpec: typeof spec = [
      ...spec,
      {
        setup: di => {
          di.override(DocModeProvider, () =>
            mockDocModeService(getEditorModeCallback, setEditorModeCallBack)
          );
        },
      },
    ];

    return newSpec;
  }
```

### DIFF
--- a/packages/affine/shared/src/services/doc-mode-service.ts
+++ b/packages/affine/shared/src/services/doc-mode-service.ts
@@ -4,18 +4,56 @@ import type { Container } from '@blocksuite/global/di';
 
 import { Extension, StdIdentifier } from '@blocksuite/block-std';
 import { createIdentifier } from '@blocksuite/global/di';
-import { type Disposable, Slot } from '@blocksuite/global/utils';
+import { type Disposable, noop, Slot } from '@blocksuite/global/utils';
 
 const DEFAULT_MODE: DocMode = 'page';
 
 export interface DocModeProvider {
-  setMode: (mode: DocMode, docId?: string) => void;
-  getMode: (docId?: string) => DocMode;
-  toggleMode: (docId?: string) => DocMode;
-  onModeChange: (
+  /**
+   * Set the primary mode of the doc.
+   * This would not affect the current editor mode.
+   * If you want to switch the editor mode, use `setEditorMode` instead.
+   * @param mode - The mode to set.
+   * @param docId - The id of the doc.
+   */
+  setPrimaryMode: (mode: DocMode, docId: string) => void;
+  /**
+   * Get the primary mode of the doc.
+   * Normally, it would be used to query the mode of other doc.
+   * @param docId - The id of the doc.
+   * @returns The primary mode of the document.
+   */
+  getPrimaryMode: (docId: string) => DocMode;
+  /**
+   * Toggle the primary mode of the doc.
+   * @param docId - The id of the doc.
+   * @returns The new primary mode of the doc.
+   */
+  togglePrimaryMode: (docId: string) => DocMode;
+  /**
+   * Subscribe to changes in the primary mode of the doc.
+   * For example:
+   * Embed-linked-doc-block will subscribe to the primary mode of the linked doc,
+   * and will display different UI according to the primary mode of the linked doc.
+   * @param handler - The handler to call when the primary mode of certain doc changes.
+   * @param docId - The id of the doc.
+   * @returns A disposable to stop the subscription.
+   */
+  onPrimaryModeChange: (
     handler: (mode: DocMode) => void,
-    docId?: string
+    docId: string
   ) => Disposable;
+  /**
+   * Set the editor mode. Normally, it would be used to set the mode of the current editor.
+   * When patch or override the doc mode service, can pass a callback to set the editor mode.
+   * @param mode - The mode to set.
+   */
+  setEditorMode: (mode: DocMode) => void;
+  /**
+   * Get current editor mode.
+   * @returns The editor mode.
+   */
+  getEditorMode: () => DocMode | null;
 }
 
 export const DocModeProvider = createIdentifier<DocModeProvider>(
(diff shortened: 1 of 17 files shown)


donteatfriedrice:
* **#8110** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/8110?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

packages/affine/shared/src/services/doc-mode-service.ts
```diff
@@ -9,13 +9,15 @@ import { type Disposable, Slot } from '@blocksuite/global/utils';
 const DEFAULT_MODE = DocMode.Page;
 
 export interface DocModeProvider {
-  setMode: (mode: DocMode, docId?: string) => void;
-  getMode: (docId?: string) => DocMode;
-  toggleMode: (docId?: string) => DocMode;
-  onModeChange: (
+  setPrimaryMode: (mode: DocMode, docId?: string) => void;
```
Saul-Mirone, line 12: Let's also remove the `DocMode` enum and use pure string literal type (`page` | `edgeless`) instead.

packages/blocks/src/root-block/page/page-root-service.ts
```diff
@@ -8,6 +12,8 @@ import { RootService } from '../root-service.js';
 export class PageRootService extends RootService {
   static override readonly flavour = RootBlockSchema.model.flavour;
 
+  override readonly mode = DocMode.Page;
```
Saul-Mirone, line 15: `BlockService` will be deprecated soon, we shouldn't add new logic on it.

Saul-Mirone:
This API is a little confusing for block developers. For example, what should I do if I want to check current editor mode?

Saul-Mirone (changes requested):
Let's optimize the API

donteatfriedrice:
> Let's optimize the API

Originally, I thought of adding mode to the root service so that when developers want to get the mode, they can do so through rootService.mode. However, since the block service is about to be deprecated, this approach might not be suitable.

Do you have any suggestions regarding the following two points:

1. A suitable location to place the mode for easy user access?
2. API design for the doc mode service.

doodlewind (approved):

packages/affine/shared/src/services/doc-mode-service.ts
```diff
@@ -4,18 +4,20 @@ import type { Container } from '@blocksuite/global/di';
...
+import { type Disposable, noop, Slot } from '@blocksuite/global/utils';
 
 const DEFAULT_MODE: DocMode = 'page';
 
 export interface DocModeProvider {
-  setMode: (mode: DocMode, docId?: string) => void;
-  getMode: (docId?: string) => DocMode;
-  toggleMode: (docId?: string) => DocMode;
-  onModeChange: (
+  setPrimaryMode: (mode: DocMode, docId: string) => void;
```
Saul-Mirone, line 12: Please add some comments here to explain when to use primary mode and when to use editor mode.

Saul-Mirone (approved):
LGTM

PR #8138, opened 2024-09-03
refactor(database): move data-view to @blocksuite/data-view

zzj3720:


packages/affine/data-view/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../../../tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "./src/",
+    "outDir": "./dist/",
+    "noEmit": false
+  },
+  "include": ["./src"],
+  "references": []
```
Saul-Mirone, line 9: You need to set `reference` according to your dependencies like https://github.com/toeverything/blocksuite/blob/master/packages/affine/block-list/tsconfig.json#L9

In a new project, this will make sure the dependencies built before ts build this package.

Saul-Mirone (approved):
LGTM 💯

PR #8148, opened 2024-09-04
refactor(edgeless): split block-surface package

doodlewind:
Before:

```ts
import { SurfaceBlockModel } from '../../surface-block/index.js'
```

After:

```ts
import { SurfaceBlockModel } from '@blocksuite/affine-block-surface'
```

TODO: remove legacy `blocks/surface-block` after all of its references are migrated to `@blocksuite/affine-block-surface`.


packages/affine/block-surface/src/element-model/mindmap.ts
```diff
@@ -0,0 +1,1087 @@
...
+
+export type SerializedMindmapElement = SerializedElement & {
+  children: Record<string, NodeDetail>;
+};
+
+type MindmapElementProps = BaseElementProps & {
+  children: Y.Map<NodeDetail>;
+};
+
+export class MindmapElementModel extends GfxGroupLikeElementModel<MindmapElementProps> {
```
Saul-Mirone, line 77: I still believe this should be part of the `affine-model`

doodlewind: Will address this in further PR.

packages/affine/block-surface/src/surface-model.ts
```diff
@@ -0,0 +1,199 @@
...
+
+        if (children?.size === 0) {
+          childrenMap.delete(id);
+        }
+      }
+    }
+  },
+} satisfies Record<string, MigrationRunner<typeof SurfaceBlockSchema>>;
+
+export const SurfaceBlockSchema = defineBlockSchema({
```
Saul-Mirone, line 132: Same here.

doodlewind: Would also leave this in new PR.

Saul-Mirone (approved):

Issue #8150, opened 2024-09-04
unexpected newline appears while pasting multiply-paragraph-blocks to code-block


raintoway:
as follow
![6](https://github.com/user-attachments/assets/b52e0cf4-52c9-4656-88f3-b54decc67ab6)

fourdim:
Closed by #8575.

PR #8154, opened 2024-09-04
fix(blocks): unexpected newline appears while pasting multiply-paragr…

raintoway:
…aph-blocks to code-block

[issue](https://github.com/toeverything/blocksuite/issues/8150)

effect as follow


packages/blocks/src/_common/adapters/plain-text.ts
```diff
@@ -232,6 +232,7 @@ export class PlainTextAdapter extends BaseAdapter<PlainText> {
       return null;
     }
     payload.file = payload.file.replaceAll('\r', '');
+    payload.file = payload.file.replace(/\n$/, '');
```
L-Sun, line 235: I think the reason is in the exporter side, not here. @fourdim PTAL.

raintoway: yeah, i think so too ,  here is the export side in `/packages/blocks/src/_common/adapters/plain-text.ts`
```ts
  private async _traverseSnapshot(
    snapshot: BlockSnapshot
  ): Promise<{ mixtext: string }> {
    ...
    walker.setEnter(o => {
      const text = (o.node.props.text ?? { delta: [] }) as {
        delta: DeltaInsert[];
      };
      switch (o.node.flavour) {
        ...
        case 'affine:paragraph': {
          buffer += text.delta.map(delta => delta.insert).join('');
          buffer += '\n';
          break;
        }
 ...
```
but i think change here will affect a lot , so i only change the adapter

fourdim: No, changing the adapter will affect a lot.

> It's not about adding an extra newline at the end of a file, it's about not removing the newline that should be there.

> The C standard specifies that a C file should end with a newline (C11, 5.1.1.2, 2.) and that a last line without a newline yields undefined behavior (C11, J.2, 2nd item).

https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

fourdim (changes requested):

packages/blocks/src/_common/adapters/plain-text.ts
fourdim: No, changing this adapter will affect a lot.

> It's not about adding an extra newline at the end of a file, it's about not removing the newline that should be there.

> The C standard specifies that a C file should end with a newline (C11, 5.1.1.2, 2.) and that a last line without a newline yields undefined behavior (C11, J.2, 2nd item).

https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

fourdim (changes requested):

fourdim:
Closed by #8575.

PR #8177, opened 2024-09-05
feat(blocks): latex block in page mode

Flrande:
https://github.com/user-attachments/assets/53bc7068-9cb9-44b4-ad6a-977ddf76702e


packages/blocks/src/latex-block/latex-service.ts
```diff
@@ -0,0 +1,6 @@
+import { LatexBlockSchema } from '@blocksuite/affine-model';
+import { BlockService } from '@blocksuite/block-std';
+
+export class LatexBlockService extends BlockService {
```
Saul-Mirone, line 4: If this service is not used, you don't need to add it.

PR #8280, opened 2024-09-10
fix: latex editor shake

Flrande:
Close [BS-1364](https://linear.app/affine-design/issue/BS-1364/[可优化的]-input-popup-的-ui-抖动)


Saul-Mirone (approved):
LGTM

PR #8287, opened 2024-09-10
feat(database): easy access to property visibility

golok727:
- Added `hide in view` in the context menu

image


zanwei:
hey @golok727 how's going?

Thanks for your contribution! We would like to implement this feature; however, there are some adjustments that need to be made. Since we are planning to redesign the database UI, you only need to adjust the position of `Hide in view`
image

Also, please remove the hidden hint at the top. Users can check the properties through the "More" button, which should be sufficient
image

zanwei:
LGTM

zanwei (approved):

Saul-Mirone:
@zzj3720 PTAL

zzj3720 (approved):

PR #8302, opened 2024-09-11
feat(database): full width POC

zzj3720:
close: BS-1329


Saul-Mirone:
Should we merge this pr?

PR #8323, opened 2024-09-12
fix(blocks): not keep block selection when trigger latex editor

Flrande:
Close [BS-1408](https://linear.app/affine-design/issue/BS-1408/[行间-latex]-input-内的删除行为会删掉整个-block)

### DIFF
--- a/packages/blocks/src/latex-block/latex-block.ts
+++ b/packages/blocks/src/latex-block/latex-block.ts
@@ -114,11 +114,7 @@ export class LatexBlockComponent extends CaptionedBlockComponent<
     this._editorAbortController?.abort();
     this._editorAbortController = new AbortController();
 
-    this.selection.setGroup('note', [
-      this.selection.create('block', {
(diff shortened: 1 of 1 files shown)


doodlewind (approved):

packages/blocks/src/latex-block/latex-block.ts
```diff
@@ -114,11 +114,7 @@ export class LatexBlockComponent extends CaptionedBlockComponent<
     this._editorAbortController?.abort();
     this._editorAbortController = new AbortController();
 
-    this.selection.setGroup('note', [
-      this.selection.create('block', {
-        blockId: this.model.id,
-      }),
-    ]);
+    this.selection.setGroup('note', []);
```
Saul-Mirone, line 117: I think you should stop propagation of the latex editor popup or add another hotkey handler which returns `true` to avoid potential keymap conflict instead of do something like this.

PR #8333, opened 2024-09-12
feat(blocks): parse doc url with mode

EYHN:
I forgot to return the `mode` field when parse doc url.
In addition, I also modified the implementation of parsing url in clipboard, and now it completely use parsedocurl for parsing.


L-Sun (approved):
LGTM

PR #8368, opened 2024-09-15
refactor(inline): range sync

Flrande:
Close [BS-1332](https://linear.app/affine-design/issue/BS-1332/latex-editor-输入光标不正常)


doodlewind (commented):
Wondering are there some RFCs or descriptions about the refactoring?

Flrande:
> Wondering are there some RFCs or descriptions about the refactoring?

No, because this PR did not cause many API changes; it mainly focuses on optimizing the existing code.

Saul-Mirone (changes requested):
Waiting for resolving IME issues.

packages/framework/inline/src/services/render.ts
```diff
@@ -0,0 +1,179 @@
...
+  };
+
+  rerenderWholeEditor = () => {
+    const rootElement = this.editor.rootElement;
+
+    if (!rootElement.isConnected) return;
+
+    rootElement.replaceChildren();
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    delete (rootElement as any)['_$litPart$'];
```
Saul-Mirone, line 159: Need to add a comment to explain what is this.

packages/framework/block-std/src/range/range-binding.ts
```diff
@@ -130,23 +123,18 @@ export class RangeBinding {
...
-
-        // Restore the DOM structure damaged by the composition
-        parent.dirty = true;
-        await parent.updateComplete;
-        await parent.updateComplete;
-        parents.push(parent);
-      }
+      this.host.renderRoot.replaceChildren();
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      delete (this.host.renderRoot as any)['_$litPart$'];
```
Saul-Mirone, line 128: Same here

Saul-Mirone (approved):

PR #8422, opened 2024-09-20
Feat/arabic support

BaytaX:
add support for arabic letters so it goes from right to left


Saul-Mirone:
We don't plan to do this in such a pattern. Basically you need to modify every single block with text. Which is unacceptable. And it also doesn't consider features in whiteboard and database.

BaytaX:
okay got it but i need it for a project i work on so how's that possible to modify in the editor and import from the one who i change

Issue #8380, opened 2024-09-17
snapshotToDoc very slow with large snapshot


kerituni12:
The performance is slower due to the use of nextTick in the _snapshotToBlock method. Specifically, I see the following code:

```js 
await new Promise(resolve => nextTick(() => resolve(undefined)));
```

While nextTick is useful for handling large content during pasting, it can slow down operations when used in snapshotToDoc. It would be beneficial to add a parameter like usingNextTick to snapshotToDoc or explore other approaches to mitigate this issue.

R151831:
@doodlewind  

We are encountering the same issue. Can we proceed with removing the nextTick line? 
Please let us know if there are any potential consequences of doing so.

RajaTheKing826:
I'm also facing this issue from recent days. Don't we have any fix for this.

EddiePengg:
我也遇到了这个问题，希望在json的处理上能够更优秀，这样就有更通用的语言，可以被更多项目使用

PR #7796, opened 2024-08-01
fix: paste large content with operational ui

Saul-Mirone:
https://github.com/user-attachments/assets/ba1f769f-17bd-4ac0-8efe-1959ad9d8dd2

PR #8427, opened 2024-09-21
perf: faster tick for snapshot

doodlewind:
This tries to optimize #8380

### DIFF
--- a/packages/framework/global/src/env/index.ts
+++ b/packages/framework/global/src/env/index.ts
@@ -26,5 +26,4 @@ export const IS_IPAD =
 
 export const IS_WINDOWS = /Win/.test(platform);
 
-export const REQUEST_IDLE_CALLBACK_ENABLED =
-  'requestIdleCallback' in globalThis;
+export const RIC_ENABLED = 'requestIdleCallback' in globalThis;
--- a/packages/framework/global/src/utils/function.ts
+++ b/packages/framework/global/src/utils/function.ts
@@ -1,3 +1,5 @@
+import { IS_WEB, RIC_ENABLED } from '../env/index.js';
+
 export async function sleep(ms: number, signal?: AbortSignal): Promise<void> {
   return new Promise(resolve => {
     if (signal?.aborted) {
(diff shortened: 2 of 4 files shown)


fourdim:
I do think it's much better to add a parameter in this function.
RequestIdleCallback will schedule the task to a not near future compared to promise.resolve, if I remember correctly.

doodlewind:
@fourdim `queueMicroTask` could be much faster (but jams the UI). I'm just not sure why we make async loading by default.

```js
function testNextTick(useRequestIdleCallback, iterations = 1000) {
  const nextTick = useRequestIdleCallback
    ? (callback) => requestIdleCallback(() => callback())
    : (callback) => queueMicrotask(() => callback());

  return new Promise((resolve) => {
    const startTime = performance.now();
    let count = 0;

    function runIteration() {
      if (count < iterations) {
        nextTick(() => {
          count++;
          runIteration();
        });
      } else {
        const endTime = performance.now();
        resolve(endTime - startTime);
      }
    }

    runIteration();
  });
}

async function runTest() {
  console.log('Starting performance test...');

  const rICTime = await testNextTick(true);
  console.log(`requestIdleCallback: ${rICTime.toFixed(2)}ms`);

  const qMTTime = await testNextTick(false);
  console.log(`queueMicrotask: ${qMTTime.toFixed(2)}ms`);

  const difference = Math.abs(rICTime - qMTTime);
  const fasterMethod = rICTime < qMTTime ? 'requestIdleCallback' : 'queueMicrotask';

  console.log(`\nDifference: ${difference.toFixed(2)}ms`);
  console.log(`${fasterMethod} is faster in this test.`);
}

runTest();
```

```
requestIdleCallback: 63213.20ms
queueMicrotask: 4.10ms
```

fourdim:
We won't jam the UI becuase we have batch import function which reserves the time for UI to response.

doodlewind:
> We won't jam the UI becuase we have batch import function which reserves the time for UI to response.

Here is also the code path when pasting very long content, see #7796

PR #7790, opened 2024-08-01
perf(store): snapshot to block in batch

fourdim:

PR #8428, opened 2024-09-22
perf: optimize snapshot job handling

### LINKED ISSUE #8380: snapshotToDoc very slow with large snapshot

doodlewind:
Close #8380

## Problem

The original process of converting snapshot JSON to blocks was significantly slowed to ensure UI responsiveness. It involved two types of async computations:

* Asynchronously converting snapshots to draft models.
* Asynchronously inserting draft models into the block tree using `doc.addBlock`.

Original import sequence:

```
Convert snapshot node A
Insert block A
Wait for tick

Convert snapshot node B
Insert block B
Wait for tick

Convert snapshot node C
Insert block C
Wait for tick
...
```

Despite `doc.addBlock` being synchronous, a `requestAnimationFrame` interval was added after each block insertion to prevent UI blocking. This resulted in long processing times when pasting thousands of blocks:

https://github.com/user-attachments/assets/3a6379c9-621e-490e-a79f-4377494d5915

## New Design

This PR reimplements the process:

1. Convert all snapshot nodes first, either using `Promise.all` or sequential awaits: Firstly we flatten the block tree, asynchronously convert all nodes, then rebuild the tree.
2. Insert the converted draft block tree in batches, **waiting for a tick after each _batch_ instead of after each _block_**.

This optimizes both block tree conversion and insertion while maintaining UI responsiveness.

### Comparing serial vs parallel (`Promise.all`) snapshot tree conversion:

- 1000 blocks: `Promise.all` 4.8ms, series 2.8ms
- 60 images: `Promise.all` 2.8ms, series 2.2ms

Serial conversion proved faster and was implemented in b90d736f65f51302750ea44efc91fc286b074f84. The flatten-rebuild logic was retained for scalability.

### Pasting 1000 blocks performance:

Original: ~10s (see screen recording above)

New: ~1s with maintained UI responsiveness

https://github.com/user-attachments/assets/0202330a-66ce-47f3-b220-44a4e9c7c37d

### Switching to imported documents:

Loading the block tree into DOM (by switching to an imported document) causes similar delays in both implementations.

---

Legacy first take:

<details>

This PR significantly improves the performance of block import operations by addressing several inefficiencies in the current implementation:

- Removed the need to wait for `requestAnimationFrame` after each block insertion (see #7796), which was causing significant slowdowns.
- <del>Fixed the batch mechanism (see #7790) that wasn't properly enabled, replacing serial execution with true parallel processing.</del>
- Instead of executing all remaining tasks at once after 100 items (still causing UI freezes), we now process tasks in consistent batches of 100 per tick.
- Replaced `setTimeout` with `requestIdleCallback` for batch ticks, better utilizing idle browser time.

These optimizations result in much faster block imports while maintaining UI responsiveness. The new implementation processes 100 tasks per tick, striking a balance between speed and smooth user experience.

Tested using [clipboard-test.md](https://github.com/user-attachments/files/17088183/clipboard-test.md) with 1000 line and 600 blocks:

## Before (took ~10s to paste)

https://github.com/user-attachments/assets/3a6379c9-621e-490e-a79f-4377494d5915

## After (instant response without blocking UI)

https://github.com/user-attachments/assets/dea7c1f9-a8c1-4926-92b7-f5cb40957fad

</details>

### DIFF
--- a/packages/framework/store/src/transformer/job.ts
+++ b/packages/framework/store/src/transformer/job.ts
@@ -3,6 +3,7 @@ import { Slot } from '@blocksuite/global/utils';
 
 import type { BlockModel, BlockSchemaType } from '../schema/index.js';
 import type { Doc, DocCollection, DocMeta } from '../store/index.js';
+import type { DraftModel } from './draft.js';
 import type {
   BeforeExportPayload,
   BeforeImportPayload,
@@ -19,7 +20,6 @@ import type {
 
 import { AssetsManager } from './assets.js';
 import { BaseBlockTransformer } from './base.js';
-import { type DraftModel, toDraftModel } from './draft.js';
 import { Slice } from './slice.js';
 import {
   BlockSnapshotSchema,
@@ -33,16 +33,39 @@ export type JobConfig = {
   middlewares?: JobMiddleware[];
 };
 
+interface FlatSnapshot {
+  snapshot: BlockSnapshot;
+  parentId?: string;
+  index?: number;
+}
+
+interface DraftBlockTreeNode {
+  draft: DraftModel;
+  snapshot: BlockSnapshot;
+  children: Array<DraftBlockTreeNode>;
+}
+
+async function nextTick() {
+  // @ts-ignore
+  if ('scheduler' in window && 'yield' in window.scheduler) {
+    // @ts-ignore
+    return window.scheduler.yield();
+  } else if (typeof requestIdleCallback !== 'undefined') {
+    return new Promise(resolve => requestIdleCallback(resolve));
+  } else {
+    return new Promise(resolve => setTimeout(resolve, 0));
+  }
+}
+
+// The number of blocks to insert in one batch
+const BATCH_SIZE = 100;
+
 export class Job {
   private readonly _adapterConfigs = new Map<string, string>();
 
   private readonly _assetsManager: AssetsManager;
 
-  private _batchCounter = 0;
-
   private readonly _collection: DocCollection;
-
-  private readonly _pendingOperations: (() => void)[] = [];
 
   private readonly _slots: JobSlots = {
     beforeImport: new Slot<BeforeImportPayload>(),
@@ -50,8 +73,6 @@ export class Job {
     beforeExport: new Slot<BeforeExportPayload>(),
     afterExport: new Slot<FinalPayload>(),
   };
-
-  private _unblockTimer?: ReturnType<typeof setTimeout>;
 
   blockToSnapshot = async (
     model: DraftModel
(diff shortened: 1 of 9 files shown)


packages/framework/store/src/transformer/job.ts
```diff
@@ -259,12 +267,12 @@ export class Job {
...
-      for (const [i, block] of content.entries()) {
-        contentBlocks.push(
-          await this._batchSnapshotToBlock(block, doc, parent, (index ?? 0) + i)
-        );
-      }
+
+      const contentBlockPromises = content.map((block, i) =>
+        this._batchSnapshotToBlock(block, doc, parent, (index ?? 0) + i)
+      );
+      const contentBlocks = await Promise.all(contentBlockPromises);
```
fourdim, line 274: No Promise.all here.
It will cause race conditions where it is not executed in the order expected.

fourdim: As you can see you passed the initial test while the second test failed.

packages/framework/store/src/transformer/job.ts
```diff
@@ -486,21 +516,17 @@ export class Job {
...
       index
     );
 
-    for (const [index, child] of children.entries()) {
-      await this._batchSnapshotToBlock(child, doc, id, index);
-    }
+    const childPromises = children.map((child, childIndex) =>
+      this._batchSnapshotToBlock(child, doc, id, childIndex)
+    );
+    await Promise.all(childPromises);
```
fourdim, line 529: Here also.

packages/framework/store/src/transformer/job.ts
doodlewind: Do you know what's the root cause of the race condition? Seems that `snapshotToBlock` also fails some test cases when running in parallel. If we can't enable `Promise.all`, the batch mechanism is not actually utilized.

fourdim: The reason is that doc.addblock needs a parent. If the parent is not there, everything could be wrong.

doodlewind:
@Saul-Mirone @fourdim refactored under new RFC, PTAL.

fourdim (approved):

fourdim:
@Saul-Mirone What about your use case in https://github.com/toeverything/blocksuite/pull/7796?
Is this PR suitable?

Saul-Mirone (approved):
LGTM

Saul-Mirone:
> @Saul-Mirone What about your use case in #7796? Is this PR suitable?

I've did some tests and the implementation in this pr seems fast enough.

doodlewind:
### Merge activity

* **Oct 14, 4:04 AM EDT**: The merge label 'merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=toeverything&repo=blocksuite) once it meets the requirements.
* **Oct 14, 4:04 AM EDT**: A user added this pull request to the [Graphite merge queue](https://app.graphite.dev/merges?org=toeverything&repo=blocksuite).

PR #8438, opened 2024-09-24
refactor: edgeless tool manager

doouding:
The `EdgelessToolManager` has been renamed to `GfxToolController`. All the legacy tools will be rewrited using the new infra.

#### How to add a new tool
Let's take an example of pan tool.

```typescript
import { BaseTool } from '@blocksuite/block-std/gfx';

// extend the BaseTool
export class PanTool extends BaseTool {
  // every tool class should have static toolName
  static override toolName = 'pan';
    
  dragStart(e) {
    this._lastPoint = [e.x, e.y];
  }

  dragEnd(e) {
    this._lastPoint = null;
  }

  dragMove(e) {
    if (!this._lastPoint) return;

    const { viewport } = this._service;
    const { zoom } = viewport;

    const [lastX, lastY] = this._lastPoint;
    const deltaX = lastX - e.x;
    const deltaY = lastY - e.y;

    this._lastPoint = [e.x, e.y];

    this.gfx.viewport.applyDeltaCenter(deltaX / zoom, deltaY / zoom);
  }
}
```

Once you finish defining the tool, just append it to the spec array.

```typescript
// edgeless-spec.ts file
import { PanTool } from '../tools/pan-tool.ts';

export const EdgelessSpec = [
   // .... omit the other part of spec

   // append it to the array
   PanTool
]
```

Continue in https://github.com/toeverything/blocksuite/pull/8471

PR #8471, opened 2024-09-26
refactor: migrating legacy tools

doouding:
Continue https://github.com/toeverything/blocksuite/pull/8438

There're some depended modules that need to be refactor also during migration. I'll list them here.

### GfxExtension
Used to extend the `GfxController`. The `GfxExtension` can perform tasks during the life cycle of `GfxController`.

```typescript
import { GfxExtension } from '@blocksuite/std/gfx';

export class ExampleExtension extends GfxExtension {
  // the key is required
  static override key = 'example';

  // you can access this.gfx inside the class context

  mounted() {
    // called when gfx has mounted
  }

  unmounted() {
    // called when gfx has unmounted
  }
}

// edgeless-spec.ts
export const edgelessSpec = [
  // append it on the spec list
  ExampleExtension
]
``` 

### KeyboardController
`KeyboardController` provide the pressing status of common key such as shift, space. So that you don't have to maintain thoese status everywhere. Now you can access it through `gfx.keyboard`.

### EdgelessSelectionManager
The legacy `EdgelessSelectionManager` has move to `block-std/gfx` and become a first-party supported module. Now you can access it through `gfx.selection`.

### OverlayExtension
The legacy overlay has to be instantiated and added manually, which is a bit of a hassle. The newly refactored overlay utilizes the new Extension infra to automatically instantiate and add itself to the overlay renderer.

```typescript
import { Overlay } from './renderer/overlay.js';
 
export class ExampleOverlay extends Overlay {
  static override overlayName = 'example';

  // the methods remain the same
}

// edgeless-spec.ts
import { ExampleOverlay } from './example-overlay.js'

export const edgelessSpec = [
  // append it on the spec list
 ExampleOverlay
]
```

When you need to get the instance of that overlay, just use the `std.get(OverlayIdentifier(theOverlayName)`.

```typescript
import { OverlayIdentifier } from './renderer/overlay.js';

const exampleOverlayInst = std.get(OverlayIdentifier('example');
```

Note that you don’t have to use the OverlayExtension if you only need a local overlay. You can still create and add it to the renderer manually.

### SurfaceMiddlewareExtesion
The new surface middleware is aimed to extend the functionality of `addElement`, `deleteElement` and `updateElement`. It's a pure local infra which is helpful when you need to do something when the previously mentioned methods are called.

You can load the middleware manually using `surfaceModel.applyMiddlewares` method. But we provide a more common way to extend the middleware using the Extension infra.

```typescript
import { type SurfaceMiddlewareBuilder } from '@blocksuite/block-std/gfx';

export const ExampleMiddlewareBuilder: SurfaceMiddlewareBuilder = (std: BlockStdScope) => {
  /**
   * you can access std here
   * the builder must return a middleware function which will be apply to surface model
   **/
  return (ctx: MiddlewareCtx) => {
    // this middleware function will be executed every time when the surface methods get called.
  } 
}


// edgeless-spec.ts
import { SurfaceMiddlewareExtension } from '@blocksuite/block-std/gfx';
import { ExampleMiddlewareBuilder } from './middlewares.js'

export const edgelessSpec = [
  // ...
  // the surface builder should be wrapped in the `SurfaceMiddlewareExtension` function to get injected
  SurfaceMiddlewareExtension([ExampleMiddlewareBuilder])
]
```

### Review guide
This PR is a bit huge, but most of changes are just migration from legacy implementation. Here the list that worth attention.

- Every change under block-std package
- New extendable overlay under block-surface/src/renderer/overlay.ts


doouding:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/8471?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#8471** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/8471?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#8438** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/8438?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`

<h2></h2>

packages/framework/block-std/src/gfx/tool/tool-controller.ts
```diff
@@ -1,52 +1,139 @@
...
+};
+
+type BuiltInHooks = {
+  beforeToolUpdate: BuiltInHookEvent<{
+    toolName: keyof BlockSuite.GfxToolsMap;
+  }>;
+  toolUpdate: BuiltInHookEvent<{ toolName: keyof BlockSuite.GfxToolsMap }>;
+};
+
+class MiniEventEmitter<T> {
```
Saul-Mirone, line 45: This should be part of `@blocksuite/global`

packages/framework/block-std/src/gfx/controller.ts
```diff
@@ -197,16 +265,75 @@ export class GfxController extends LifeCycleWatcher {
...
 ) as ServiceIdentifier<GfxController>;
+
+export const GfxExtensionIdentifier =
+  createIdentifier<GfxExtension>('GfxExtension');
+
+const GfxClassExtenderIdentifier = createIdentifier<{
+  extendFn: (gfx: GfxController) => void;
+}>('GfxClassExtender');
+
+export abstract class GfxExtension extends Extension {
```
Saul-Mirone, line 301: Put this in another file.

packages/framework/block-std/src/gfx/selection.ts
```diff
@@ -0,0 +1,390 @@
...
+  get selectedIds() {
+    return [...this._selectedSet];
+  }
+
+  get selectedSet() {
+    return this._selectedSet;
+  }
+
+  get stdSelection() {
+    return this.std.selection;
```
Saul-Mirone, line 134: Personally I think `gfxSelection.stdSelection` makes no differences with `gfxSelectionManager.std.selection`

packages/framework/block-std/src/gfx/selection.ts
```diff
@@ -0,0 +1,390 @@
...
+
+  private _remoteSurfaceSelectionsMap = new Map<number, SurfaceSelection[]>();
+
+  private _selectedSet = new Set<string>();
+
+  private _surfaceSelections: SurfaceSelection[] = [];
+
+  disposable: DisposableGroup = new DisposableGroup();
+
+  readonly slots = {
```
Saul-Mirone, line 62: It makes more sense to use signals instead of slots here because you're watching reactive data instead of handling events.

packages/framework/block-std/src/gfx/tool/tool-controller.ts
```diff
@@ -155,9 +415,90 @@ export class ToolController {
...
+    };
+
+    this.std.provider.getAll(ToolIdentifier).forEach(tool => {
+      // @ts-ignore
+      tool['eventTarget'] = eventTarget;
+      this._register(tool);
+    });
+  }
+
+  // @ts-ignore
```
Saul-Mirone, line 438: Why do you need this?

packages/framework/block-std/src/gfx/tool/tool-controller.ts
```diff
@@ -155,9 +415,90 @@ export class ToolController {
...
+      toolName: toolNameStr,
+    });
+    this._eventEmitter.emit('beforeToolUpdate', beforeUpdateCtx.evt);
+
+    this.currentTool$.peek()?.deactivate();
+    this.currentToolName$.value = toolNameStr;
+
+    const currentTool = this.currentTool$.peek();
+    if (!currentTool) {
+      throw new Error(`Tool "${this.currentToolName$.value}" is not defined`);
```
Saul-Mirone, line 470: Use `BlockSuiteError`

packages/framework/block-std/src/gfx/selection.ts
doouding: I just copy from the old file, the refactor of part haven't start yet.

Saul-Mirone (approved):

doouding:
### Merge activity

* **Oct 21, 11:25 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/toeverything/blocksuite/8471) disabled "merge when ready" on this PR due to: a merge conflict with the target branch; resolve the conflict and try again..

PR #8485, opened 2024-10-02
fix(block-std): dispatcher pointerUp calls twice

majula2111:
When subscribing to pointerUp it is patched 2 times due to a duplicate in ClickController


L-Sun (approved):
Good catch!

PR #8498, opened 2024-10-08
fix: press backspace at the start of first line when edgeless text exist

Flrande:
Close BS-1528


Saul-Mirone (approved):

doodlewind:
GitHub warns about using CJK in branch name, which should be best to avoid.

image

PR #8543, opened 2024-10-16
perf(edgeless): use css var to place collaboration cursors-n-selections on board zoom change

oliseviche:
Instead of redrawing the whole _EdgelessRemoteSelectionWidget_ component during _updateTransform signal we could just place the selections and cursors by using css vars anc calculations right in css engine


Saul-Mirone:
@doouding PTAL

doouding (approved):

PR #8616, opened 2024-10-28
refactor(store): signal based canUndo and canRedo

L-Sun:


packages/framework/store/src/store/doc/block-collection.ts
```diff
@@ -139,14 +150,14 @@ export class BlockCollection {
     if (this.readonly) {
       return false;
     }
-    return this._history.canRedo();
+    return this._canRedo$.value;
```
Flrande, line 153: Should we use `peek` here? I concerned about the potential dependencies to signal.

Flrande (approved):

PR #8657, opened 2024-11-01
chore: make playground run in secure context

siyou:
when accessing the playground via IP on the LAN, the following error is thrown
image


doodlewind (changes requested):
Thanks for your work. It's great for anyone meeting clipboard issue, but it looks overkill since it fails E2E.

PR #8680, opened 2024-11-06
fix: edgeless crash on mobile safari

doodlewind:
This continues https://github.com/toeverything/blocksuite/compare/edgeless_perf_debug?expand=1


doodlewind (commented):
Most test cases are fixed, awaiting @Flrande for the edgeless-text fix.

PR #8748, opened 2024-11-14
refactor(database): avoid modifying all views when editing a single view

zzj3720:
close: BS-1808


packages/affine/shared/src/utils/insert.ts
```diff
@@ -1,3 +1,5 @@
+import { generateKeyBetween } from 'fractional-indexing';
```
EYHN, line 1: please copy this https://github.com/toeverything/AFFiNE/blob/59dfc80bc2019e87a13cdbdf8bd466db32c02e53/packages/common/infra/src/utils/fractional-indexing.ts#L11
 from affine

much safer then origin fractional-indexing

PR #8829, opened 2024-11-29
feat: support heading toggle

Flrande:


packages/affine/block-paragraph/src/commands/indent-paragraph.ts
```diff
@@ -80,9 +84,31 @@ export const indentParagraphCommand: Command<'indentContext'> = (ctx, next) => {
...
+
+  if (
+    matchFlavours(model, ['affine:paragraph']) &&
+    model.type.startsWith('h') &&
+    model.collapsed
+  ) {
+    const collapsedSiblings = calculateCollapsedSiblings(model);
+    doc.moveBlocks([model, ...collapsedSiblings], previousSibling);
+  } else {
+    doc.moveBlocks([model], previousSibling);
```
L-Sun, line 108: I think `getNearestHeadingBefore` and updating collapsed state should be after indenting paragraph. There is a case show in the following video

https://github.com/user-attachments/assets/4e0ef45a-d3e2-436f-b867-b14dfd7f3cef

Flrande: I didn't consider this case; it seems that the collapse state needs to be updated both before and after moving lol.

https://github.com/user-attachments/assets/25c61db4-26a7-44cc-8f5a-8cd073f60d2b

packages/affine/model/package.json
```diff
@@ -23,6 +23,7 @@
     "@blocksuite/global": "workspace:*",
     "@blocksuite/inline": "workspace:*",
     "@blocksuite/store": "workspace:*",
+    "@preact/signals-core": "^1.8.0",
```
Saul-Mirone, line 26: why adding this?

packages/affine/block-paragraph/src/paragraph-drag-extension.ts
```diff
@@ -0,0 +1,51 @@
...
+  ParagraphBlockSchema,
+} from '@blocksuite/affine-model';
+import { DragHandleConfigExtension } from '@blocksuite/affine-shared/services';
+import {
+  calculateCollapsedSiblings,
+  captureEventTarget,
+  matchFlavours,
+} from '@blocksuite/affine-shared/utils';
+
+export const ParagraphDragHandleOption = DragHandleConfigExtension({
```
Saul-Mirone, line 12: This `DragHandleConfigExtension` is not used in new drag and drop.

packages/affine/model/package.json
Flrande: It was originally used to listen for signals in the paragraph model, but it is no longer needed and will be deleted.

packages/affine/block-paragraph/src/paragraph-drag-extension.ts
Flrande: What should I use now?

L-Sun:
It seems that the range is not correct after collapsing:
1. set cursor on the begin of paragraph
2. collapse heading 1
3. press up arrow key
4. throw an error


1. set cursor on the begin of paragraph
2. collapse heading 1
3. type some characters
4. `Heading 2` -> `Hxxxxeading 2`
     expected: `xxxxHeading 2` ?

https://github.com/user-attachments/assets/952256ba-a959-41ca-9ccc-dda3e1121a5e

packages/affine/block-paragraph/src/paragraph-drag-extension.ts
Saul-Mirone: emmm, the new drag and drop havn't support extending now and you need to write the logic in [drag-event-watcher](https://github.com/toeverything/blocksuite/blob/master/packages/blocks/src/root-block/widgets/drag-handle/watchers/drag-event-watcher.ts)

L-Sun (approved):

Saul-Mirone (approved):

Flrande:
### Merge activity

* **Dec 17, 2:24 AM EST**: The merge label 'merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=toeverything&repo=blocksuite) once it meets the requirements.

PR #8836, opened 2024-12-02
feat(blocks): provide a mobile spec patches

L-Sun:
This PR provides a mobile patches for better user experience. Close [BS-1824](https://linear.app/affine-design/issue/BS-1824/移动端去除-slash的响应)

### What Changes
- hide inline reference popup
- use `keyboard-toolbar` instead of slash menu and block toolbar widgets
- a new @ menu for mobile
- update paragraph placeholder description

### How to use?
In the setup stage of editor, add this spec patches to editor
```ts
import {  MobileSpecsPatches } from '@blocksuite/blocks';
import { IS_MOBILE } from '@blocksuite/global/env';

if (IS_MOBILE) {
  editor.pageSpecs.push(MobileSpecsPatches);
  editor.edgelessSpecs.push(MobileSpecsPatches);
}
```

### DIFF
--- a/packages/blocks/src/_specs/preset/mobile-patch.ts
+++ b/packages/blocks/src/_specs/preset/mobile-patch.ts
@@ -1,0 +1,103 @@
+import type { Container } from '@blocksuite/global/di';
+
+import {
+  type ReferenceNodeConfig,
+  ReferenceNodeConfigIdentifier,
+} from '@blocksuite/affine-components/rich-text';
+import {
+  type BlockStdScope,
+  LifeCycleWatcher,
+  WidgetViewMapIdentifier,
+  type WidgetViewMapType,
+} from '@blocksuite/block-std';
+
+import { pageRootWidgetViewMap } from '../../root-block/page/page-root-spec.js';
+import { AFFINE_EMBED_CARD_TOOLBAR_WIDGET } from '../../root-block/widgets/embed-card-toolbar/embed-card-toolbar.js';
+import { AFFINE_FORMAT_BAR_WIDGET } from '../../root-block/widgets/format-bar/format-bar.js';
+import { AFFINE_SLASH_MENU_WIDGET } from '../../root-block/widgets/slash-menu/index.js';
+
+export class MobileSpecsPatches extends LifeCycleWatcher {
+  static override key = 'mobile-patches';
+
+  constructor(std: BlockStdScope) {
+    super(std);
+
+    std.doc.awarenessStore.setFlag('enable_mobile_keyboard_toolbar', true);
+    std.doc.awarenessStore.setFlag('enable_mobile_linked_doc_menu', true);
+  }
+
+  static override setup(di: Container) {
+    super.setup(di);
+
+    // Hide reference popup on mobile.
+    {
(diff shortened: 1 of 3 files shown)


L-Sun:
* **#8836** <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/8836?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/toeverything/blocksuite/8836?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`

<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

packages/blocks/src/_specs/preset/mobile-patch.ts
```diff
@@ -0,0 +1,103 @@
...
+import { AFFINE_FORMAT_BAR_WIDGET } from '../../root-block/widgets/format-bar/format-bar.js';
+import { AFFINE_SLASH_MENU_WIDGET } from '../../root-block/widgets/slash-menu/index.js';
+
+export class MobileSpecsPatches extends LifeCycleWatcher {
+  static override key = 'mobile-patches';
+
+  constructor(std: BlockStdScope) {
+    super(std);
+
+    std.doc.awarenessStore.setFlag('enable_mobile_keyboard_toolbar', true);
```
Flrande, line 25: Why we need to set flag here🤔

L-Sun: It may be better to add both widget to root widgets map.

L-Sun: After trying, it is difficult to control component rendering through extension patch. These two `setFlag` lines are a good example:

- `keyboard-toolbar` is an independent widget for the mobile version, and it is easy to control by overriding the `pageRootWidgetViewMap`.
- However, the latter flag is used to control whether to enable the mobile at-menu, and the mobile at-menu shares most of its logic with the desktop version. So, it is difficult to achieve this goal using the spec patch method only.

```ts
// affine-linked-doc-widget

  show = (mode: 'desktop' | 'mobile' = 'desktop') => {
    // ...

    const enableMobile = this.doc.awarenessStore.getFlag(
      'enable_mobile_linked_doc_menu'
    );

    this._show$.value = enableMobile ? mode : 'desktop';
  };

  override render() {
    if (this._show$.value === 'none') return nothing;

    return html`${this._renderInputMask()}
      <blocksuite-portal
        .shadowDom=${false}
        .template=${choose(
          this._show$.value,
          [
            ['desktop', this._renderLinkedDocPopover],
            ['mobile', this._renderLinkedDocMenu],
          ],
          () => html`${nothing}`
        )}
      ></blocksuite-portal>`;
  }
```

In conclusion, I think setting both flags in the constructor of the patch extension is acceptable. They can be removed in the next version release.

Flrande (approved):

PR #8863, opened 2024-12-05
fix(database): storage should be accessed via globalThis

zzj3720:


doodlewind (approved):

packages/blocks/src/database-block/utils/current-view.ts
```diff
@@ -8,9 +8,10 @@ const currentViewListSchema = z.array(
 );
 const maxLength = 20;
 const currentViewListKey = 'blocksuite:databaseBlock:view:currentViewList';
+const storage = globalThis.sessionStorage;
 const createCurrentViewStorage = () => {
   const getList = () => {
-    const string = localStorage.getItem(currentViewListKey);
+    const string = storage.getItem(currentViewListKey);
```
EYHN, line 14:
```suggestion
    const string = storage?.getItem(currentViewListKey);
```

PR #8870, opened 2024-12-05
fix(std): pan and pinch events were calculate relative to only one finger

majula2111:
Improves the calculation of multi-touch gestures and makes them smoother on Android and fixes them on iOS
**master android samsung galaxy s24 chrome:**

https://github.com/user-attachments/assets/5a41d922-bfdd-4743-b445-0cfa69175136

Visible jumps during pan gesture
**current branch with fix android samsung galaxy s24 chrome:**

https://github.com/user-attachments/assets/647bb8e9-0e58-4aaa-bf14-179be74f31a6

Smooth pan gesture with two fingers without jumping

### DIFF
--- a/packages/framework/global/src/utils/math.ts
+++ b/packages/framework/global/src/utils/math.ts
@@ -1,3 +1,4 @@
+import type { IPoint } from '../utils.js';
 import type { Bound, IBound } from './model/bound.js';
 
 import { PointLocation } from './model/point-location.js';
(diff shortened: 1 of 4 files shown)


packages/framework/global/src/utils/math.ts
```diff
@@ -536,3 +537,11 @@ export function getCenterAreaBounds(bounds: IBound, ratio: number) {
     rotate,
   };
 }
+
+export function dotProduct(vectorA: IPoint, vectorB: IPoint) {
```
doodlewind, line 541: Please use `Vec.dpr` and `Vec.len` instead.

majula2111: thx, replaced

doodlewind (approved):
Thanks for the fix!

PR #8893, opened 2024-12-07
fix(database): move cursor in kanban card title by arrow keys

### LINKED ISSUE #8643: ​Kanview view cannot use left and right key to move cusor when editing title of cards
### What happened?

image
​Kanview view cannot use left and right key to move cusor when editing title of cards

0.17.5

### Distribution version

Windows x64

_No response_

### Are you self-hosting?

- [ ] Yes

### Relevant log output

_No response_

### Anything else?

_No response_

OlegDev1:
Fixes #8643

- Fixes a bug where a user can not move the cursor with right and left arrow keys, when editing title of kanban cards
- Added an e2e test case

Ran these test cases, they all pass:
`tests/database/selection.spec.ts`
`tests/database/database.spec.ts`


doodlewind (approved):
Thanks for the fix!

PR #8940, opened 2024-12-11
perf(std): cache dom rect for pointer controller

doodlewind:
Before (query DOM rect on every mouse move):

https://github.com/user-attachments/assets/7edd5d29-cf43-4bdd-a055-c7194a80a121

After:

https://github.com/user-attachments/assets/71c8b721-3978-46ba-b580-5a213308b2d6

### DIFF
--- a/packages/framework/block-std/src/event/control/pointer.ts
+++ b/packages/framework/block-std/src/event/control/pointer.ts
@@ -29,11 +29,35 @@ function createContext(
 }
 
 abstract class PointerControllerBase {
+  private _cachedRect: DOMRect | null = null;
+
+  private _resizeObserver: ResizeObserver | null = null;
+
   protected get _rect() {
-    return this._dispatcher.host.getBoundingClientRect();
+    if (!this._cachedRect) {
+      this._updateRect();
+    }
+    return this._cachedRect as DOMRect;
   }
 
-  constructor(protected _dispatcher: UIEventDispatcher) {}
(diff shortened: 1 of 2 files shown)


doodlewind:
Thanks for the good catch! Wondering can we trigger `updateRect` via some API in there cases?

L-Sun:
BTW, after testing, using `MutationObserver` can cover the both cases mentioned above. I will create another PR to fix it

```ts
  constructor(protected _dispatcher: UIEventDispatcher) {
    this._resizeObserver = new ResizeObserver(() => this._updateRect());
    this._mutationObserver = new MutationObserver(() => this._updateRect());
  }

  protected _updateRect() {
    console.log('update rect cache');
    if (!this._dispatcher.host) return;

    if (this._resizeObserver) {
      this._resizeObserver.observe(this._dispatcher.host);
    }

    if (this._mutationObserver) {
      this._mutationObserver.observe(this._dispatcher.host, {
        attributes: true,
        childList: true,
        subtree: true,
      });
    }

    this._cachedRect = this._dispatcher.host.getBoundingClientRect();
  }

  dispose() {
    this._resizeObserver?.disconnect();
    this._mutationObserver?.disconnect();
    this._resizeObserver = null;
  }
```

doodlewind:
Let's put a `nextTick` after the polling, at least we are using idle callback in this case 😈

PR #9040, opened 2024-12-20
fix: can not select all block when some block collapsed

Flrande:


Saul-Mirone:
Please move the pr to affine

PR #9044, opened 2024-12-23
mobile testing infrastructure

alsuren:
My eventual aim is to make kanban drag and drop work on mobile. I've still not managed that, but I thought I could contribute my testing infrastructure already.

To avoid burning github actions minutes and have a billion failing tests to sort out, I made it only run interesting tests on mobile. The rest of the tests are run on desktop.

* .spec.ts files are run on desktop only
* .mobile.spec.ts files are run on mobile only
* .shared.spec.ts files are run on both desktop and mobile

If you remove the test.skip(), my mobile test does correctly fail. I may need to rewrite it to send touch events once I've implemented the feature, but I will think about that later.

### DIFF
--- a/tests/database/kanban.shared.spec.ts
+++ b/tests/database/kanban.shared.spec.ts
@@ -1,0 +1,31 @@
+
+import {
+  enterPlaygroundRoom,
+  initKanbanViewState,
+} from '../utils/actions/index.js';
+import { test } from '../utils/playwright.js';
+import {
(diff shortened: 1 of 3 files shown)


alsuren:
closed due to https://github.com/toeverything/blocksuite/issues/9045 (https://github.com/toeverything/blocksuite/pull/9048#issuecomment-2564176930)

Issue #9045, opened 2024-12-25
Announcement: The Evolution of BlockSuite


Saul-Mirone:
Dear Community,

We are excited to share some important updates about the future of BlockSuite and AFFiNE.

BlockSuite has been the core editor powering AFFiNE, designed initially as a universal editor framework. Over time, as both BlockSuite and AFFiNE evolved, BlockSuite gradually incorporated more and more of AFFiNE’s specific business logic. While this close integration brought significant enhancements to AFFiNE, it also led BlockSuite away from its original vision as a standalone, universal framework.

To realign with our goals and streamline development, we have made the decision to merge BlockSuite's codebase into AFFiNE. This change allows us to focus on refining the editor specifically for AFFiNE while maintaining its excellent performance and features for our users.

At the same time, we remain committed to the idea of a universal editor framework. Therefore, we will extract and rebuild the core elements of BlockSuite into a new, independent project. This will allow us to return to the original mission of BlockSuite—creating a flexible, universal editor framework that can be used beyond AFFiNE.

## What This Means for You
For AFFiNE users: The transition will improve the editor experience within AFFiNE, as we can now tailor BlockSuite's capabilities > to meet the specific needs of AFFiNE users more effectively.

For developers and contributors: The new universal editor project derived from BlockSuite's core will provide an exciting opportunity to collaborate on an open-source editor framework designed for broader use cases.

## What’s Next?
We are currently working on integrating BlockSuite into AFFiNE while simultaneously outlining the roadmap for the new universal editor framework. We will keep you updated on both efforts and invite you to contribute your feedback and ideas.

Thank you for being a part of this journey with us. We’re excited about the possibilities this change brings for AFFiNE, BlockSuite, and the community at large.

Stay tuned, and let’s build the future of editing together!

Best regards,
The AFFiNE Team

medzhidov:
Will you continue support blocksuite as separate project?

Saul-Mirone:
> Will you continue support blocksuite as separate project?

Yes, we plan to move back the infra part once it's stable.

Saul-Mirone:
@alterhuman no, it's a long term plan and will take several months.

goldsam:
[...]

PR #9058, opened 2025-01-22
fix(editor): brush xywh do not behave predictably when changing lineWidth of the rotated brush element

EGRoBBeRTiT:
This pull request fixes the bug with incorrect calculation xywh of the brush element when changing lineWidth when the element is rotated (see the attached video)

https://github.com/user-attachments/assets/b28a606b-ba0a-40cc-9ec8-13fee1dd9c9f

### DIFF
--- a/packages/affine/model/src/elements/brush/brush.ts
+++ b/packages/affine/model/src/elements/brush/brush.ts
@@ -138,19 +138,21 @@ export class BrushElementModel extends GfxPrimitiveElementModel<BrushProps> {
     instance['_local'].delete('commands');
   })
   @derive((lineWidth: number, instance: Instance) => {
+    const oldBound = Bound.fromXYWH(instance.deserializedXYWH);
+
     if (
       lineWidth === instance.lineWidth ||
-      instance.w === 0 ||
-      instance.h === 0
+      oldBound.w === 0 ||
+      oldBound.h === 0
     )
       return {};
 
     const points = instance.points;
     const transformed = transformPointsToNewBound(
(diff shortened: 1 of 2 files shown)


packages/affine/model/src/elements/brush/brush.ts
```diff
@@ -138,21 +138,19 @@ export class BrushElementModel extends GfxPrimitiveElementModel<BrushProps> {
     instance['_local'].delete('commands');
   })
   @derive((lineWidth: number, instance: Instance) => {
-    const oldBound = instance.elementBound;
```
L-Sun, line 141: Good catch! This change effectively addresses the issue mentioned, but it could be even better by fixing it with just one line of code.

```ts
    const oldBound = Bound.fromXYWH(instance.deserializedXYWH);
```

EGRoBBeRTiT: Done!

L-Sun:
During this PR process, the blocksuite code has already been migrated to the AFFiNE side (more information in https://github.com/toeverything/blocksuite/issues/9045). It would be better if you could resubmit the changes from this PR on the [AFFiNE](https://github.com/toeverything/AFFiNE) side.

The new file path in AFFiNE is `blocksuite/affine/model/src/elements/brush/brush.ts`

L-Sun (approved):

L-Sun:
This PR has been transfered to https://github.com/toeverything/AFFiNE/pull/11524. Thank you for your contribution

